PR #36, opened 2015-12-29
Modifications to GuildBanAdd, closes #29 #33.  Idiomatic improvements to ChannelMessages func.

### LINKED ISSUE #29: Add support for delete-message-days=x on bans endpoint

### LINKED ISSUE #33: GuildBanAdd should be renamed
In order to match the wording used for other functions, GuildBanAdd should be renamed to GuildBanCreate.

iopred:
Fixes #29
Fixes #33

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -275,10 +275,10 @@ func (s *Session) GuildBans(guildID string) (st []User, err error) {
 	return
 }
 
-// GuildBanAdd bans the given user from the given guild.
+// GuildBanCreate bans the given user from the given guild.
 // guildID   : The ID of a Guild.
 // userID    : The ID of a User
-func (s *Session) GuildBanAdd(guildID, userID string) (err error) {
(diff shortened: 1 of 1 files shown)


bwmarrin:
Hey, thanks for all the help!  I really do appreciate it.  With the Delete Comments addition I'd actually like to add that to the existing GuildBanCreate function instead of making a new function for it.  I'd like to have this match how the ChannelMessages func works.  Just add a check if days is over 0 then include it similar to that func.  If you want to make those changes I'll import them in for sure.  Otherwise I can just import this and modify it slightly :)  Again, though.  I really appreciate the help!

iopred:
I have made the changes you suggested, I also cleaned up ChannelMessage a little because there was a more idiomatic way to achieve it.

bwmarrin:
Oh, great :) That does look much better! Thank you very much for the help.

PR #37, opened 2015-12-29
Make update status a little more idiomatic.

iopred:
This also reduces the line count a smidge.

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -60,12 +60,12 @@ func (s *Session) Handshake() (err error) {
 }
 
 type updateStatusGame struct {
-	Name *string `json:"name"`
+	Name string `json:"name"`
 }
 
 type updateStatusData struct {
-	IdleSince *int             `json:"idle_since"`
-	Game      updateStatusGame `json:"game"`
+	IdleSince *int              `json:"idle_since"`
+	Game      *updateStatusGame `json:"game"`
 }
 
 type updateStatusOp struct {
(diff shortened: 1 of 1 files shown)


bwmarrin:
Oh that's fantastic. I knew there had to be a better way to accomplish what I needed.    One question though, at line 68 should the type updateStatusGame also be a pointer?  If no game is set the Game value in the updateStatusData has to be set to "null" in the json data sent.  I had used interface before because that allowed me to set it to nil (which turns into null when marshaled) but I didn't know I could use pointers to accomplish that too!


https://github.com/iopred/discordgo/commit/974d5862a448eee2c96a5aa012e8eb1557bf55a3#diff-9c7b1115ae6d16ebcb782b325ee2b8b3R68

iopred:
I was going off these docs:
http://discordapi-unoffical.readthedocs.org/en/latest/reference/websockets/general.html#update-status

Which says game.name should be null.

bwmarrin:
Oh, great.  When they first announced that change they  (Discord devs) said that the "game" field should be null but if you can just set game.name to null and that works.  Then whichever is fine.  I suppose it saves a couple bytes to just set game to null though but that's pretty trivial difference :)  

I could run a quick test and find out which methods actually work correctly.

bwmarrin:
Ah, got clarification from Discord :)

The entire game object should be set to null.  The method of setting game.name to null should not be used and will likely be rejected in the future.  The docs are wrong.

Would you like the adjust for the change? Then I'll pull in straight in, otherwise I can make the changes afterwords.

iopred:
I have updated the change, thanks for the clarification from the devs (it makes more sense this way)!

bwmarrin:
Again, thank you a lot for helping.  Especially on this type of stuff as I just started learning Go with this project and there's a lot of little idiomatic type things I am still trying to figure out :)

PR #42, opened 2016-01-02
Implement state tracking.

iopred:
Currently maintains a list of Guilds (Members/Channels) and PrivateChannels.

This allows a lot of nice things like a really easy 'is a message private' check:
```
func (s *Session) IsMessagePrivate(m *Message) bool {
	_, err := s.State.GetPrivateChannelByID(m.ID)
	return err == nil
}
```

Or a nice check where we grab an invite, and only accept it if we need to.
```
int := "discord.gg/blah"
i, _ := d.Session.Invite(inv)
if _, err := d.Session.State.GetGuildByID(i.Guild.ID); err == nil {
    // There was no error, we are already in the guild.
    fmt.Println("I have already joined that server.")
    return
}
d.Session.InviteAccept(inv)
fmt.Println("I have joined the server.")
```

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -62,7 +62,8 @@ func New(args ...interface{}) (s *Session, err error) {
 
 	// Create an empty Session interface.
 	s = &Session{
-		State: NewState(),
+		State:        NewState(),
+		StateEnabled: true,
 	}
 
 	// If no arguments are passed return the empty Session interface.
--- a/discord.go
+++ b/discord.go
@@ -61,7 +61,9 @@ type voiceUDP struct {
 func New(args ...interface{}) (s *Session, err error) {
 
 	// Create an empty Session interface.
-	s = &Session{}
+	s = &Session{
+		State: NewState(),
+	}
 
 	// If no arguments are passed return the empty Session interface.
 	// Later I will add default values, if appropriate.
--- a/state.go
+++ b/state.go
@@ -1,0 +1,195 @@
+package discordgo
+
+import "errors"
+
+// NewState creates an empty state.
+func NewState() *State {
+	return &State{
+		Ready: Ready{
+			PrivateChannels: []Channel{},
+			Guilds:          []Guild{},
+		},
+	}
+}
+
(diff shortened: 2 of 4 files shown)


iopred:
I can see maybe you would want to make this a toggle, let me know, it would just be adding a bunch more if's in wsapi :ok_hand:

bwmarrin:
Oh, nice! Great addictions.  I wanted to at some point start adding some of these sort of helper functions into the library :)

I do think anything that caches data should be optional though.  Since in some cases it would be unused by someone using the library.  So we don't want to be wasting memory storing data that will never be used.  However it's absolutely handy for a lot of people as well :)

Lets make it optional and if a client user calls the discord.go New() func that sort of just sets everything up for them, default to having this these types of features enabled by default.  Of course they can still toggle it off.

If someone sets up their connection using the old manual methods it would default to having this type of features disabled.  Of course they can still toggle it on.

Also, some funcs have names like GetFuncName but throughout the rest of the library I'm not using names like that.  I think we should be consistent in naming across the board.   So something like GetGuildChannelByID should match the style in restapi GuildChannels  so maybe just State.GuildChannel(string)

Then, my final idea :) A lot of these functions would be handy regardless of caching and I've in the past liked the idea of adding helper functions to each of the main structs like Message, Guild, Channel, etc.

Would it be better to have a Guild.Channel(string) function that would pull from the cache if existed/enabled, otherwise do the leg work to get the information from Discord?  Similar functions for many of the other ones you've added?  I'm asking your thoughts on this one because I'm not sure what makes the best sense.  But I think these would be handy and creating two sets of similar functions maybe wouldn't be best?

Sorry, I know that's a lot of comments.

iopred:
I have changed the method names and made state tracking optional, additionally I have made it null safe just in-case someone does something silly like:
```
s := &discordgo.Session{}
s.StateEnabled = true
```

I think it would be super cool to make the Guild/Channel/User methods go to cache first, but I think that should be implemented as a layer on top of State, instead of intermingling those concerns.

PR #44, opened 2016-01-03
Add and Parse Emojis struct.

iopred:


wsapi.go
```diff
@@ -394,8 +405,7 @@ func (s *Session) event(messageType int, message []byte) (err error) {
 		}
 	default:
 		fmt.Println("UNKNOWN EVENT: ", e.Type)
-		// TODO learn the log package
-		// log.print type and JSON data
+		printJSON(message)
```
iopred: This is really handy, looks like it solves the TODO.

bwmarrin:
Well, I still want to setup logging to a file or stderr or something. I'm sure that's the right way to go about doing it :)  But, I do like that printJSON function!  Makes it really easy to read.

PR #48, opened 2016-01-06
Add emoji endpoint.

iopred:


bwmarrin:
So,to match current naming.  Could we call this just EMOJI  instead?

:)

Issue #49, opened 2016-01-06
Add support for registering a new account.


bwmarrin:
The Register endpoint is already defined but need to create a function to interact with it.

Rivalo:
The Register endpoint only registers an unverified account. It has no way of using the login() function, since there is no information to do an auth request.
For verifying the account (I haven't tested this, this is according to meewO on the discord server) the client sends a /user/@me PATCH request with the email and password in its payload.

bwmarrin:
Ah, thanks @Rivalo that is very helpful info!

Rivalo:
Yesterday for a test bot I already wrote a Register function in the style of the login function. 
`` func Register(u string, dg discordgo.Session) (token string, err error) {```

I could simply rewrite it for the Session interface and do a pull request, and maybe give the Verification function a try. Is development/forking done in the develop branch?

bwmarrin:
Yup, please do all PR's against the develop branch.  Thanks for any help! I really do appreciate it.

PR #55, opened 2016-01-07
Register & Verify

Rivalo:
This would fix: https://github.com/bwmarrin/discordgo/issues/49

A first handshake with the Discord API, without having any authentication details would be:
```go
	// Register to the Discord server and store the authentication token
	dg.Token, err = dg.Register("Username")
	if err != nil {
		fmt.Println(err)
		return
	}

	// Add authentication information to User
	_, err = dg.UserVerify("Username", "myemail@address.com", "Password")
	if err != nil {
		fmt.Println(err)
		return
	}

	// Login to the Discord Server and store the new authentication token
	dg.Token, err = dg.Login("myemail@address.com", "Password")
	if err != nil {
		fmt.Println(err)
		return
	}
```
Afterwards it is just like in the api_basic example:

```go
	// Login to the Discord Server and store the new authentication token
	dg.Token, err = dg.Login("myemail@address.com", "Password")
	if err != nil {
		fmt.Println(err)
		return
	}
```
Note that actually verifying the account by using the link in the email is necessary, but I could still login using the details without having to actually verify it.

This is all using the api_basic example, it doesn't integrate at the moment with the New() function. I'll have to look into that since I'm still using the old way in my bot.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -242,21 +242,6 @@ func (s *Session) UserGuilds(userID string) (st []Guild, err error) {
 	return
 }
 
-// UserVerify updates account information to make vertification possible.
-// Account still needs to be activated using the send email
-func (s *Session) UserVerify(username, email, password string) (st User, err error) {
-
-	data := struct {
-		Username string `json:"username"`
-		Email    string `json:"email"`
-		Password string `json:"password"`
-	}{username, email, password}
-
-	body, err := s.Request("PATCH", USER("@me"), data)
-	err = json.Unmarshal(body, &st)
-	return
-}
-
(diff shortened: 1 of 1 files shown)


bwmarrin:
I think the existing func UserUpdate also does the job if the new func UserVerify.  Lets just stick with the existing function for that task.  If you can update your PR with UserVerify removed I'll merge it.  The Register func looks great.  Thanks for the help!

Rivalo:
Yeah, it's basically the same function, but it might be necessary to address this in the library reference or in that New() function.

bwmarrin:
Thanks a bunch @Rivalo :)

PR #91, opened 2016-01-17
Add a new optional token to New, which is a cheaper way to login. Closes #89.

### LINKED ISSUE #89: Add support to verify a token
Per Jake, you will soon be able to verify a token is correct in a way that is much less expensive (cpu/mem) to Discord's servers.  This is done by sending username, password, AND the cached or provided token to the Discord Login endpoint.  Doing this (per Jake) will not count against your rate limit either.  The Username and Password are sent the same as they are already being sent, also the Token is also still sent via Headers in this case.  If the Token is invalid you will be issued a new Token, otherwise you will be sent back the same Token if it is valid.  If all info, token and user/pass are invalid then it should return a HTTP 400.

iopred:
Eventually we should consider allowing Login/LoginWithToken to mutate
s.Token, it would probably simplify the API a bit.

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -20,9 +20,15 @@ const VERSION = "0.10.0-alpha"
 
 // New creates a new Discord session and will automate some startup
 // tasks if given enough information to do so.  Currently you can pass zero
-// arguments and it will return an empty Discord session. If you pass a token
-// or username and password (in that order), then it will attempt to login to
-// Discord and open a websocket connection.
+// arguments and it will return an empty Discord session.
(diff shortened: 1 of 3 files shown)


bwmarrin:
Yeah that's not a bad idea really.  I still want to give the lib user access to the token so they can store it but they can just pull it from Session.Token :)  Do we need two Login funcs it would it be easier just to have one that will handle all situations?

iopred:
Ok, I did the proposed behavior and updated New to use an optional Token.

PR #93, opened 2016-01-21
Refactor open/close logic. Support OnConnect and OnDisconnect. Closes #5.

### LINKED ISSUE #5: Manage goroutines better.
I fire off a few goroutines but all of these need to have code added to monitor and control them.  Need to track what goroutines are running and ability to tell them to shut down.  Heartbeats and websocket listen goroutines for example.

iopred:
This has a backwards incompatible change to New.

The reason is that with an OnConnect event, there is no time to hook it up, eg:
```
a := discordgo.New()
a.OnConnect = func(s *discordgo.Session){
// Not called.
}
```

Instead, there is now an explicit (and simple) call to start listening, the above example is now:
```
a := discordgo.New()
a.OnConnect = func(s *discordgo.Session){
// Called.
}
a.OpenAndListen()
```

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -73,7 +73,7 @@ func (s *Session) Open() (err error) {
 	// Create listening outside of listen, as it needs to happen inside the mutex
 	// lock.
 	s.listening = make(chan interface{})
-	go s.listen(s.listening)
+	go s.listen(s.wsConn, s.listening)
 
 	s.Unlock()
 
@@ -112,9 +112,9 @@ func (s *Session) Close() (err error) {
 
 // listen polls the websocket connection for events, it will stop when
 // the listening channel is closed, or an error occurs.
-func (s *Session) listen(listening <-chan interface{}) {
+func (s *Session) listen(wsConn *websocket.Conn, listening <-chan interface{}) {
 	for {
-		messageType, message, err := s.wsConn.ReadMessage()
+		messageType, message, err := wsConn.ReadMessage()
 		if err != nil {
 			// There has been an error reading, Close() the websocket so that
 			// OnDisconnect is fired.
--- a/wsapi.go
+++ b/wsapi.go
@@ -70,6 +70,11 @@ func (s *Session) Open() (err error) {
 		return
 	}
 
+	// Create listening outside of listen, as it needs to happen inside the mutex
+	// lock.
+	s.listening = make(chan interface{})
+	go s.listen(s.listening)
+
 	s.Unlock()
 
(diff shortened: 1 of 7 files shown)


wsapi.go
```diff
@@ -202,8 +222,8 @@ func (s *Session) event(messageType int, message []byte) (err error) {
 			if s.OnReady != nil {
 				s.OnReady(s, st)
 			}
-			go s.Heartbeat(st.HeartbeatInterval)
 		}
+		go s.heartbeat(st.HeartbeatInterval)
```
iopred: I can't see any reason to keep Heartbeat as a public method, we removed the need to call it manually previously, and it is cleaned up with the Listen() loop on Close().

iopred:
Please pull and give it a run. I've been running Septapus on this branch all night and things look really great.

bwmarrin:
Reading though this now.  I think this touches on some things I'd been thinking about changing, maybe now is a good time.  

#1 - Why would anybody every called Open without wanting to call Listen/Heartbeat?  We could probably just combine those.

#2 - No idea how to do this, but I really liked how simple New() was.  Not sure if there's a magic trick we can use here to keep it as simple as it was before.

bwmarrin:
https://github.com/bwmarrin/discordgo/pull/93/files#diff-9c7b1115ae6d16ebcb782b325ee2b8b3R567

sendHeartbead should probably use an anonymous struct instead of a map to keep it consistent with the rest of the library.

iopred:
Cool, I agree, I didn't want to go too wild at first :D

iopred:
Ok, so here are my thoughts:
New() shouldn't need to listen. It's perfectly reasonable to want to use the rest api without using wsapi. I could see people wanting just to send a message in a cron job or something.

Open/Listen, I like the separation because Open() is blocking, but Listen isn't.
I want to consume the errors from Open, but I can't if it also does listening, as you can't get errors from goroutines.

I could see the last line in Open being go Listen(), but it seems more obvious to a developer imo.

Finally I updated the anonymous struct.

bwmarrin:
Stop making sense, you.

bwmarrin:
I do actually sort of like the idea of having go Listen as the last section of Open() we could even later add a little code so Open() can check via a chan/bool that the Listen didn't hit some error immediately.

I guess the fine line of exposing enough to make it a useful library but keeping it as simple as possible so it's easy for people to "just use".

iopred:
Actually I have been convinced about adding listen to the end of Open, making changes now.

PR #98, opened 2016-01-27
Support GuildMemberEdit api.

iopred:


bwmarrin:
Looking at this, wondering about when using this func would it be easier to pass a []string of the roleIDs?  Not sure which way would make the easiest for end users to use it.  Thoughts?

bwmarrin:
Here's an example of it being used, sort of :)
https://github.com/blackhole12/sweetiebot/blob/beb7d53f40602fac0e62b77b290c2ee79bff8f75/sweetiebot/spam_module.go#L28

iopred:
Done.

PR #102, opened 2016-02-08
VoiceServerUpdate and VoiceStateUpdate. Closes #101.

### LINKED ISSUE #101: VOICE_STATE_UPDATE does not fire user callback function
Events switch block need updated to fire the user definable callback function.

https://github.com/bwmarrin/discordgo/blob/develop/wsapi.go#L325

iopred:


bwmarrin:
Sorry, I think we both were working on this.  I did half of it already and created a new issue for the other half.  You're welcome to do the VoiceServerUpdate part matching how I did VoiceStateUpdate if you want :)

bwmarrin:
I'm probably missing something else, but why have this block here

https://github.com/bwmarrin/discordgo/pull/102/files#diff-9c7b1115ae6d16ebcb782b325ee2b8b3R347

Instead of the return here
https://github.com/bwmarrin/discordgo/pull/102/files#diff-9c7b1115ae6d16ebcb782b325ee2b8b3R348

With the return in it's own separate block the general OnEvent won't fire but since the specific OnEventName.. func didn't fire due to the unmarshal error, wouldn't it be good to let it flow through to the OnEvent func?  

Not sure if one way is better than the other for any specific reason, though letting it fall though would make the code a bit simpler.

iopred:
It's consistent with the rest of the handlers.

This can be cleaned up however, previously this method returned errors, now it does not.

PR #109, opened 2016-02-15
Mapped Event Handlers

iopred:
This change breaks the current API but makes it much more powerful.

Currently only one handler can be added for each event, with this change, any number of handlers can be added for the same event. Additionally it supports adding handlers for interface{} which will be triggered on every event.

The existing code:
```
dg.OnMessageCreate = func(s *discordgo.Session, m *discordgo.Message){
  fmt.Println(m.Content)
}
```
is now:
```
dg.AddHandler(func(s *discordgo.Session, m *discordgo.MessageCreate){
  fmt.Println(m.Content)
})
```

Not bad for a net delta of -:100: lines.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -29,45 +29,7 @@ type Session struct {
 	Token string // Authentication token for this session
 	Debug bool   // Debug for printing JSON request/responses
 
-	// Settable Callback functions for Internal Events
-	// OnConnect is called when the websocket connection opens.
-	OnConnect func(*Session)
-	// OnDisconnect is called when the websocket connection closes.
-	// This is a good handler to add reconnection logic to.
-	OnDisconnect func(*Session)
-
 	Handlers map[interface{}][]interface{}
-
-	// Settable Callback functions for Websocket Events
-	OnEvent                   func(*Session, *Event)
-	OnReady                   func(*Session, *Ready)
-	OnTypingStart             func(*Session, *TypingStart)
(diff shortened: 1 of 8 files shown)


wsapi.go
```diff
@@ -296,406 +322,23 @@ func (s *Session) event(messageType int, message []byte) {
...
-				if err != nil {
-					fmt.Println("error :", err)
-				}
-			}
-			if s.OnChannelDelete != nil {
-				s.OnChannelDelete(s, st)
-			}
+		// Attempt to unmarshal our event.
+		// If there is an error we should handle the event itself.
+		if err = unmarshal(e.RawData, i); err != nil {
```
iopred: We unmarshal every event now.

It's a small change, and doesn't really change the runtime performance that I can see, but it is required to support the interface{} event type, which is really handy :ok_hand:

structs.go
```diff
@@ -13,84 +13,66 @@ package discordgo
...
+	// Should the session request compressed websocket data.
+	Compress bool
+
+	// Should state tracking be enabled.
+	// State tracking is the best way for getting the the users
+	// active guilds and the members of the guilds.
+	StateEnabled bool
 
 	// Exposed but should not be modified by User.
-	SessionID  string // from websocket READY packet
```
iopred: I have removed SessionID as it is accessible from State.SessionID.

PR #132, opened 2016-02-24
Add support for setting guild region and verification level

nstafie:


structs.go
```diff
@@ -148,24 +148,32 @@ type Emoji struct {
...
+	Presences         []*Presence   `json:"presences"`
+	Channels          []*Channel    `json:"channels"`
+	VoiceStates       []*VoiceState `json:"voice_states"`
+}
+
+// A GuildParams stores all the data needed to update discord guild settings
+type GuildParams struct {
+	Name              string `json:"name"`
+	Region            string `json:"region"`
+	VerificationLevel *int   `json:"verification_level"`
```
iopred: Consider using a VerificationLevel type then you can do:

type VerificationLevel uint
const (
VerificationLevelNone VerificationLevel = iota
VerificationLevelWhatever
VerificationLevelBlah
)

restapi.go
```diff
@@ -371,14 +371,44 @@ func (s *Session) GuildCreate(name string) (st *Guild, err error) {
...
+
+	// Bounds checking for VerificationLevel, interval: [0, 3]
+	if g.VerificationLevel != nil {
+		val := *g.VerificationLevel
+		if val < 0 || val > 3 {
+			err = errors.New("VerificationLevel out of bounds, should be between 0 and 3")
+			return
+		}
+	}
+
```
iopred: Is name optional too?

nstafie: Yes it is, any of those those 3 values can be used together or individually.

structs.go
nstafie: Will do.

bwmarrin:
Looking at this, I'd almost like the struct to be named `GuildEditParams` just to make it more clear but maybe that's just too long.  thoughts?

nstafie:
Maybe we could even get rid of GuildParams and just use Guild and only set the relevant attributes, much like ChannelInviteCreate does with Invite?

bwmarrin:
I like that :) Add if they add more things we can change we don't need to change much.  Also keeps the syntax shorter.

nstafie:
Looking back on it, doing that would result in having to make Guild.VerificationLevel a `*int` which is then ugly to access and just makes it more confusing all around.

bwmarrin:
By the way, thanks a lot @nstafie  I really do appreciate you helping out the project!

nstafie:
:) I got some time to add consts for the values, I'll open a new PR!

PR #135, opened 2016-02-26
Add VerificationLevel value consts and a typedef for int

nstafie:
```go
level := discord.VerificationLevelNone
updateStruct := discordgo.Guild{VerificationLevel: &level}
guild, err := s.GuildEdit(GuildID, updateStruct)
```

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -145,35 +145,46 @@ type Emoji struct {
 	RequireColons bool     `json:"require_colons"`
 }
 
+// Custom VerificationLevel typedef for int
+type VerificationLevel int
+
+// Constants for VerificationLevel levels from 0 to 3 inclusive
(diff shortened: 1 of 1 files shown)


structs.go
```diff
@@ -145,35 +145,46 @@ type Emoji struct {
 	RequireColons bool     `json:"require_colons"`
 }
 
+// Custom VerificationLevel typedef for int
+type VerificationLevel int
+
+// Constants for VerificationLevel levels from 0 to 3 inclusive
+const (
+	VerificationLevelNone   VerificationLevel = iota
+	VerificationLevelLow    VerificationLevel = iota
```
iopred: You can remove the = iota on all but the first line.

PR #136, opened 2016-02-26
Add VerificationLevel value consts and a typedef for int

nstafie:
This time it'll work! :)

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -145,35 +145,46 @@ type Emoji struct {
 	RequireColons bool     `json:"require_colons"`
 }
 
+// Custom VerificationLevel typedef for int
+type VerificationLevel int
+
+// Constants for VerificationLevel levels from 0 to 3 inclusive
+const (
+	VerificationLevelNone   VerificationLevel = iota
+	VerificationLevelLow    VerificationLevel = iota
+	VerificationLevelMedium VerificationLevel = iota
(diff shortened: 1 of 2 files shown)


structs.go
```diff
@@ -145,35 +145,46 @@ type Emoji struct {
 	RequireColons bool     `json:"require_colons"`
 }
 
+// Custom VerificationLevel typedef for int
+type VerificationLevel int
+
+// Constants for VerificationLevel levels from 0 to 3 inclusive
+const (
+	VerificationLevelNone   VerificationLevel = iota
```
iopred: You can remove the = iota on every line after the first

nstafie:
```go
const (
	VerificationLevelNone   VerificationLevel = iota
	VerificationLevelLow    VerificationLevel
	VerificationLevelMedium VerificationLevel
	VerificationLevelHigh   VerificationLevel
)
```

Like this?

iopred:
Even better:

const (
    VerificationLevelNone   VerificationLevel = iota
    VerificationLevelLow
    VerificationLevelMedium
    VerificationLevelHigh
)

nstafie:
Yeah that's what I ended up doing, the way I asked about won't even compile

bwmarrin:
Yay :)

PR #139, opened 2016-03-03
Replace HTTP status codes for net/http constants.

noisypixy:


iopred:
Hmm, looks like it failed to build.

bwmarrin:
Thanks a lot for helping!

PR #141, opened 2016-03-05
Add the ability to have multiple cross-guild voice connections open.

b1naryth1ef:
This PR adds support for multiple cross-guild voice connections (one per guild), while also refactoring and cleaning the voice API. This is still a little messy, and could probably do with some love/cleanup before merg'n.

**NB**

1. Previously the library assumed we would get a `VoiceStateUpdate` _before_ a `VoiceServerUpdate`. Because we can't guarantee packets are processed in that order, we now explicitly wait for the VoiceStateUpdate data to be passed into the VoiceServerUpdate. This does create a possible deadlock when initially connecting to voice servers, so it may be advisable to add a timeout condition.
2. Previously the library did _not_ handle VoiceStateUpdate's for other users, causing our internal `State` to get desynced. We now track them, albeit somewhat messily.
3. Currently we only allow for one voice connection per guild (which will be valid going forward), however we don't really prevent users from smashing the internal session tracking of voice connections by creating new connections. We may want to either return the current voice session, or return an error when a user tries to open another voice connection to the same guild.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -54,8 +54,8 @@ type Session struct {
 	// Whether the UDP Connection is ready
 	UDPReady bool
 
-	// Stores all details related to voice connections
-	Voice *Voice
+	// Stores a mapping of channel id's to VoiceConnections
+	VoiceConnections map[string]*VoiceConnection
 
 	// Managed state object, updated internally with events when
 	// StateEnabled is true.
--- a/structs.go
+++ b/structs.go
@@ -54,7 +54,7 @@ type Session struct {
 	// Whether the UDP Connection is ready
 	UDPReady bool
 
-	// Stores a mapping of channel id's to VoiceConnections
+	// Stores a mapping of guild id's to VoiceConnections
 	VoiceConnections map[string]*VoiceConnection
 
 	// Managed state object, updated internally with events when
--- a/structs.go
+++ b/structs.go
@@ -203,6 +203,7 @@ type VoiceState struct {
 	UserID    string `json:"user_id"`
 	SessionID string `json:"session_id"`
 	ChannelID string `json:"channel_id"`
+	GuildID   string `json:"guild_id"`
 	Suppress  bool   `json:"suppress"`
 	SelfMute  bool   `json:"self_mute"`
 	SelfDeaf  bool   `json:"self_deaf"`
--- a/wsapi.go
+++ b/wsapi.go
@@ -55,6 +55,8 @@ func (s *Session) Open() (err error) {
 		}
 	}()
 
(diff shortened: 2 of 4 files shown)


wsapi.go
```diff
@@ -333,58 +336,49 @@ type voiceChannelJoinOp struct {
...
-	return
+	return voice, err
 }
 
 // onVoiceStateUpdate handles Voice State Update events on the data
 // websocket.  This comes immediately after the call to VoiceChannelJoin
 // for the session user.
 func (s *Session) onVoiceStateUpdate(se *Session, st *VoiceStateUpdate) {
+	// If we don't have a connection for the channel, don't bother
+	if st.ChannelID == "" {
```
b1naryth1ef: Need to fix this blocking the update case below.

structs.go
```diff
@@ -54,8 +54,8 @@ type Session struct {
 	// Whether the UDP Connection is ready
 	UDPReady bool
 
-	// Stores all details related to voice connections
-	Voice *Voice
+	// Stores a mapping of channel id's to VoiceConnections
```
b1naryth1ef: Comment should read `Stores a mapping of guild id's to VoiceConnections`

wsapi.go
```diff
@@ -333,58 +336,49 @@ type voiceChannelJoinOp struct {
...
+		return
+	}
 
-	// Ignore if Voice is nil
-	if s.Voice == nil {
+	channel, err := s.Channel(st.ChannelID)
```
b1naryth1ef: Since we only need the guild id, some kind of caching here would be dope.

wsapi.go
```diff
@@ -401,12 +395,49 @@ func (s *Session) onVoiceStateUpdate(se *Session, st *VoiceStateUpdate) {
...
+				break
+			}
+		}
+
+		if !exists {
+			return
+		}
+
+		// Handle Leaving Channel
+		if st.ChannelID == "" {
```
iopred: Can you move this state tracking code into state.go?

Also you can do:

```
	for i, state := range guild.VoiceStates {
		if state.UserID == st.UserID {
			guild.VoiceStates = append(guild.VoiceStates[:i], guild.VoiceStates[i+1:]...)
                         break
		}
	}
```

Instead of constructing a new array.

b1naryth1ef:
Things left:

- [ ] Some kind of timeout for `WaitUntilConnected`
- [ ] Return the existing voice connection from `ChannelVoiceJoin`
- [ ] Cleanup the comments and code a bit more

bwmarrin:
Decided to pull this in and we can still make updates to improve / cleanup as needed :)

PR #154, opened 2016-03-20
Fix channels from a GUILD_CREATE packet having null GuildID's

b1naryth1ef:
For any channels that are part of the Guild object given within a GUILD_CREATE packet, discord doesn't send 'guild_id' keys to save space within the WS frame. This means when handling GUILD_CREATE packets for brand new (e.g. untracked) we need to manually set the `guild_id` for every channel.

### DIFF
--- a/state.go
+++ b/state.go
@@ -73,6 +73,11 @@ func (s *State) GuildAdd(guild *Guild) error {
 		}
 	}
 
+	// Otherwise, update the channels to point to the right guild
+	for _, c := range guild.Channels {
+		c.GuildID = guild.ID
+	}
+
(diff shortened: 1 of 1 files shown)


state.go
```diff
@@ -73,6 +73,11 @@ func (s *State) GuildAdd(guild *Guild) error {
 		}
 	}
 
+	// Otherwise, update the channels to point to the right guild
+	for _, c := range guild.Channels {
+		c.GuildID = guild.ID
+	}
+
```
iopred: This happens on GUILD_CREATE too? Welp.

bwmarrin:
Yay :) Thanks @b1naryth1ef

PR #161, opened 2016-04-04
Fix lazy guild loading

b1naryth1ef:
We're (Discord) moving to a method of lazy-loading guilds to help with extremely large `READY` packets. With this new method, clients will now receive only the 'id' and 'unavailable' fields for every guild in the `READY` packet, and will lazily receive `GUILD_CREATE` events for those guilds as they are connected by our backend.

Previously discordgo used the Guild objects originally passed in the `READY` payload as the base object for all future events, however with this new method we now case this to when the original guild is not unavailable.

Something this commit does not add, but may be helpful is a way to detect the difference between `GUILD_CREATE`'s used for the handshake process, vs `GUILD_CREATE`'s that actually represent a guild being created/added. Due to the way the state works right now, the handler is always called before any of the users and therefore you cannot compare the state before/after the GUILD_CREATE to understand whether it is a new or lazily loaded guild.

:eyes: @iopred @bwmarrin

### DIFF
--- a/state.go
+++ b/state.go
@@ -57,22 +57,24 @@ func (s *State) GuildAdd(guild *Guild) error {
 	s.Lock()
 	defer s.Unlock()
 
-	// If the guild exists, replace it.
-	for i, g := range s.Guilds {
-		if g.ID == guild.ID {
-			// Don't stomp on properties that don't come in updates.
-			guild.Members = g.Members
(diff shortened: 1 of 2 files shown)


iopred:
Looks good to me.

bwmarrin:
Thanks!  LGTM too...

PR #162, opened 2016-04-09
Add Troubleshooting section to README.md

aviolentgrace:
This adds helpful troubleshooting information for people who are new to go and/or have brand new go environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,6 @@ cd $GOPATH/src/github.com/bwmarrin/discordgo
 git checkout develop
 ```
 
-
-
 ### Usage
 
 Import the package into your project.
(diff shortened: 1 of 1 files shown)


bwmarrin:
I've been thinking about this PR for a couple days.  I think what I would rather do is modify the readme to link to a wiki page that includes trouble shooting tips.  That way the readme doesn't get excessively long and it also makes updating the tips easier.  

I've created the wiki page already and if you would like, you can contribute this same content of the PR into the wiki page itself.  

https://github.com/bwmarrin/discordgo/wiki/Troubleshooting

aviolentgrace:
Thanks, b. Sounds like the right solution. =^)

PR #165, opened 2016-04-13
Fix channelMap not being filled for bot users/>100 guilds

b1naryth1ef:
Previously we only filled the channelMap inside of READY. Because of lazy guild loading, this was not valid for the majority of users (e.g. bot users and users with more than 100 guilds). We now properly handle both cases, by filling in the channelMap on GUILD_CREATE.

This commit also refactors the way we do locking inside of GUILD_CREATE and CHANNEL_CREATE to be a bit cleaner.

@iopred :eyes: plz

### DIFF
--- a/state.go
+++ b/state.go
@@ -45,7 +45,6 @@ func (s *State) OnReady(r *Ready) error {
 
 		for _, c := range g.Channels {
 			c.GuildID = g.ID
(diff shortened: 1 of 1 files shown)


iopred:
LGTM ship it.

PR #167, opened 2016-04-13
Support Gateway v4

b1naryth1ef:
This adds support for gateway v4 (which we're technically already running; whoops). Really it just moves the version to the URL, and explicitly uses JSON as the encoding. Everything should be supported from previous commits.

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -15,6 +15,7 @@ import (
 	"compress/zlib"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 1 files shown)


bwmarrin:
@b1naryth1ef care to update this for me?  It conflicted with your sharding PR that I just merged.  I could probably figure out the right fix but figured you know it better :)

bwmarrin:
Having reverted the other commit :) This one can now merge cleanly. Thanks @b1naryth1ef for the PR.

PR #174, opened 2016-04-24
Added some fields to ready event

jonas747:
As the title says


bwmarrin:
Thanks a lot @jonas747 and @iopred  too!  :heart:

PR #176, opened 2016-04-25
Updated settings struct

jonas747:
Removed muted_channels 
And added:
restricted_guilds (members from these guilds can't direct message you)
allow_email_friend_request
convert_emoticons
friend_source_flags
inline_attachment_media


bwmarrin:
:heart: Thanks.

PR #181, opened 2016-04-25
Add a new example that acts like airhorn.solutions

iopred:


bwmarrin:
Wow, this is cooler than the real airhorn!

b1naryth1ef:
Lame.

PR #186, opened 2016-04-28
Fixed ChannelMessageAck

jonas747:
ChannelMessageAck does not like `application/json` as the content type


bwmarrin:
Oh thanks @jonas747  :)

PR #189, opened 2016-04-30
Nickname support

jonas747:
Added the GuildMemberNickname function and the nick field to the guild member struct

One little thing, when changing your own nickname via @me, you have to add "/nick" to the url


bwmarrin:
Thanks @jonas747  :)  That's weird about the @me/nick

PR #196, opened 2016-05-08
Add Role tracking. Support disabling individual features in state. Closes #195.

### LINKED ISSUE #195: Newly created roles not being detected
when people add new roles the bot can not detect them.
i have to restart completely and than everything works fine

but if you're asking to pinpoint it's anything role related. like changing a role. detecting etc. if they Invite Echo and than make a bunch of roles. none will be detected
hard restart the bot and all roles are detected

**Updated:**
Tried going direct with `s.GuildRoles` instead of using State (for testing purposes)
and It still won't find the new role (even with AI's method using `GuildRoleCreate` and
`GuildRoleUpdate` events to update Guild in the State.

iopred:


bwmarrin:
What do you think about using a bitmask int to enable/disable options vs bools?

iopred:
Generally I think that matters when you need to transmit the data, in this
case it'll add complexity for no gain. (Like less bandwidth etc.)

bwmarrin:
Fair enough. :)

PR #199, opened 2016-05-14
Bulk delete

VagantemNumen:
Added bulk delete endpoint and bulk delete support.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1111,7 +1111,7 @@ func (s *Session) ChannelMessageDelete(channelID, messageID string) (err error) 
 // If only one messageID is in the slice call channelMessageDelete funciton.
 // If the slice is empty do nothing.
 // channelID : The ID of the channel for the messages to delete.
-// messages  : The IDs of the messages to be deleted. A slice of string IDs.
+// messages  : The IDs of the messages to be deleted. A slice of string IDs. A maximum of 100 messages.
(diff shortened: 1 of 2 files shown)


restapi.go
```diff
@@ -1107,6 +1107,34 @@ func (s *Session) ChannelMessageDelete(channelID, messageID string) (err error)
...
+		return
+	}
+
+	if len(messages) == 1 {
+		err = s.ChannelMessageDelete(channelID, messages[0])
+		return
+	}
+
+	if len(messages) > 100 {
+		messages = messages[:100]
```
iopred: Perhaps you could call bulk delete multiple times for each group of 100 messages?

VagantemNumen:
This is why I did not do it like that,
```
warn This endpoint has a rate limit of 1 request per second per guild, and is limited to 
100 messages per request. 
The request will fail if less than two snowflakes are provided. 
Only bot accounts can use this endpoint.
```

restapi.go
VagantemNumen: This is why I did not do it like that,
```
warn This endpoint has a rate limit of 1 request per second per guild, and is limited to 
100 messages per request. 
The request will fail if less than two snowflakes are provided. 
Only bot accounts can use this endpoint.
```

iopred: Nice, Looks good to me!

bwmarrin:
Great job @VagantemNumen and thanks a lot for the help! I really appreciate it!

PR #200, opened 2016-05-15
Rename GoDiscord to Discordgo in LICENSE.

IMcPwn:


IMcPwn:
The build failed because it was tested against Go 1.3 which websocket does not support. 

The change is valid but I'm not sure why this failed and not master.

bwmarrin:
That is odd, but could you commit this to the develop branch instead, pretty please :)

Also, use "DiscordGo" or "discordgo" those are the two accepted spelling options at the moment.  You're pick, whatever you like best :)

PR #201, opened 2016-05-16
Avatar Examples

VagantemNumen:
Examples demonstrating changing of avatar with local files and remote urls.


bwmarrin:
I think for the purpose of these examples we should remove the Message Create call back entirely so that the examples just focus on the require calls to change the avatar.  So also won't even need to open a websocket either.  Keep them very short and to the point so it's easier to understand for new users what's actually required.

Also, thanks a lot for the PR :) I really do appreciate the help.  If you don't mind could you try updating them to remove the websocket/handler stuff?

VagantemNumen:
Sure.

VagantemNumen:
Done.

bwmarrin:
Great.  One more thing, I was thinking it might be best to not include any images.  Keep the repo lighter but avoid any potential complains for using these trademarked images.  I think the example still shows the steps clear enough that users can provide their own image.

VagantemNumen:
Done.

bwmarrin:
Awesome :) Thanks a lot, it's a great addition.

VagantemNumen:
Glad to be of help. Well if you think of any other example that maybe useful I am more than happy to code them.

PR #203, opened 2016-05-18
Add guild_positions field to settings

jonas747:
probably gonna be the only one who uses this heh


bwmarrin:
Awesome, thanks @jonas747  :)

PR #211, opened 2016-05-30
Added GuildChannelsUpdate

jonas747:
It only updates the positions so should maybe be GuildChannelsReorder?

Also apparently does not return any data, even though the docs says so


bwmarrin:
Yeah lets go with GuildChannelsReorder if that's the only thing it does.  Might want to use GuildChannelsUpdate later for something covering a bigger scope.

Issue #219, opened 2016-06-24
Add support for pinned messages


bwmarrin:
https://github.com/hammerandchisel/discord-api-docs/issues/80
https://github.com/hammerandchisel/discord-api-docs/pull/81

PR #223, opened 2016-06-28
add support for pinning

RealCodingTeam:
#219


bwmarrin:
Hey, thanks a lot of the PR :) Looks great. 

One question, when you call these endpoints does Discord return anything?  Often it will return the  object that was modified.  If that's the case here I'd like to return that as well from these functions back to the caller.

RealCodingTeam:
Calling them returns an empty 204 response and judging by how similar returns were handled, I handled them the same.

bwmarrin:
k, just checking :)   Looks great - again thanks a lot, I really appreciate the help. :+1:

PR #226, opened 2016-07-04
Added ChannelEditTopic

albru123:
So it's simply a copied ChannelEdit function with different field (topic) being set. I tested it and it's apparently working. Change the method name by your taste.


LEGOlord208:
Why is this closed?

PR #227, opened 2016-07-11
Retry on 502's

jonas747:
Will retry up to Session.MaxRestRetries times

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -45,11 +45,13 @@ func (s *Session) Request(method, urlStr string, data interface{}) (response []b
 		}
 	}
 
-	return s.request(method, urlStr, "application/json", body)
+	return s.request(method, urlStr, "application/json", body, 0)
 }
 
 // request makes a (GET/POST/...) Requests to Discord REST API.
-func (s *Session) request(method, urlStr, contentType string, b []byte) (response []byte, err error) {
+// Sequence is the sequence number, if it fails with a 502 it will
+// retry with sequence+1 until it either succeeds or sequence >= session.MaxRestRetries
+func (s *Session) request(method, urlStr, contentType string, b []byte, sequence int) (response []byte, err error) {
 
 	// rate limit mutex for this url
 	// TODO: review for performance improvements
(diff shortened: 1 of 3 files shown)


structs.go
```diff
@@ -53,6 +53,9 @@ type Session struct {
 	// Whether the Data Websocket is ready
 	DataReady bool // NOTE: Maye be deprecated soon
 
+	// Max number of REST API retries
+	MaxRestRetries int
```
iopred: Set a default value in New, 3 seems reasonable.

jonas747: Done

restapi.go
```diff
@@ -708,7 +720,7 @@ func (s *Session) GuildMemberNickname(guildID, userID, nickname string) (err err
 // guildID   : The ID of a Guild.
 func (s *Session) GuildChannels(guildID string) (st []*Channel, err error) {
 
-	body, err := s.request("GET", EndpointGuildChannels(guildID), "", nil)
+	body, err := s.request("GET", EndpointGuildChannels(guildID), "", nil, 0)
```
iopred: This can be changed to s.Request("GET", EndpointGuildChannels(guildID), nil)

restapi.go
```diff
@@ -1084,7 +1096,7 @@ func (s *Session) ChannelMessage(channelID, messageID string) (st *Message, err
 // messageID : the ID of a Message
 func (s *Session) ChannelMessageAck(channelID, messageID string) (err error) {
 
-	_, err = s.request("POST", EndpointChannelMessageAck(channelID, messageID), "", nil)
+	_, err = s.request("POST", EndpointChannelMessageAck(channelID, messageID), "", nil, 0)
```
iopred: This can be changed to s.Request("GET", EndpointChannelMessageAck(channelID, messageID), nil)

jonas747: If my memory serves right reason it was not using Request was because that function adds the application/json content type, which the ack endpoint did not like, although atm the ack in discordgo is completely broken i think since it does stuff with tokens (haven't looked into it too much)

iopred: Hmm, that makes sense, we should probably remove the json header if we pass nil into Request, for now this seems good.

PR #235, opened 2016-07-28
Timestamp type; a string with a Parse() function

### LINKED ISSUE #204: Parse timestamps
Right now discordgo does not parse timestamps, and IMO it really should, my proposal is a timestamp struct: 

```go
type Timestamp struct {
	Raw    string
	Parsed time.Time
}

func (t *Timestamp) UnmarshalJSON(data []byte) (err error) {
	if len(data) < 1 { // Not always a timestamp returned
		return
	}
	err = json.Unmarshal(data, &t.Raw)
	if err != nil {
		return
	}

	t.Parsed, err = time.Parse("2006-01-02T15:04:05-07:00", t.Raw)
	return
}
```

Not sure if the raw, unparsed string is needed to be available but yeah
This would break everything using timestamps currently though, but fixes are simple

uppfinnarn:
Closes #204

### DIFF
--- a/message.go
+++ b/message.go
@@ -19,8 +19,8 @@ type Message struct {
 	ID              string               `json:"id"`
 	ChannelID       string               `json:"channel_id"`
 	Content         string               `json:"content"`
-	Timestamp       string               `json:"timestamp"`
-	EditedTimestamp string               `json:"edited_timestamp"`
+	Timestamp       Timestamp            `json:"timestamp"`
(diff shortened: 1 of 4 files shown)


jonas747:
This layout fails to parse `2016-08-15T00:53:06+00:00` which i got from a message
I use `2006-01-02T15:04:05-07:00` as the layout for discord stuff

robbix1206:
I done a other branch based on uppfinnarn with the correct pattern see Pull Request #245

PR #236, opened 2016-07-29
Added UserGuild struct, UserGuilds now returns a UserGuild slice

jonas747:


iopred:
Discord API says that a Guild is returned from that endpoint, why is this needed?

jonas747:
Guess it's not really needed, just for clarity, could just add the missing fields into the full guild struct.

also the User Guilds endpoint returns User Guild objects (https://discordapp.com/developers/docs/resources/user#user-guild-object)

restapi.go
```diff
@@ -351,7 +351,7 @@ func (s *Session) UserChannelCreate(recipientID string) (st *Channel, err error)
 }
 
 // UserGuilds returns an array of Guild structures for all guilds.
```
iopred: Update this comment.

PR #245, opened 2016-09-03
Timestamps

### LINKED ISSUE #204: Parse timestamps

robbix1206:
Closes #204
Based on uppfinnarn work
Just replace the Parser pattern by the time pattern of RFC3339

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -19,7 +19,7 @@ import (
 )
 
 // VERSION of Discordgo, follows Symantic Versioning. (http://semver.org/)
-const VERSION = "0.13.0"
+const VERSION = "0.14.0-dev"
 
 // New creates a new Discord session and will automate some startup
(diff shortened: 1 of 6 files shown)


robbix1206:
The merge destination branch was not the good one.

PR #248, opened 2016-09-04
Add the Bot prefix to authorization.

### LINKED ISSUE #247: "Bot" Token Type is Required
See https://github.com/hammerandchisel/discord-api-docs/issues/119 for more informations
This change will happen tomorrow.

robbix1206:
Closes #247
See https://github.com/hammerandchisel/discord-api-docs/issues/119

Current problem. This PR don't Authorize discordgo to use normal users however discordgo don't know if it connect has a Bot or as a normal user.

Do we keep allowed normal user accounts on Discordgo ?

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -27,6 +27,7 @@ const VERSION = "0.14.0-dev"
 // There are 3 ways to call New:
 //     With a single auth token - All requests will use the token blindly,
 //         no verification of the token will be done and requests may fail.
+//				 Please use this way only for bot users.
 //     With an email and password - Discord will sign in with the provided
 //         credentials.
 //     With an email, password and auth token - Discord will verify the auth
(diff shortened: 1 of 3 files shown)


nstafie:
> Do we keep allowed normal user accounts on Discordgo ?

Yeah, that's one of main goals of this library, to support both clients and bots. A better solution is needed...

robbix1206:
I think I've got a better one ^^
Can we consider the user is a bot if only token is provided ??

nstafie:
Well, that breaks a use case. User accounts can connect via tokens too, and we're been told in the past to actually cache the tokens because the login endpoints are expensive for discord, so someone developing a client might want to do that.

jonas747:
It also breaks the oauth use case since that uses the Bearer type

You can just manually specify the token type yourself atm (in the examples you would do `-t "Bot {token}"` for example)

robbix1206:
@nstafie Ok for this point I agree with you....
So maybe add something like dg.Bot() or somthing like that ??

@jonas747 . If you use '-t "Bot {token}"' you will break oauth too (example : "Authorization: Bearer Bot token" will not be accepted).
Discordgo don't use oauth2 no ?
Oauth2 endpoints ("/oauth2/applications/@me") is the only one that bot can access via discord go for bot accounts no ? (This endpoint accept Bot token type yes Bot is a Token type!)

robbix1206:
Hmmm.... well... I think jonas747 has right.
So i think this pull request is useless ^^
We just need to Precise that the token type `Bot` need to be precised with the token

PR #254, opened 2016-09-25
Fix message timestamp being empty when update is handled before create

jonas747:


jonas747:
Actually there is a lot more missing field in that case

PR #257, opened 2016-09-26
Fix for 401 Auth Error

NotBela:
It seems like the API requires a "Bot " + Token in the HTTP Auth header from now on.

"Authorization: TOKEN_TYPE TOKEN"

- From https://discordapp.com/developers/docs/reference


NotBela:
My bad, havent read the previous PR.

bwmarrin:
:heart:

PR #258, opened 2016-09-27
Add Support of changing user status

robbix1206:
Recently Discord add the support to change the User status beetween Online, Idle, Do not Distrub and Invisible.
This PR is to implement that new functionality.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -325,6 +325,7 @@ func (s *Session) UserSettings() (st *Settings, err error) {
 func (s *Session) UserUpdateStatus(status Status) (st *Settings, err error) {
 	if status == StatusOffline {
 		err = errors.New("You can't set your Status to offline")
+		return
 	}
 
 	data := struct {
--- a/restapi.go
+++ b/restapi.go
@@ -328,7 +328,7 @@ func (s *Session) UserUpdateStatus(status Status) (st *Settings, err error) {
 	}
 
 	data := struct {
-		Status string `json:"status"`
+		Status Status `json:"status"`
 	}{status}
 
 	body, err := s.Request("PATCH", EndpointUserSettings("@me"), data)
--- a/restapi.go
+++ b/restapi.go
@@ -322,7 +322,10 @@ func (s *Session) UserSettings() (st *Settings, err error) {
 
 // UserUpdateStatus update the user status
 // status   : The new status (Actual valid status are 'online','idle','dnd','invisible')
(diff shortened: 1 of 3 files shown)


restapi.go
```diff
@@ -320,6 +320,23 @@ func (s *Session) UserSettings() (st *Settings, err error) {
 	return
 }
 
+// UserUpdateStatus update the user status
+// status   : The new status (Actual valid status are 'online','idle','dnd','invisible')
+func (s *Session) UserUpdateStatus(status string) (st *Settings, err error) {
```
iopred: Consider making a type and constants for the valid values, like VerificationLevel, so that we can restrict the values of status.

robbix1206: The problem is actually you have 5 status.
"online","idle","dnd","invisible","offline".
However the Status can be set by the user as "online","idle","dnd" or "invisible"
And the library can see only Users as "online","idle","dnd" or "offline"
So even if there is a Status type the user can send a status bad type ("offline")
I will do that and make a verification in the function.

robbix1206:
So Status is now a type.
Look like it works ^^.

structs.go
```diff
@@ -254,9 +254,9 @@ type VoiceState struct {
 
 // A Presence stores the online, offline, or idle and game status of Guild members.
 type Presence struct {
-	User   *User  `json:"user"`
-	Status string `json:"status"`
-	Game   *Game  `json:"game"`
+	User   *User   `json:"user"`
+	Status *Status `json:"status"`
```
iopred: Status Status `json:"status"`

structs.go
```diff
@@ -292,20 +292,34 @@ type User struct {
...
+	EnableTtsCommand       bool               `json:"enable_tts_command"`
+	MessageDisplayCompact  bool               `json:"message_display_compact"`
+	ShowCurrentGame        bool               `json:"show_current_game"`
+	ConvertEmoticons       bool               `json:"convert_emoticons"`
+	Locale                 string             `json:"locale"`
+	Theme                  string             `json:"theme"`
+	GuildPositions         []string           `json:"guild_positions"`
+	RestrictedGuilds       []string           `json:"restricted_guilds"`
+	FriendSourceFlags      *FriendSourceFlags `json:"friend_source_flags"`
+	Status                 *Status            `json:"status"`
```
iopred: Status instead of *Status

robbix1206: Ok will fix it

PR #259, opened 2016-09-28
Remove API call in onVoiceStateUpdate

b1naryth1ef:
This removes an unnecessary API call to `/api/users/@me`, which was previously used to check the current user id against the VoiceStateUpdate's user id. However, we have this information from the ready payload stored in state, so lets just use that.

pls stop spam my api @bwmarrin @iopred 

kthx ❤️ 🙈

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -17,7 +17,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


bwmarrin:
@iopred we need to make sure that even if State tracking is disabled, we still store the user info from the Ready packet.

Issue #263, opened 2016-09-29
GuildBans() Doesn't Work Due To API Change


blackhole12:
The GuildBans() function will now return a huge list of blank user objects (all fields are blank or 0). After some investigation, I discovered that this is because of an **undocumented API change** in discord. In addition to the user, discord now returns a blank "reason" field, presumably for an upcoming feature.

Performing a raw `GET` call on the endpoint returns this:

``` json
[
  {
    "reason": null,
    "user": {
      "username": "\uff21\uff44\uff41\uff4d\uff53\uff4b\uff41",
      "discriminator": "2033",
      "id": "58950874608697344",
      "avatar": "965d9836d0d7c797ff85ea4b59a39342"
    }
  },
  {
    "reason": null,
    "user": {
      "username": "mr chin",
      "discriminator": "6001",
      "id": "58992348637429760",
      "avatar": "ff6eb86907f4e553f6e893b7fff94159"
    }
  },
  {
    "reason": null,
    "user": {
      "username": "DAYS BEFORE CUNT",
      "discriminator": "1828",
      "id": "63017505282596864",
      "avatar": "a8fab7da9a5c2fd458345ba759582dc2"
    }
  },
  {
    "reason": null,
    "user": {
      "username": "Sachi",
      "discriminator": "5443",
      "id": "64860778150236160",
      "avatar": "e020f3682acebdef8428a308b4849781"
    }
  },
  {
    "reason": null,
    "user": {
      "username": "moof",
      "discriminator": "7624",
      "id": "74900312636264448",
      "avatar": "5edb51278ad007f3d9353b312e806dd5"
    }
  },
  {
    "reason": null,
    "user": {
      "username": "Pikscast",
      "discriminator": "6656",
      "id": "81084716635131904",
      "avatar": "6495f16761acd7b13b34dbf10fcf5bb9"
    }
  },
  {
    "reason": null,
    "user": {
      "username": "Chris",
      "discriminator": "3063",
      "id": "81277025163149312",
      "avatar": "50cfe15411e30f2718d867b8a265a0b8"
    }
  },
  {
    "reason": null,
    "user": {
      "username": "wusawusle",
      "discriminator": "1759",
      "id": "82743072772849664",
      "avatar": "708b583a92c1bfc52b0ef06a5e4c1c57"
    }
  }
]
```

Obviously, in order for this to be unmarshalled, discordgo will have to return a `type struct { string; user }` instead of just a `user` object.

PR #264, opened 2016-09-29
Add a new structure for Banned Users of a Guild

robbix1206:
Try to fix issue #263

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -432,6 +432,12 @@ type GuildBan struct {
 	GuildID string `json:"guild_id"`
 }
 
+// A GuildBans stores data for a banned user.
+type GuildBans struct {
+	Reason string `json:"reason"`
+	User   *User  `json:"user"`
+}
+
 // A GuildEmojisUpdate stores data for a guild emoji update event.
 type GuildEmojisUpdate struct {
 	GuildID string   `json:"guild_id"`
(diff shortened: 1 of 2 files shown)


structs.go
```diff
@@ -432,6 +432,12 @@ type GuildBan struct {
 	GuildID string `json:"guild_id"`
 }
 
+// A GuildBans stores data for a banned user.
+type GuildBans struct {
```
iopred: This should be `GuildBan`

robbix1206: Can't be GuildBan because GuildBan specify a guild however the GuildBans function already specified the guild and so doesn't have the guild argument (maybe can merge them if i put `GuildID string \`json:"guild_id"\`` has optionnal.

iopred: I think we should change GuildBan to this structure.

We should add GuildID to the GuildBanAdd and GuildBanRemove structs, and change the embedded struct to User to match the discord api: 
https://discordapp.com/developers/docs/topics/gateway#guild-ban-add

So:
type GuildBanAdd struct {
  *User
  GuildID string
}

What do you think?

@bwmarrin?

iopred:
I have gone ahead and made the changes I requested in https://github.com/bwmarrin/discordgo/pull/276
Sorry for not merging this directly, but I thank you for bringing this to my attention!
<3 <3

PR #268, opened 2016-10-05
Fixes a couple of typos

austin1237:
casing matters for some reason tried BOT and it returns a 401.
Also got the name of the executable wrong in mytoken (damn copy paste...)


austin1237:
I can recommit this onto to dev or dev can pull from master either will work.

PR #273, opened 2016-10-19
Ratelimits

jonas747:
Handles the new ratelimit headers
- X-RateLimit-Remaining
- X-RateLimit-Reset
- X-RateLimit-Global

I stuck it in an internal package, hope that's okay.
This is probably not the best way to do the removal of minor url variables, the alternative is to redo the whole endpoint system.

As for the reliability of this, after sending around 1k requests non-stop i didn't encounter a single 429.

I pad the reset time with 1 second, because if i don't it will still occasionally run into 429's of the size 1ms - 50ms (yeah, i got ratelimited for 1 ms a lot...), padding with anything less than 1 second would still give those tiny ratelimits sometimes.

### DIFF
--- a/internal/ratelimit/ratelimit.go
+++ b/internal/ratelimit/ratelimit.go
@@ -1,0 +1,187 @@
+package ratelimit
+
+import (
+	"net/http"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+// Ratelimiter holds all ratelimit buckets
+type RateLimiter struct {
+	sync.Mutex
+	buckets         map[string]*Bucket
+	globalRateLimit time.Duration
+}
+
+// New returns a new RateLimiter
+func New() *RateLimiter {
+
+	return &RateLimiter{
+		buckets: make(map[string]*Bucket),
+	}
+}
+
+// getBucket retrieves or creates a bucket
+// ratelimiter is expected to be locked when calling this
+func (r *RateLimiter) getBucket(key string) *Bucket {
+
+	if bucket, ok := r.buckets[key]; ok {
+		return bucket
+	}
+
+	b := &Bucket{remaining: 1, r: r, Key: key}
+	r.buckets[key] = b
+	return b
+}
+
+// LockBucket Locks until a request can be made
+func (r *RateLimiter) LockBucket(path string) *Bucket {
+
+	bucketKey := ParseURL(path)
+
+	r.Lock()
+	b := r.getBucket(bucketKey)
+	r.Unlock()
+
+	b.mu.Lock()
+
+	// If we ran out of calls and the reset time is still ahead of us
+	// then we need to take it easy and relax a little
+	for b.remaining < 1 && b.reset.After(time.Now()) {
+		// Sleep for an extra 500ms incase of time slighly out of sync
+		// (i got ratelimited for 1 and 2 milliseconds a lot when testing...)
+		toSleep := b.reset.Sub(time.Now()) + time.Millisecond*500
+		time.Sleep(toSleep)
+	}
+
+	// Lock and unlock to check for global ratelimites after sleeping
+	r.Lock()
+	r.Unlock()
+
+	b.remaining--
+	return b
+}
+
+// Bucket represents a ratelimit bucket, each bucket gets ratelimited individually (-global ratelimits)
+type Bucket struct {
+	Key string
+
+	mu        sync.Mutex
+	remaining int
+	limit     int
+	reset     time.Time
+	r         *RateLimiter
+}
+
+// Release unlocks the bucket and reads the headers to update the buckets ratelimit info
+// and locks up the whole thing in case if there's a global ratelimit.
+func (b *Bucket) Release(headers http.Header) error {
+
+	defer b.mu.Unlock()
+	if headers == nil {
+		return nil
+	}
+
+	remaining := headers.Get("X-RateLimit-Remaining")
+	reset := headers.Get("X-RateLimit-Reset")
+	global := headers.Get("X-RateLimit-Global")
+	retryAfter := headers.Get("Retry-After")
+
+	// If it's global just keep the main ratelimit mutex locked
+	if global != "" {
+		parsedAfter, err := strconv.Atoi(retryAfter)
+		if err != nil {
+			return err
+		}
+
+		// Lock it in a new goroutine so that this isn't a blocking call
+		go func() {
+			// Make sure if several requests were waiting we don't sleep for n * retry-after
+			// where n is the amount of requests that were going on
+			sleepTo := time.Now().Add(time.Duration(parsedAfter) * time.Millisecond)
+
+			b.r.Lock()
+
+			sleepDuration := sleepTo.Sub(time.Now())
+			if sleepDuration > 0 {
+				time.Sleep(sleepDuration)
+			}
+
+			b.r.Unlock()
+		}()
+
+		return nil
+	}
+
+	// Update reset time if either retry after or reset headers are present
+	// Prefer retryafter because it's more accurate with time sync and whatnot
+	if retryAfter != "" {
+		parsedAfter, err := strconv.ParseInt(retryAfter, 10, 64)
+		if err != nil {
+			return err
+		}
+		b.reset = time.Now().Add(time.Duration(parsedAfter) * time.Millisecond)
+
+	} else if reset != "" {
+		unix, err := strconv.ParseInt(reset, 10, 64)
+		if err != nil {
+			return err
+		}
+
+		b.reset = time.Unix(unix, 0)
+	}
+
+	// Udpate remaining if header is present
+	if remaining != "" {
+		parsedRemaining, err := strconv.ParseInt(remaining, 10, 32)
+		if err != nil {
+			return err
+		}
+		b.remaining = int(parsedRemaining)
+	}
+
+	return nil
+}
+
+var (
+	minorVariables = []*regexp.Regexp{
+		// Snowflake
+		regexp.MustCompile(`permissions/[0-9]+`),
+		regexp.MustCompile(`pins/[0-9]+`),
+		regexp.MustCompile(`members/[0-9]+`),
+		regexp.MustCompile(`messages/[0-9]+`),
+		regexp.MustCompile(`users/[0-9]+`),
+		regexp.MustCompile(`roles/[0-9]+`),
+		regexp.MustCompile(`bans/[0-9]+`),
+		regexp.MustCompile(`integrations/[0-9]+`),
+		regexp.MustCompile(`application/[0-9]+`),
+
+		// Not snowflake (strings, hashes etc..)
+		regexp.MustCompile(`invites/[0-9a-zA-Z]+`),
+		regexp.MustCompile(`icons/[0-9a-zA-Z]+`),
+		regexp.MustCompile(`splashes/[0-9a-zA-Z]+`),
+		regexp.MustCompile(`emojis/[0-9a-zA-Z]+`),
+		regexp.MustCompile(`avatars/[0-9a-zA-Z]+`),
+	}
+)
+
+// ParseURL parses the url, removing everything not relevant to identifying a bucket.
+// such as minor variables
+func ParseURL(url string) string {
+
+	// Remove url parameters
+	noParam := strings.SplitN(url, "?", 2)[0]
+
+	for _, r := range minorVariables {
+
+		noParam = r.ReplaceAllStringFunc(noParam, func(s string) string {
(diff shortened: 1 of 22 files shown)


internal/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,185 @@
...
+	r.buckets[key] = b
+	return b
+}
+
+// LockBucket Locks until a request can be made
+func (r *RateLimiter) LockBucket(path string) *Bucket {
+
+	bucketKey := ParseURL(path)
+
+	r.Lock()
```
iopred: Use sync.RWMutex and call r.RLock/r.RUnlock here

structs.go
```diff
@@ -17,6 +17,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/bwmarrin/discordgo/internal/ratelimit"
```
iopred: Just move ratelimit into the dgo directory, not sure we need to have an internal package.

internal/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,185 @@
...
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+// Ratelimiter holds all ratelimit buckets
+type RateLimiter struct {
+	sync.Mutex
+	buckets         map[string]*Bucket
```
iopred: Might be better to have:
global *Bucket

instead of passing the rate limiter into a bucket, circular references feel wrong.

internal/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,185 @@
...
+		regexp.MustCompile(`icons/[0-9a-zA-Z]+`),
+		regexp.MustCompile(`splashes/[0-9a-zA-Z]+`),
+		regexp.MustCompile(`emojis/[0-9a-zA-Z]+`),
+		regexp.MustCompile(`avatars/[0-9a-zA-Z]+`),
+	}
+)
+
+// ParseURL parses the url, removing everything not relevant to identifying a bucket.
+// such as minor variables
+func ParseURL(url string) string {
```
iopred: Is it possible to pass the endpoint into the ratelimiter in DGO, it seems unfortunate to construct a URL, and then re-parse it with a regex to get the endpoint out.

internal/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,131 @@
+package ratelimit
```
iopred: Thank you! 👍

internal/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,185 @@
...
+	r.Lock()
+	b := r.getBucket(bucketKey)
+	r.Unlock()
+
+	b.mu.Lock()
+
+	// If we ran out of calls and the reset time is still ahead of us
+	// then we need to take it easy and relax a little
+	for b.remaining < 1 && b.reset.After(time.Now()) {
+		time.Sleep(b.reset.Sub(time.Now()))
```
iopred: I would like to figure out a way to rate limit without sleeping, however I realise that this is still an improvement, so let's keep that in mind for the future.

internal/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,185 @@
...
+
+	r.Lock()
+	b := r.getBucket(bucketKey)
+	r.Unlock()
+
+	b.mu.Lock()
+
+	// If we ran out of calls and the reset time is still ahead of us
+	// then we need to take it easy and relax a little
+	for b.remaining < 1 && b.reset.After(time.Now()) {
```
iopred: Does this need to be a for loop? It looks like you're sleeping enough to be fine.

internal/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,185 @@
...
+		time.Sleep(b.reset.Sub(time.Now()))
+	}
+
+	// Lock and unlock to check for global ratelimites after sleeping
+	r.Lock()
```
iopred: Why is this dance required?

internal/ratelimit/ratelimit.go
jonas747: Well getBucket will also modify the map if the bucket isn't there already

internal/ratelimit/ratelimit.go
jonas747: After sleeping, another request might have triggered a global ratelimit

iopred: Hmm, I worry about using mutex's for synchronization. Is there not a way to detect and sleep for max(bucketTime, globalTime)?

internal/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,185 @@
...
+		b.reset = time.Now().Add(time.Duration(parsedAfter) * time.Millisecond)
+
+	} else if reset != "" {
+		unix, err := strconv.ParseInt(reset, 10, 64)
+		if err != nil {
+			return err
+		}
+
+		// Add a second to account for time desync and such
+		b.reset = time.Unix(unix, 0).Add(time.Second)
```
iopred: Also, can you play with this value.. You mentioned a few milliseconds don't work, what about 100*time.Millisecond, 200? 500?

internal/ratelimit/ratelimit.go
jonas747: it dosen't

internal/ratelimit/ratelimit.go
jonas747: Well the problem is that while there may not be any global ratelimit when it started sleeping, there may be one after sleeping

internal/ratelimit/ratelimit.go
jonas747: You mean each bucket having a reference to a global bucket instead of the ratelimiter?

internal/ratelimit/ratelimit.go
iopred: It's a much bigger change, but I would add:
RequestWithBucketId()

Doing that instead of changing Request() would be good for backwards compatibilities sake.

That would allow you to modify the private request() to have a nilable bucket ID.

internal/ratelimit/ratelimit.go
iopred: Yeah exactly.

internal/ratelimit/ratelimit.go
jonas747: Some data i've collected about this:

with no added time it's around every 15~ request
with 250ms: every 100~ request
with 500ms: every 500~ requests
with 800ms: every 1000~ requests
with 900ms: every 2000~ requests

haven't encountered one with 1s yet, gonna let the test run for a while.

these are just rough averages done in my head, not very accurate.

My thoughts on causes for this:
 - system time desync between me and discord, my time is synced up using arch's ntp servers though.
 - unix time rounding on discords end (if they store the time in a higher precision than seconds, but round the time down to seconds when sending it to us)

I'll do some more testing on a vps.

iopred:
That's sufficient, 1s seems reasonable.

internal/ratelimit/ratelimit.go
jonas747: Was this how you meant?

restapi.go
```diff
@@ -269,7 +251,7 @@ func (s *Session) Logout() (err error) {
 // userID    : A user ID or "@me" which is a shortcut of current user ID
 func (s *Session) User(userID string) (st *User, err error) {
 
-	body, err := s.Request("GET", EndpointUser(userID), nil)
+	body, err := s.RequestWithBucketID("GET", EndpointUser(userID), nil, EndpointUser(""))
```
jonas747: I updated some of them to this, did you mean to avoid the strings.SplitN call when using Request as well?

jonas747:
Latest commit makes it use the date header instead of requiring synced time, also lowered the extra time padding down to 250ms with that

iopred:
Yeah exactly, just pass in the bucket ID directly, I suggest using
EndpointUsers, EndpointGuilds etc. as the ID's as they are already strings.

jonas747:
so just to be 100% sure, replace stuff like 
```go
body, err := s.Request("GET", EndpointGuild(guildID), nil)
```
with 
```go
body, err := s.RequestWithBucketID("GET", EndpointGuild(guildID), nil, EndpointGuild(guildID))
```
?

iopred:
body, err := s.RequestWithBucketID("GET", EndpointGuild(guildID), nil,
EndpointGuilds)

jonas747:
Then requests for different guilds would be rate limited with the same bucket, guildID is a major variable so they should be rate limited on their own according to [the docs](https://discordapp.com/developers/docs/topics/rate-limits)

iopred:
I see, I'm trying to remove `bucketID = strings.SplitN(urlStr, "?", 2)[0]` from RequestWithBucketID, will that be sufficient if you pass EndpointGuild(guildID) in?

If so, Looks good to me :) :+1:

jonas747:
the `strings.SpltiN` is in `Request(..)` for backwards compatibility, since that function dosent take a separate bucket id.

you mean like 
```go
s.RequestWithBucketID("GET", EndpointGuild(guildID), nil, EndpointGuild(guildID))
```
? if so yeah, that wont call `strings.SplitN`

iopred:
Yeah I don't think we can avoid it in Request.

On Oct 26, 2016 10:41 AM, "jonas747" <notifications@github.com> wrote:

the strings.SpltiN is in Request(..) for backwards compatibility, since
that function dosent take a separate bucket id.

you mean like

s.RequestWithBucketID("GET", EndpointGuild(guildID), nil,
EndpointGuild(guildID))

? if so yeah, that wont call strings.SplitN

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/bwmarrin/discordgo/pull/273#issuecomment-256423301>,
or mute
the thread

jonas747:
dunso

ratelimit.go
```diff
@@ -0,0 +1,158 @@
...
+
+	b.remaining--
+	return b
+}
+
+// Bucket represents a ratelimit bucket, each bucket gets ratelimited individually (-global ratelimits)
+type Bucket struct {
+	Key string
+
+	mu        sync.Mutex
```
iopred: Do: 

type Bucket struct {
sync.Mutex
}

Then you can simply call:

bucket.Lock()

ratelimit.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// If we ran out of calls and the reset time is still ahead of us
+	// then we need to take it easy and relax a little
+	if b.remaining < 1 && b.reset.After(time.Now()) {
+		time.Sleep(b.reset.Sub(time.Now()))
+
+	}
+
+	// Check for global ratelimits
+	r.global.mu.Lock()
```
iopred: With the comment below this can be changed to:

r.global.Lock()

ratelimit.go
```diff
@@ -0,0 +1,158 @@
...
+	limit     int
+	reset     time.Time
+	global    *Bucket
+}
+
+// Release unlocks the bucket and reads the headers to update the buckets ratelimit info
+// and locks up the whole thing in case if there's a global ratelimit.
+func (b *Bucket) Release(headers http.Header) error {
+
+	defer b.mu.Unlock()
```
iopred: b.Unlock() with the comment above

ratelimit.go
```diff
@@ -0,0 +1,158 @@
...
+			return err
+		}
+
+		// Lock it in a new goroutine so that this isn't a blocking call
+		go func() {
+			// Make sure if several requests were waiting we don't sleep for n * retry-after
+			// where n is the amount of requests that were going on
+			sleepTo := time.Now().Add(time.Duration(parsedAfter) * time.Millisecond)
+
+			b.global.mu.Lock()
```
iopred: etc. :)

ratelimit.go
```diff
@@ -0,0 +1,158 @@
...
+
+			sleepDuration := sleepTo.Sub(time.Now())
+			if sleepDuration > 0 {
+				time.Sleep(sleepDuration)
+			}
+
+			b.global.mu.Unlock()
```
iopred: etc.

jonas747:
Also done

iopred:
Sorry dude, one last request, could you add Buckets for Reactions/Webhooks <3

jonas747:
finito

iopred:
Phew! Thanks so much dude. This is awesome! <3

Distortions81:
The default rate limit response is seconds, not milliseconds.
https://discord.com/developers/docs/topics/rate-limits#more-precise-rate-limit-resets

CarsonHoffman:
@Distortions81 where exactly is the issue in this change with respect to that?

PR #277, opened 2016-11-05
Add GuildReady event emitted when all guilds have been lazy loaded.

### LINKED ISSUE #205: Need a way to detect that all guilds have been lazy loaded.
Possibly, change Session.DataReady to  =  true after all guilds are loaded.

Also, since Voice ready bool has moved, DataReady should probably be changed to just Session.Ready

Also, it might be good to create a event to emit when this happens as well.

alex-j-butler:
This pull request fixes #205, adding a new event that is emitted when all the guilds from the initial `Ready` event have been lazy loaded.

Not quite sure if events should be used in this way, or if they should be exclusively for Discord gateway events. Let me know if there are any changes I should make to this.

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -42,6 +42,8 @@ func New(args ...interface{}) (s *Session, err error) {
 		ShouldReconnectOnError: true,
 		ShardID:                0,
 		ShardCount:             1,
+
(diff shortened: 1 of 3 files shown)


alex-j-butler:
Oops, forgot to fork this from `develop`.

PR #278, opened 2016-11-05
Add GuildReady event emitted when all guilds have been lazy loaded.

### LINKED ISSUE #205: Need a way to detect that all guilds have been lazy loaded.

alex-j-butler:
This pull request fixes #205, adding a new event that is emitted when all the guilds from the initial Ready event have been lazy loaded.

Not quite sure if events should be used in this way, or if they should be exclusively for Discord gateway events. Let me know if there are any changes I should make to this.

Note: This is the same pull request as #277, I forgot to fork from the `develop` branch on that pull request though.

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -249,7 +249,7 @@ func (s *Session) initialize() {
 func (s *Session) onReady(se *Session, r *Ready) {
 
 	for _, g := range r.Guilds {
-		if g.Unavailable != nil && !*g.Unavailable {
+		if g.Unavailable != nil {
 			s.loadedGuildMap[g.ID] = false
 		}
 	}
@@ -282,6 +282,6 @@ func (s *Session) onGuildCreate(se *Session, gc *GuildCreate) {
 
 	// If guilds are fully lazy loaded, emit a 'GuildReady' evnet.
 	if fullyLoaded {
-		s.handle(GuildReady{})
+		s.handle(&GuildReady{})
 	}
 }
--- a/discord.go
+++ b/discord.go
@@ -44,6 +44,8 @@ func New(args ...interface{}) (s *Session, err error) {
 		ShardID:                0,
 		ShardCount:             1,
 		MaxRestRetries:         3,
+
+		loadedGuildMap: make(map[string]bool),
 	}
 
 	// If no arguments are passed return the empty Session interface.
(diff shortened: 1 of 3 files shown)


iopred:
The ready event will also fire on all follow on GUILD_CREATE events. Can we restrict this to happening once per READY event?

discord.go
```diff
@@ -240,11 +242,18 @@ func (s *Session) initialize() {
...
 	s.AddHandler(s.State.onReady)
 	s.AddHandler(s.State.onInterface)
+	s.AddHandler(s.onGuildCreate)
 }
 
 // onReady handles the ready event.
 func (s *Session) onReady(se *Session, r *Ready) {
 
+	for _, g := range r.Guilds {
+		if g.Unavailable == true {
```
iopred: if g.Unavailable {

discord.go
```diff
@@ -258,3 +267,25 @@ func (s *Session) onResumed(se *Session, r *Resumed) {
...
+// onGuildCreate handles the guild creation event.
+func (s *Session) onGuildCreate(se *Session, gc *GuildCreate) {
+	if _, ok := s.loadedGuildMap[gc.ID]; ok {
+		s.loadedGuildMap[gc.ID] = true
+	}
+
+	// Iterate over the guilds that must be loaded, and check if they all have been.
+	numLoadedGuilds := 0
+	for _, g := range s.loadedGuildMap {
+		if g == true {
```
iopred: if g {

discord.go
```diff
@@ -258,3 +267,25 @@ func (s *Session) onResumed(se *Session, r *Resumed) {
...
+			numLoadedGuilds++
+		}
+	}
+
+	// If guilds are fully lazy loaded, emit a 'GuildReady' event.
+	if len(s.loadedGuildMap) > 0 && numLoadedGuilds == len(s.loadedGuildMap) {
```
iopred: Do this `len(s.loadedGuildMap) > 0` at the top of the this function, an early escape would be good, and this seems like a good one :)

PR #280, opened 2016-11-06
Adding new permissions and fixing type widths

rchristi:
Fairly simple changes, just a few missing permissions since your master branch (and Admins being in the wrong location in the bitfield). I also think the choice of int for the permissions type-width is incorrect given that this is a bit-field, see my commit comment there.


iopred:
Half of this change is already in `develop`.

The first commit (ints) seems reasonable for 32 bit systems however, please PR that into the develop branch.

PR #282, opened 2016-11-07
Add support for the prune endpoint

VagantemNumen:
Adds functions to get the amount of members that could be pruned
and to prune members using the prune endpoint.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -890,7 +890,7 @@ func (s *Session) GuildPruneCount(guildID string, days uint32) (count uint32, er
 		Pruned uint32 `json:"pruned"`
 	}{}
 
-	uri := EndpointGuildPrune(guildID) + "?days=" + fmt.Sprintf("%d", days)
+	uri := EndpointGuildPrune(guildID) + fmt.Sprintf("?days=%d", days)
 	body, err := s.RequestWithBucketID("GET", uri, nil, EndpointGuildPrune(guildID))
 
 	err = unmarshal(body, &p)
--- a/restapi.go
+++ b/restapi.go
@@ -886,20 +886,8 @@ func (s *Session) GuildPruneCount(guildID string, days uint32) (count uint32, er
 		return
 	}
 
-	u, err := url.Parse(EndpointGuildPrune(guildID))
-	if err != nil {
-		return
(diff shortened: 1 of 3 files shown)


VagantemNumen:
I am fairly certain the account does not have proper permissions for ``TestGuildPrune()``.
It passes all the tests on my machine.
[discordgo_tests.log.txt](https://github.com/bwmarrin/discordgo/files/576828/discordgo_tests.log.txt)

restapi.go
```diff
@@ -874,6 +874,79 @@ func (s *Session) GuildRoleDelete(guildID, roleID string) (err error) {
...
+		err = errors.New("The number of days should be more than or equal to 1.")
+		return
+	}
+
+	u, err := url.Parse(EndpointGuildPrune(guildID))
+	if err != nil {
+		return
+	}
+
+	q := u.Query()
```
iopred: Should be fine just to do:

uri := EndpointGuildPrune(guildID) + "?days="+days

VagantemNumen: Looked ugly and the ``net/url`` package was already imported anyway.
Should I change?

VagantemNumen:
Done.

restapi.go
```diff
@@ -874,6 +874,71 @@ func (s *Session) GuildRoleDelete(guildID, roleID string) (err error) {
...
+	if days <= 0 {
+		err = errors.New("The number of days should be more than or equal to 1.")
+		return
+	}
+
+	p := struct {
+		Pruned uint32 `json:"pruned"`
+	}{}
+
+	uri := EndpointGuildPrune(guildID) + "?days=" + fmt.Sprintf("%d", days)
```
iopred: Might be cleaner:

uri := EndpointGuildPrune(guildID) + fmt.Sprintf("?days="%d", days)

iopred:
If the test is failing, it's probably better to leave it out :(

Otherwise we need @bwmarrin to change the testing bot permissions.

VagantemNumen:
Okay, I will comment it out.

VagantemNumen:
Though it is pretty weird that only one test is failing when the docs say they both require the same permissions.

PR #285, opened 2016-11-11
Clean up ordering of internal handlers.

iopred:


iopred:
I've been running this on Septapus for a couple of days, no issues.

PR #288, opened 2016-11-14
Support embed messages.

iopred:


restapi.go
```diff
@@ -1228,6 +1228,28 @@ func (s *Session) ChannelMessageSendTTS(channelID string, content string) (st *M
 	return s.channelMessageSend(channelID, content, true)
 }
 
+// ChannelMessageSendEmbed sends a message to the given channel with embedded data (bot only).
+// channelID : The ID of a Channel.
+// embed     : The embed data to send.
+func (s *Session) ChannelMessageSendEmbed(channelID string, embed *MessageEmbed) (st *Message, err error) {
```
VagantemNumen: Should not it be ``ChannelMessageWithEmbedSend`` or ``ChannelMessageEmbedSend`` or similar?
Sorry about commenting on a merged pull request. Just something I noticed.

iopred: I was being consistent with ChannelMessageSendTTS, but ChannelMessageSendWithEmbed might have been my choice.

VagantemNumen: Oh, that is fine then I suppose. I just found it odd, considering the naming of the other functions. And changing it now might be inconvenient for the people already using the implement.

marcelgcom:
You can send an embedded message along with regular content. Please add a content string parameter to this function and send { 'content': ... , 'embed': ... }.

restapi.go
marcelgcom: You can send an embedded message along with regular content. Please add a content string parameter to this function and send { 'content': ... , 'embed': ... }.

PR #300, opened 2016-12-11
Switch to V5 gateway

klausenbusk:
This switch the gateway to V5, and change the heartbeat logic to
get the heartbeat_interval from the new Hello opcode instant of
READY/RESUME event.
See: hammerandchisel/discord-api-docs#18

Fix: #220

### DIFF
--- a/event.go
+++ b/event.go
@@ -210,8 +210,6 @@ func (s *Session) onInterface(i interface{}) {
 		setGuildIds(t.Guild)
 	case *GuildUpdate:
 		setGuildIds(t.Guild)
-	case *Resumed:
-		s.onResumed(t)
 	case *VoiceServerUpdate:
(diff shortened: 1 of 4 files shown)


blaenk:
This actually fixed a problem I was having with voice websocket connections which would yield an error after about 2 minutes, consistently, forcing a reconnect. I was advised to test this PR before digging further into that issue and it's been about 10 minutes so far without any problems. I'll be testing this from now on since I have no alternative, but so far it seems perfect! 👌

blaenk:
I'm grateful! Talking to the devs it sounded like the issue was on my end and I was on my own, so I really appreciate your looking into it and solving it :)

iopred:
To be fair the devs did suggest trying this PR ;)

PR #303, opened 2016-12-28
Fix racey heartbeat sequence

jonas747:


wsapi.go
```diff
@@ -89,13 +90,14 @@ func (s *Session) Open() (err error) {
...
-	if s.sessionID != "" && s.sequence > 0 {
+	sequence := atomic.LoadInt32(s.sequence)
+	if s.sessionID != "" && sequence > 0 {
 
 		p := resumePacket{}
 		p.Op = 6
 		p.Data.Token = s.Token
 		p.Data.SessionID = s.sessionID
-		p.Data.Sequence = s.sequence
+		p.Data.Sequence = int(sequence)
```
iopred: Let's just change our internal sequence vars to int32 to avoid this casting. WDYT?

jonas747: Got me thinking, maybe int64 is better?

iopred: Yeah I also agree with that.

PR #304, opened 2016-12-28
Handle MessageDeleteBulk event

jonas747:


state.go
```diff
@@ -710,6 +710,12 @@ func (s *State) onInterface(se *Session, i interface{}) (err error) {
 		if s.MaxMessageCount != 0 {
 			err = s.MessageRemove(t.Message)
 		}
+	case *MessageDeleteBulk:
+		if s.MaxMessageCount != 0 {
+			for _, mID := range t.Messages {
+				s.MessageRemove(&Message{ID: mID, ChannelID: t.ChannelID})
```
iopred: If we changed MessageRemove to call messageRemoveById() this could be modified to not create a bunch of junk data.

jonas747: o7

state.go
```diff
@@ -519,12 +519,12 @@ func (s *State) MessageAdd(message *Message) error {
 }
 
 // MessageRemove removes a message from the world state.
-func (s *State) MessageRemove(message *Message) error {
+func (s *State) MessageRemove(channelID, messageID string) error {
```
iopred: Arghh sorry to be a butthole dude. But I didn't want to change the existing API.

jonas747: oh woops i misread what you wrote, sorry

state.go
```diff
@@ -519,11 +519,16 @@ func (s *State) MessageAdd(message *Message) error {
...
+func (s *State) MessageRemove(message *Message) error {
 	if s == nil {
 		return ErrNilState
 	}
 
+	return s.messageRemoveById(message.ChannelID, message.ID)
+}
+
+// messageRemoveById removes a message by channelID and messageID from the world state.
+func (s *State) messageRemoveById(channelID, messageID string) error {
```
iopred: lint is complaining this should be: messageRemoveByID

PR #306, opened 2017-01-02
added Around parameter to ChannelMessages (which resolves #293)

### LINKED ISSUE #293: Add around functionality to ChannelMessages
The discord api has the ability to specify an around parameter to the GET/channels/{channel.id}/messages call. Mostly it would be used for pulling the most recent set of messages _including_ the most recent on a channel in an end user client or the like. It should just be a small change to the ChannelMessages function such that aroundID can be passed and used just as beforeID and afterID currently are.

The other option I can think of is to change the function to make the around call to discord if both the before and after ids have values but that seems to be a bit complicated and counter intuitive to me.

I am willing to make any of the changes, I just wanted to check and see what people might prefer.

Moonlington:
Since i missed this for my current project, I thought i'd pr it myself

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1210,7 +1210,7 @@ func (s *Session) ChannelTyping(channelID string) (err error) {
 // limit     : The number messages that can be returned. (max 100)
 // beforeID  : If provided all messages returned will be before given ID.
 // afterID   : If provided all messages returned will be after given ID.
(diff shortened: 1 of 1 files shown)


bwmarrin:
Could you add a comment line describing the new parameter?

Moonlington:
Should be done now

bwmarrin:
Thanks for the PR! :)

PR #307, opened 2017-01-03
Fix all found race conditions, and dont send on nil channel

jonas747:
Some locking was incorrect, a lot was missing, i fixed all race conditions i could find using the `-race` detector on different bots.

I have tested shorter voice connections by having a bot connect to voice, play a 2 second long audio file, then disconnect, every 5 seconds for some hours.

Tested longer connections by having a music bot play a track on loop for a couple hours.

This will lower performance a bit, but it's better than to rely on undefined behaviour. 

This pr is probably gonna be annoying to go through with all the new locking, sorry about that.

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -100,11 +100,15 @@ func (v *VoiceConnection) Speaking(b bool) (err error) {
 	v.wsMutex.Lock()
 	err = v.wsConn.WriteJSON(data)
 	v.wsMutex.Unlock()
+
+	v.Lock()
+	defer v.Unlock()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


iopred:
Awesome, I'm going to run this on Septapus for a few days and make sure there's no deadlocks, then merge.

jonas747:
nice nice

jonas747:
you find any problems with it @iopred ?

iopred:
Ack, so sorry Jonas, Septapus is now running with this patch. This time tomorrow should be enough :P

PR #319, opened 2017-01-27
Add support for @me in GuildMemberNickname (Resolves #318)

### LINKED ISSUE #318: Can't easily change own nickname
I'm writing a bot using DiscordGo.  Attempting to have it change its nickname proved harder than expected, due to GuildMemberNickname throwing a 403 Permission error when I provided the bot's own ID as a target.  This is due to, as already referenced in #189, editing the current user's nickname needing an additional "/nick" in the API call, making it "/guild/{guild.id}/user/@me/nick" instead of "/guild/{guild.id}/user/{user.id}"
Maybe an enhancement can be made here?  Either adding a check for "@me" or the current user's user ID in GuildMemberNickname which would set the ID to be @me and add the "/nick" to the API call.  Or alternatively, making a different unique function altogether for editing the current user's nickname.
Thoughts?

notdisliked:


notdisliked:
I might have done this wrong I apologize I don't contribute to github often

PR #322, opened 2017-02-16
add ChannelEditTopic and ChannelEditObject

proxikal:
ChannelEditObject allows developers to have more freedom when editing channels.
ChannelEditTopic allows developers to change the channel topic.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1187,6 +1187,41 @@ func (s *Session) ChannelEdit(channelID, name string) (st *Channel, err error) {
 	return
 }
 
+// ChannelEditTopic edits the given channel
+// channelID  : The ID of a Channel
+// topic       : The new topic to assign the channel.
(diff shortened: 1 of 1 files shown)


bwmarrin:
The function should take defined struct instead of an interface.  Then we need to pull out only the valid fields that can be updated before passing it to Discord.  So we're not sending a bunch of junk.  Each field will need to be validated to some degree.

Take a look at the GuildEdit function for an example of how it's doing this same thing.

LEGOlord208:
@proxikal Any update on this?

Not like I need it, just kinda feel like it's been here for a while.

iopred:
I think https://github.com/bwmarrin/discordgo/pull/352/files is a good example of how to do this PR. We should have a new struct and have ChannelEditComplex. The struct should have pointer members that are all omitempty.

PR #323, opened 2017-02-16
ChannelEditTopic

LEGOlord208:
Literally just a copy. The other pull request (not from me) about this was closed, so this is another.


bwmarrin:
Most likely this will be solved with https://github.com/bwmarrin/discordgo/pull/322 by making ChannelEdit work like GuildEdit to accept a range of settable values.

LEGOlord208:
Yep, good idea!

PR #329, opened 2017-02-28
Added method to be able to log in using TOTP token

Sense545:
Coded this up in order to fix this issue https://github.com/Rivalo/discord-cli/issues/34 and thought I might as well submit this as a pull request here.

### DIFF
--- a/endpoints.go
+++ b/endpoints.go
@@ -35,6 +35,9 @@ var (
 	EndpointResetPassword  = EndpointAuth + "reset"
 	EndpointRegister       = EndpointAuth + "register"
 
+        EndpointAuthMfa        = EndpointAuth + "mfa/"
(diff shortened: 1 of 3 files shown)


bwmarrin:
Accessing the mfa/totp endpoint is strictly prohibited by Discord themselves and doing so reportedly will result in having the user account banned permanently from Discord.  I understand we can do this, but out of respect to Discord I'm not going to add this into DiscordGo.  If Discord changes their stance at a later time, then I would happily allow it then. 

For reference, see https://github.com/hammerandchisel/discord-api-docs/issues/69#issuecomment-223886862

However, I do want to thank you for the effort you put into this and the PR.  I truly do appreciate it!

Sense545:
Ah, no problem. Thanks for the reference.

PR #335, opened 2017-03-10
Added ChannelMessageSendEmbedWithMessage

Seklfreak:


restapi.go
```diff
@@ -1335,6 +1335,30 @@ func (s *Session) ChannelMessageSendEmbed(channelID string, embed *MessageEmbed)
 	return
```
iopred: Could you change this method to:

return s.ChannelSendEmbedWithMessage(channelID, "", embed)

seems like that should work.

Seklfreak:
Thanks. I applied your suggested change and tested it. It works.

LEGOlord208:
So... shouldn't we do this with edit as well?

PR #336, opened 2017-03-13
UserAvatar to accept user object.

LEGOlord208:
One of the most important thing with this library is that it does 1 request per function. You have 100% control over how many web requests get made.
UserAvatar breaks that.

UserAvatar now accepts a user, which not only makes you know how many web requests gets made, but might also save on web requests if you have an existing user object.


iopred:
Please PR into the develop branch, thank you!

PR #343, opened 2017-03-24
Replace global ratelimit bucket with atomic

jonas747:
Instead of a global ratelimit bucket with a mutex, atomically store the unix time in nanoseconds it resets.
Also did some general cleanup in the ratelimit code, the performance is pretty much the same.

### DIFF
--- a/ratelimit.go
+++ b/ratelimit.go
@@ -4,13 +4,14 @@ import (
 	"net/http"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


iopred:
Going to run this on Sep for a day, but looks great, thanks jonas!

jonas747:
Np my dude

iopred:
Looks good, thanks!

PR #345, opened 2017-04-01
Handle PresenceUpdate in state tracking

### LINKED ISSUE #328: State tracking misses some member updates
Specifically:

- If a user changes their username, the state retains their old username until the next update for that user.
- Users who go online/offline don't get added/removed from the list of guild members until the next update for that user.

For example, if a user is offline when the session (and hence the state tracking) starts, they'll only show up in the guild's member list once they've done something which causes an update for that user.

If you have a user with username `Foo` with no nickname set who is offline when the session starts, they won't start in the members list.
Comes online: still not in the members list.
Changes their nickname to `Bar` (so user=`Foo`, nick=`Bar`): enters the members list.
Changes their username to `Bar` (user=`Bar`, no nick): still in the members list as `Foo`.
Changes their nickname to `Bar` (user=`Bar`, nick=`Bar`): members list updated, now in list as `Bar`.
Goes offline: still in the members list.

GinjaNinja32:
Fixes #328

### DIFF
--- a/state.go
+++ b/state.go
@@ -714,6 +714,35 @@ func (s *State) onInterface(se *Session, i interface{}) (err error) {
 		if s.TrackVoice {
 			err = s.voiceStateUpdate(t)
 		}
+	case *PresenceUpdate:
+		if s.TrackMembers {
+			var m *Member
+			m, err = s.Member(t.GuildID, t.User.ID)
+
+			if err != nil {
+				// Member not found; this is a user coming online
+				m = &Member{
+					GuildID: t.GuildID,
+					Nick:    t.Nick,
+					User:    t.User,
+					Roles:   t.Roles,
+				}
+				err = s.MemberAdd(m)
+			} else {
+				if t.Nick != "" {
+					m.Nick = t.Nick
+				}
+
+				if t.User.Username != "" {
+					m.User.Username = t.User.Username
+				}
+
+				// PresenceUpdates always contain a list of roles, so there's no need to check for an empty list here
+				m.Roles = t.Roles
+
+				err = s.MemberAdd(m)
+			}
+		}
 	}
 
 	return
--- a/state.go
+++ b/state.go
@@ -716,6 +716,10 @@ func (s *State) onInterface(se *Session, i interface{}) (err error) {
 		}
 	case *PresenceUpdate:
 		if s.TrackMembers {
+			if t.Status == "offline" {
+				return
+			}
+
 			var m *Member
 			m, err = s.Member(t.GuildID, t.User.ID)
 


state.go
```diff
@@ -714,6 +714,35 @@ func (s *State) onInterface(se *Session, i interface{}) (err error) {
 		if s.TrackVoice {
 			err = s.voiceStateUpdate(t)
 		}
+	case *PresenceUpdate:
+		if s.TrackMembers {
+			var m *Member
+			m, err = s.Member(t.GuildID, t.User.ID)
```
iopred: Even though you'd be shadowing the err, it should still be fine to do:

m, err := s.Member.....

GinjaNinja32: Line 730 currently needs to assign the `err` defined in the function's signature, so that that error can be returned. That could be changed to `return s.MemberAdd(m)` if you'd prefer, and this to `:=`?

state.go
```diff
@@ -714,6 +714,35 @@ func (s *State) onInterface(se *Session, i interface{}) (err error) {
...
+
+				if t.User.Username != "" {
+					m.User.Username = t.User.Username
+				}
+
+				// PresenceUpdates always contain a list of roles, so there's no need to check for an empty list here
+				m.Roles = t.Roles
+
+				err = s.MemberAdd(m)
+			}
```
iopred: Small simplification, you could put `err = s.MemberAdd(m)` outside of the if else, as it is the same in both paths.

state.go
iopred: Ah understood. Looks good.

GinjaNinja32:
My bot's breaking with this code, not sure why, DNM until I figure out whether it's this code or something I did on my side.

GinjaNinja32:
Above commit seems to fix my empty-username issues - it appears Discord sent a `PresenceUpdate` for a user the state didn't know about going offline, which added an empty user to the state.

state.go
```diff
@@ -714,6 +714,39 @@ func (s *State) onInterface(se *Session, i interface{}) (err error) {
...
+
+			if err != nil {
+				// Member not found; this is a user coming online
+				m = &Member{
+					GuildID: t.GuildID,
+					Nick:    t.Nick,
+					User:    t.User,
+					Roles:   t.Roles,
+				}
+				err = s.MemberAdd(m)
```
ammario: this could become

`return s.MemberAdd(m)`

state.go
```diff
@@ -714,6 +714,39 @@ func (s *State) onInterface(se *Session, i interface{}) (err error) {
...
+			} else {
```
ammario: this could be outdented if previous recommendation is followed

jonas747:
There is 1 field that is missing in the presence update for members, joined_at, wouldn't this break existing code that expects this to be there for members in state?

iopred:
This was done in https://github.com/bwmarrin/discordgo/commit/f539136306da89c9110e9a340f34e2f09eeffc26 Thanks for getting this conversation started though <3 <3

PR #346, opened 2017-04-04
Added error when 2FA is enabled

LEGOlord208:
Did some reverse engineering on Discord's 2FA, and implemented it here.

BTW, have you any plans to merge master and develop?  
I wanna use the "stable" branch, but still have some of the later features :>

Thanks in advance!

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1319,7 +1319,7 @@ func (s *Session) ChannelMessageSendTTS(channelID string, content string) (st *M
 	return s.channelMessageSend(channelID, content, true)
 }
 
-// ChannelMessageSendEmbed sends a message to the given channel with embedded data (bot only).
+// ChannelMessageSendEmbed sends a message to the given channel with embedded data.
 // channelID : The ID of a Channel.
 // embed     : The embed data to send.
 func (s *Session) ChannelMessageSendEmbed(channelID string, embed *MessageEmbed) (st *Message, err error) {
@@ -1327,7 +1327,7 @@ func (s *Session) ChannelMessageSendEmbed(channelID string, embed *MessageEmbed)
 	return s.ChannelMessageSendEmbedWithMessage(channelID, "", embed)
 }
 
-// ChannelMessageSendEmbedWithMessage sends a message to the given channel with embedded data (bot only) and a message.
+// ChannelMessageSendEmbedWithMessage sends a message to the given channel with embedded data and a message.
 // channelID : The ID of a Channel.
 // content   : The message to send.
 // embed     : The embed data to send.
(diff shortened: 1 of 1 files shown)


Rivalo:
Given that Discord really doesn't like 2FA automation, and you'll actually get a ban for it. I doubt this is going to get into the DiscordGo Library.

LEGOlord208:
But if it's controlled by a user?  
I planned to use this in a command line application where the user types their code and everything...

Rivalo:
I find it irritating as well. Since people keep asking for the solution for this in my Command Line Application discord as well 😄 

But yeah: https://github.com/hammerandchisel/discord-api-docs/issues/69#issuecomment-223886862

They don't like you automating that "/api/auth/mfa/totp"
Given that text client is not really a self-bot or automating anything on the background, it might be okay. 

> Q: What about smaller/private bots? Bots that are hosted on my own user account that I use discord with. Will you ban those?
A: If they become a problem, yes. Be smart and use an official bot account.

The Docs are not conclusive on clients, since the API is mostly used for automated bots. So an external solution outside of the DiscordGo library might be the solution to this problem, still, a bug in the client could definitely be a problem from users of it down the line.

Rivalo:
https://github.com/hammerandchisel/discord-api-docs/issues/69#issuecomment-224030612

Hmm, they are actually conclusive on it. 
The text client going into a loop bug and thus making the user keep calling the 2FA endpoint will result in bans.

LEGOlord208:
Thanks a lot for the heads up.

I guess I'll close this pull. But first: Is there any chance DiscordGo could still keep some of it?  
So in case we need to **detect** 2FA, but not use it, that's possible?

Rivalo:
No clue, wait for @bwmarrin or another contributor I guess 😛

LEGOlord208:
I suppose. Thanks a lot! You have been very helpful :smile: 

Have we met before btw? I think I joined your discord server for a day or two. The discord-cli one :P

Rivalo:
I'm the original dev of discord-cli, so it could be :)
Normally I'm more on the background, since I currently don't really have the time to work on it, neither the environment to program on, but lately there have be tons of questions about 2FA.

bwmarrin:
Related : https://github.com/bwmarrin/discordgo/pull/329

I'm leery about adding any totp/2fa code into DiscordGo for fear of the wrath of Discord.  What portions are you suggesting we keep @LEGOlord208

LEGOlord208:
Sorry if I'm late. I suggest we keep the MFA and maybe Ticket things from New/Login, since the only way to accuire them currently would be to rewrite both functions.

LEGOlord208:
Oh, I'm so sorry. I were trying to get a local version of the repo when I accidently removed this pull, therefor automatically closing it.

Whatever... @bwmarrin, should I make another one?  
Or weren't you going to merge either way? :P

bwmarrin:
Well, coming back to see this now. I'm not sure without seeing the code :)  But if it doesn't touch the 2ta/totp endpoints, then I'm sure it'll go through.

LEGOlord208:
See #359

PR #348, opened 2017-04-05
It's not bot only though...

LEGOlord208:


bwmarrin:
I'm guessing that what you're saying is you tested this with a non-bot account, and it works just fine with normal user accounts as well? :)

LEGOlord208:
Definitely, yes. It has worked for ages, and both me and a friend sent tons today.

PR #349, opened 2017-04-05
ChannelMessageSendComplex

LEGOlord208:
I was getting tired of all the ChannelMessageSendThisWithThisAndThisAndThis.  
So, well, this is my solution.

I kept the others for compatibility. And `ChannelMessageSend` should always be kept for ease of use.

Anyways, I'm unsure whether I should also include `ChannelFileSend` in this as well. It'd certainly be useful (sending embeds and files together, maybe).  
But that's for later.

### DIFF
--- a/message.go
+++ b/message.go
@@ -36,7 +36,7 @@ type MessageSend struct {
 	Content string        `json:"content"`
 	Tts     bool          `json:"tts"`
 	Embed   *MessageEmbed `json:"embed"`
-	Nounce  string        `json:"nounce"`
+	Nonce   string        `json:"nonce"`
 }
 
 // MessageEdit stores all parameters you can send with ChannelMessageSendComplex.
--- a/message.go
+++ b/message.go
@@ -31,6 +31,14 @@ type Message struct {
 	Reactions       []*MessageReactions  `json:"reactions"`
 }
 
+// A MessageSend stores all parameters you can send with ChannelMessageSendComplex.
+type MessageSend struct {
+	Content string        `json:"content"`
+	Tts     bool          `json:"tts"`
+	Embed   *MessageEmbed `json:"embed"`
+	Nounce  string        `json:"nounce"`
+}
+
 // A MessageAttachment stores data for message attachments.
 type MessageAttachment struct {
 	ID       string `json:"id"`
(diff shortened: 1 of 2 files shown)


restapi.go
```diff
@@ -1332,23 +1324,7 @@ func (s *Session) ChannelMessageSendEmbed(channelID string, embed *MessageEmbed)
 // content   : The message to send.
 // embed     : The embed data to send.
 func (s *Session) ChannelMessageSendEmbedWithMessage(channelID string, content string, embed *MessageEmbed) (st *Message, err error) {
```
iopred: Good call, let's kill this method, it was added in develop, I don't care too much to keep it around.

iopred:
We should probably add EditComplex also.

LEGOlord208:
I was going to do that, but Edit doesn't have as many functions and combinations. At all.

restapi.go
```diff
@@ -1303,63 +1305,46 @@ func (s *Session) channelMessageSend(channelID, content string, tts bool) (st *M
...
-
-	// Send the message to the given channel
-	response, err := s.RequestWithBucketID("POST", EndpointChannelMessages(channelID), data, EndpointChannelMessages(channelID))
-	if err != nil {
-		return
-	}
-
-	err = unmarshal(response, &st)
-	return
+func (s *Session) ChannelMessageSendEmbedWithMessage(channelID string, content string, embed *MessageEmbed) (*Message, error) {
```
iopred: Let's just kill this guy, WDYT?

LEGOlord208: Personally, yes. But it deserves mercy, too. And because I'm the kind of person that likes compatibility.

iopred: I love me some compatibility, but this method never made it to master.

LEGOlord208: Ok you're right. I will remove them

message.go
```diff
@@ -31,6 +31,20 @@ type Message struct {
 	Reactions       []*MessageReactions  `json:"reactions"`
 }
 
+// MessageSend stores all parameters you can send with ChannelMessageSendComplex.
+type MessageSend struct {
+	Content string        `json:"content"`
+	Tts     bool          `json:"tts"`
+	Embed   *MessageEmbed `json:"embed"`
+	Nounce  string        `json:"nounce"`
```
iopred: This is `nonce`

LEGOlord208: Oohh, right. Thanks. Will fix!

PR #350, opened 2017-04-07
add String() method to User

ammario:


iopred:
It might be good to move the User struct into this file, similar to Message.go

user_test.go
```diff
@@ -0,0 +1,18 @@
+package discordgo
+
+import "testing"
+
+func TestUser(t *testing.T) {
+	t.Parallel()
+
+	t.Run("String()", func(t *testing.T) {
```
iopred: 1.6 doesn't support Run so Travis is breaking, could you just inline the inside of this method to TestUser? We don't **need** this test to run in parallel :)

user.go
```diff
@@ -0,0 +1,23 @@
...
+	Discriminator string `json:"discriminator"`
+	Token         string `json:"token"`
+	Verified      bool   `json:"verified"`
+	MFAEnabled    bool   `json:"mfa_enabled"`
+	Bot           bool   `json:"bot"`
+}
+
+//String returns a unique identifier of the form username#discriminator
+func (u *User) String() string {
+	return fmt.Sprintf("%v#%v", u.Username, u.Discriminator)
```
iopred: Sorry, last thing, %s is the correct way to format a string. `"%s#%s"` would work best here

user_test.go
ammario: Sure, done w/ e19d4f4acb53dd048c20bd9676e1f823a6b3b9d6. The t.Parallel call should work with 1.6, I used the `t.Run` for organization not parallelism

iopred:
Thanks for the PR! <3

Issue #351, opened 2017-04-09
Discussion: What should we do about this issue?


LEGOlord208:
Since my pull request #349, which adds ChannelMessageSendComplex and ChannelMessageEditComplex... it also raises a new... interesting problem.

ChannelMessageEditEmbed was apparently slightly changed.  
It now clears the previous message's contents.

Why does it do this?  
Well, I assume it's because the content is sent, even if it's empty.

**Which finally raises the question**:
We can obviously fix this with omitempty.  
_But do we want to?_  

I bet most people want ChannelMessageEditEmbed to clear the previous message,
but it also raises an impossibility to change it without querying the previous message.

If it's omitempty, I assume we can just leave a space, and Discord might even trim that to empty.  
Which works fine.

Any ideas?

iopred:
Change the message field to a *string and then we can omit on `nil` only.

LEGOlord208:
Good idea actually. Except... will the JSON library catch up on that?

iopred:
Yeah, sadly the way to use it is gross though, you have to have a reference to a string, and can't set it in a single expression:

https://play.golang.org/p/yNmBv6NqS8

Still kinda seems right.

iopred:
(we can probably have a 'SetMessage' function on the MessageEdit struct, would simplify the code:

e := discordgo.MessageEdit{}
e.SetContent("sup")
e.SetEmbed(&discordgo.MessageEmbed{})

LEGOlord208:
Good idea as well! Would kinda break a pattern though...

iopred:
Yeah, but this is the 'complex' version. :)

Luckily all of the existing wrapped callsites don't need to change, they should work fine.

LEGOlord208:
I'll do it later!

PR #352, opened 2017-04-09
SendContent

LEGOlord208:
See #351

### DIFF
--- a/message.go
+++ b/message.go
@@ -31,22 +31,27 @@ type Message struct {
 	Reactions       []*MessageReactions  `json:"reactions"`
 }
 
+// This is used in MessageSend and MessageEdit to share common parameters.
+type MessageParams struct {
+	Content *string       `json:"content,omitempty"`
+	Embed   *MessageEmbed `json:"embed"`
+}
+
 // MessageSend stores all parameters you can send with ChannelMessageSendComplex.
 type MessageSend struct {
-	MessageEdit
+	MessageParams
 	Tts bool `json:"tts"`
 }
 
 // MessageEdit stores all parameters you can send with ChannelMessageSendComplex.
 type MessageEdit struct {
-	Content *string       `json:"content,omitempty"`
-	Embed   *MessageEmbed `json:"embed"`
+	MessageParams
 }
 
 // SetContent is the same as setting the variable Content,
 // except it doesn't take a pointer.
 // Only a conveniance function.
-func (m *MessageEdit) SetContent(str string) {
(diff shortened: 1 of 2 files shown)


LEGOlord208:
One problem I see is that you need to set `MessageEdit: &MessageEdit{}` to be able to use SetContent.

Should I make a getter method? Should I remove the pointer, so it defaults to an empty struct?

LEGOlord208:
Since I accidently changed my local `develop` branch, I used this pull early for one project.

message.go
```diff
@@ -33,18 +33,24 @@ type Message struct {
 
 // MessageSend stores all parameters you can send with ChannelMessageSendComplex.
 type MessageSend struct {
-	Content string        `json:"content"`
-	Tts     bool          `json:"tts"`
-	Embed   *MessageEmbed `json:"embed"`
-	Nonce   string        `json:"nonce"`
+	MessageEdit
+	Tts   bool          `json:"tts"`
+	Embed *MessageEmbed `json:"embed"`
```
iopred: Embed is defined on MessageEdit, you shouldn't need this line.

LEGOlord208: Oo right! Missed that! 1 second!

message.go
```diff
@@ -33,18 +33,24 @@ type Message struct {
...
 }
 
 // MessageEdit stores all parameters you can send with ChannelMessageSendComplex.
 type MessageEdit struct {
```
iopred: Is there a better name here? Do we need the MessageEdit? What if we just document that Tts doesn't work when editing? In that way we could just have 'MessageParameters' as the only struct that both ComplexSend and ComplexEdit take.

LEGOlord208: Right, OK!

iopred:
Lint comment errors breaking the build, but this looks good.

LEGOlord208:
Already on it ;)

message.go
```diff
@@ -31,18 +31,28 @@ type Message struct {
 	Reactions       []*MessageReactions  `json:"reactions"`
 }
 
+// MessageParams is used in MessageSend and MessageEdit to share common parameters.
+type MessageParams struct {
+	Content *string       `json:"content,omitempty"`
+	Embed   *MessageEmbed `json:"embed"`
```
iopred: SORRY! Last thing! This should be omitempty also.

LEGOlord208: Oh, right! No problem!

iopred:
Thanks so much dude! 💯 💯 🥇

PR #354, opened 2017-04-11
Added more fields

LEGOlord208:


iopred:
We probably want these to be pointers like VerificationLevel, so we can determine which things to send to Discord.

Also omitempty wont do anything unless they are pointers... finally, this struct doesn't go over the network so it doesn't need any JSON tags.

LEGOlord208:
Pretty sure omitempty on a string makes it omit if empty. e.g. `== ""`

iopred:
https://play.golang.org/p/YFv6JMZsqZ

Sadly no :(

LEGOlord208:
That only affects pointers. If the string is not a pointer, it will omit empty.

iopred:
:100: 

Well my other comment still stands, the GuildParams struct doesn't need JSON attributes, unless you're planning on changing GuildEdit to just pass the struct along instead of copying all the fields.

LEGOlord208:
Ohhh it doesn't do that!  
Didn't realize, sorry

PR #355, opened 2017-04-12
API change

LEGOlord208:
The API seems to have changed to return an array now.


LEGOlord208:
Nevermind! It must be removed completely! Because the invite it gave was an existing one

iopred:
It could also be that that endpoint only works for Users, and not Bots.

LEGOlord208:
Meh, it's the opposite. Discord said on reddit they automatically unverifies your account if you try to use it (it's true, btw!), and then deleted their post.

IDK why they would do this, I love being able to use invites, but... Now it literally doesn't work and I have to re-verify my account. It f*cking sucks.

iopred:
Guh!

LEGOlord208:
_(Can't tell if you mean me or them)_

iopred:
Them, that sucks :cry:

PR #361, opened 2017-04-27
Added user notes

legolord208:
Only added the set endpoint, as that is the only one google told me.
I don't know if you can set them.

Please check if you think I set the correct ratelimiting endpoint.
I was maybe supposed to set it to EndpointUser...

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1847,6 +1847,20 @@ func (s *Session) MessageReactions(channelID, messageID, emojiID string, limit i
 }
 
 // ------------------------------------------------------------------------------------------------
+// Functions specific to user notes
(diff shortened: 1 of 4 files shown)


restapi.go
```diff
@@ -1847,6 +1847,20 @@ func (s *Session) MessageReactions(channelID, messageID, emojiID string, limit i
 }
 
 // ------------------------------------------------------------------------------------------------
+// Functions specific to user notes
+// ------------------------------------------------------------------------------------------------
+
+// UserNotesSet sets the note for a specific user.
+func (s *Session) UserNotesSet(userID string, message string) (err error) {
```
iopred: UserNoteSet seems correct here :)

PR #364, opened 2017-04-28
add Tag method to user

ammario:
Simplifies process of tagging a user in messages

### DIFF
--- a/user.go
+++ b/user.go
@@ -22,5 +22,5 @@ func (u *User) String() string {
 
 //Tag return a string which tags the user
 func (u *User) Tag() string {
-	return fmt.Sprintf("<@%v>", u.ID)
+	return fmt.Sprintf("<@%s>", u.ID)
 }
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -17,3 +19,8 @@ type User struct {
 func (u *User) String() string {
 	return u.Username + "#" + u.Discriminator
 }
+
+//Tag return a string which tags the user
+func (u *User) Tag() string {
+	return fmt.Sprintf("<@%v>", u.ID)
```
iopred: %s for strings.

iopred:
Sorry to be a butt, there's a couple of things we should do here:

a) We should probably call this `Mention()`
b) We do string concatenation in `String()` above, we should be consistent here instead of formatting.

diraven:
It's still Sprintf, not concatenation.

diraven:
Also in https://github.com/bwmarrin/discordgo/releases it still says function name is "Tag".

PR #366, opened 2017-04-28
STOP USING PRINTF PLS

jD91mZM2:


legolord208:
Not sure about the `regexp.QuoteMeta` usage in the latest commit. Thought it was safer to quote the ID, in case it would change from numbers for some reason. Not sure though, so... Yeah, should maybe be removed for less safety and more speed :racehorse:

qaisjp:
Why?

legolord208:
Because printf is evil. It has to loop through the string and parse it. Rust's `println` is better because it's at runtime. But since we don't have that in Go, it's better to just use the fastest way available - concatination.

And come on, `fmt.Sprintf("%s %s", "Hello", "World");` is REALLY ugly compared to `"Hello" + " " + "World"`

qaisjp:
Hmm.

I can't agree it looks ugly in every case, but I do agree that:

```
uri = fmt.Sprintf("%s?%s", uri, v.Encode())		
```

Is a bit ridiculous. Far more concise and easier on the eyes to do `uri += "?" + v.Encode()`.

Imo Sprintf should only be used when concatenating multiple times in a string. For example, [this would be reasonable](https://github.com/qaisjp/go-discord-irc/blob/37b3e544e3f3fb4580f593d2542ddddb91b4beb9/bridge/utils.go#L21).

I really don't agree with all your changes, but your PR is more likely to be merged if you make the changes to only the unnecessary ones (it being unnecessary is is just my opinion though)

legolord208:
While that does sound reasonable, remember printf literally:

- creates a new string writer
- loops through the string, checks for a lot of formatting options, haves a lot of variables left right and center, and writes to the string
- reads the string from the writer

Feels like way to much work to force on a user of the library.

restapi.go
```diff
@@ -1382,7 +1382,10 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
 
 		for i, file := range files {
 			h := make(textproto.MIMEHeader)
-			h.Set("Content-Disposition", fmt.Sprintf(`form-data; name="file%d"; filename="%s"`, i, quoteEscaper.Replace(file.Name)))
+			h.Set("Content-Disposition",
+				`form-data; name="file`+strconv.Itoa(i)+
+					`"; filename="`+quoteEscaper.Replace(file.Name)+
+					`"`)
```
robbix1206: I agree with legolord on that point printf just take performances and on most cases can be replaced, properly however that's true that sometimes It could make the code a bit less readable.
In the pull request I think the only case where printf could have sense is this one.

sn0w:
+1 for banning printf.
Trading off speed for usability is ok in projects but not in libraries.

Please merge this soon 🙏

restapi.go
```diff
@@ -1382,7 +1382,11 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
 
 		for i, file := range files {
 			h := make(textproto.MIMEHeader)
-			h.Set("Content-Disposition", fmt.Sprintf(`form-data; name="file%d"; filename="%s"`, i, quoteEscaper.Replace(file.Name)))
```
iopred: Not sure I agree with expanding this line.

voice.go
```diff
@@ -543,7 +544,7 @@ func (v *VoiceConnection) udpOpen() (err error) {
 		return fmt.Errorf("empty endpoint")
 	}
 
-	host := fmt.Sprintf("%s:%d", strings.TrimSuffix(v.endpoint, ":80"), v.op2.Port)
+	host := strings.TrimSuffix(v.endpoint, ":80") + ":" + strconv.Itoa(v.op2.Port)
```
iopred: Use formatInt to be consistent with above.

restapi.go
```diff
@@ -87,7 +87,7 @@ func (s *Session) request(method, urlStr, contentType string, b []byte, bucketID
 
 	req.Header.Set("Content-Type", contentType)
 	// TODO: Make a configurable static variable.
-	req.Header.Set("User-Agent", fmt.Sprintf("DiscordBot (https://github.com/bwmarrin/discordgo, v%s)", VERSION))
+	req.Header.Set("User-Agent", "DiscordBot (https://github.com/bwmarrin/discordgo, v"+VERSION)
```
iopred: You missed the closing bracket in the string.

voice.go
jD91mZM2: Itoa == FormatInt though...

b1naryth1ef:
> Trading off speed for usability is ok in projects but not in libraries.

This is wrong. Libraries should use best practices in having clear, concise and readable code. The performance different your looking at with optimizations like this is negligible to the language your working in.

> Because printf is evil.
If you want to PR a change that negatively impacts the readability of a library, you should justify it properly. If this is a PR for performance, include clear before and after benchmarks.

sn0w:
Of course printf isn't evil and I fully support using helpers like these when appropriate.
(Like an assignment that deals with a lot of different types or otherwise complicated expressions).
But you can't seriously accept things like these:
```go
fmt.Sprintf("%s?%s", uri, v.Encode())
```
```go
fmt.Sprintf("%s#%s", u.Username, u.Discriminator)
```
```go
fmt.Sprintf("DiscordBot (https://github.com/bwmarrin/discordgo, v%s)", VERSION))
```

> The performance difference your looking at with optimizations like this is negligible

All of these otherwise super simple string operations force the passed arguments to escape to the heap due to interface-boxing and cause quite some runtime overhead for parsing and applying the format-string. (Left aside the reflection overhead depending on the used formatter.)

Is that really worth it just to avoid typing that one extra cast or calling `.String()` "manually" ?

> Libraries should use best practices in having clear, concise and readable code

> If you want to PR a change that negatively impacts the readability of a library [...]

Of course style-related changes like these are prone to a lot of subjective discussion.
But this PR isn't just about style.
This is about a ridiculous amount of avoidable overhead for plain laziness or arguably nicer code.

And like I said: Avoidable overhead is something that separates the good and bad libraries.

marcelgcom:
Now I am definitely interested in seeing some benchmarks.

toby3d:
I make simple benchmarks and get this results:
```
BenchmarkPlus-4          1000000        1376 ns/op
    string(24)+"_hello"+"_world"

BenchmarkSprintf-4       1000000        1541 ns/op
    fmt.Sprintf("%d_%s_%s", 24, "hello", "world")

BenchmarkSprintln-4      1000000        1573 ns/op
    fmt.Sprintln(24, "hello", "world")

BenchmarkSprint-4        1000000        1585 ns/op
    fmt.Sprint(24, "_hello", "_world")

BenchmarkErrorf-4        1000000        1665 ns/op
    fmt.Errorf("24_hello_world")

BenchmarkErrors-4        1000000        1462 ns/op
    errors.New("24_hello_world")
```

sn0w:
Ok since there seems to be some demand for benchmarks, (and since toby's tests are disturbingly slow for some reason), I pushed a little write-up to github gist instead of cluttering this discussion even further.

Check it out at https://gist.github.com/sn0w/0013ca081ef6bfd86bf3d4a0a69e39d3 if you want.

toby3d:
@sn0w An important refinement - **without output/writing** (for example: `_ := fmt.Sprintf(1, 2, "3")`), the processing speed of data is really very different. If to make a log or record (for example: `txt := fmt.Sprintf(1, 2, "3") => log.Println(txt)`), then under loads the results practically do not differ.

toby3d:
[Same benchmarks](#issuecomment-332092003), but without any writing and output via logs:
```
BenchmarkPlus-4         2000000000           0.56 ns/op
BenchmarkSprintf-4      10000000         133 ns/op
BenchmarkSprintln-4     20000000         112 ns/op
BenchmarkSprint-4       10000000         155 ns/op
BenchmarkErrorf-4       10000000         170 ns/op
BenchmarkErrors-4       2000000000           0.82 ns/op
```
~~So fast, very wow, don't you?~~

sn0w:
@toby3d I did this without printing because none of jD91mZM2's proposed changes actually log stuff either. Why would we want to benchmark things that don't apply to our discussion?

> without output/writing the processing speed of data is really very different.

The difference is still the same. It's just shadowed by the printing overhead.
Again, that's why I tested that way.
We're comparing raw `sprintf()` and `+` performance.
Not the various ways to log things to console.

IceflowRE:
Something new here?

sn0w:
@IceflowRE nope

We have multiple arguments for and against the codestyle and multiple benchmarks that all show the obvious "`+` is faster" result. At this point it's really up to the maintainer(s) to decide if this PR has a future or not. I'd also love to hear a clear "yes this will be a thing someday" or "no we will never merge this" from them, but I guess we'll just have to play the waiting game.

bwmarrin:
@sn0w that's fair, and personally I'm in favour of the faster method.  It's clearly debatable on which style is easier to read, looks better, whatever.  I think one style looks better in some cases, the other in other cases, so I'm mixed on that.  The performance difference is pretty clear cut and that serves as a tie breaker for me.  So, this will very likely be merged and hopefully I can find free time to work on it soon.

qaisjp:
@bwmarrin there's nothing wrong with using both, using the correct ones where required ("consistency" isn't required here).

PR #367, opened 2017-04-28
Added reason go GuildBanCreate

jonas747:
yupyup


iopred:
Curse go and it's lack of overloading... I don't really want to break our API here.... GuildBanCreateWithReason? :|

jonas747:
sure sure

PR #368, opened 2017-04-29
Fix typo in description of VERSION constant

calmandniceperson:
It is called 'Semantic Versioning' and I thought it's a typo worth fixing.


bwmarrin:
Good catch, and thank you! Every fix is worth it :)

PR #369, opened 2017-04-29
Support for passing contexts to handlers

liclac:
Adds support for `func(*discordgo.Session, *discordgo.EVENT, context.Context)` handlers, and a factory function on Session to create passed contexts.

### DIFF
--- a/discord_test.go
+++ b/discord_test.go
@@ -1,6 +1,7 @@
 package discordgo
 
 import (
+	"context"
 	"os"
 	"runtime"
 	"sync/atomic"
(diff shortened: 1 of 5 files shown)


liclac:
Note: this drops support for Go 1.6, as that's the version that introduces contexts into the standard library.

One thing I'd kinda like to do in addition is to have a root context, which is attached to all HTTP requests + passed to ContextFactory, thus enabling cleaner shutdown, but I'm a little iffy on if that goes against the ideology of the library, and it's not as major as passing contexts to handlers, at least for me.

bwmarrin:
We'll hold off on this until at least some time after releasing v16.

bwmarrin:
I know it's been awhile @liclac but I'm finally catching up with life and that's letting me spend more time on this too :)

Can you give me a use case or two of what adding the context to the handlers would do for bot developers using this library?  If there's a pretty solid use case or two we could then refactor this PR and get it merged :)

treeder:
So you can setup the context once prior, instead of having to set it up every single time the handler is called. Also, it's just standard Go style to do this now.

iopred:
While you can pass state with Context, the primary purpose for Context is for cancellation, which isn't handled in this case.

A simple closure is going to give you all the benefit of this PR, with the added benefit of keeping type safety.

I am not a fan of this PR.

hvent90:
Passing context into the handlers can help with instrumentation of metrics and tracing libraries, such as [OpenCensus](https://opencensus.io) :)

enkeyz:
I know this is closed, but any reason not to use Context, like disgord does?

bwmarrin:
@enkeyz at this point it might make sense to open an issue with a proposal on adding context to just start a discussion and see where that goes.  Changing it would break everyone using the library so there would need to be a pretty compelling reason to make that change.

PR #370, opened 2017-04-30
ChannelFileSendWithComplex

legolord208:
What's a better way of doing this? Reflect? Generator? I'm open to suggestions

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1311,7 +1311,49 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
 		data.Embed.Type = "rich"
 	}
 
-	response, err := s.RequestWithBucketID("POST", EndpointChannelMessages(channelID), data, EndpointChannelMessages(channelID))
+	var response []byte
+	if data.File != nil {
+		body := &bytes.Buffer{}
+		bodywriter := multipart.NewWriter(body)
+
+		// What's a better way of doing this? Reflect? Generator? I'm open to suggestions
+
+		if err = bodywriter.WriteField("content", data.Content); err != nil {
+			return nil, err
+		}
+		var embed []byte
+		embed, err = json.Marshal(data.Embed)
+		if err != nil {
+			err = bodywriter.WriteField("embed", string(embed))
+			if err != nil {
+				return nil, err
+			}
+		}
+		if err = bodywriter.WriteField("tts", strconv.FormatBool(data.Tts)); err != nil {
+			return nil, err
+		}
+
+		var writer io.Writer
+		writer, err = bodywriter.CreateFormFile("file", data.File.Name)
+		if err != nil {
+			return nil, err
+		}
+
+		_, err = io.Copy(writer, data.File.Reader)
+		if err != nil {
+			return
+		}
+
+		err = bodywriter.Close()
+		if err != nil {
+			return
+		}
+
+		endpoint := EndpointChannelMessages(channelID)
+		response, err = s.request("POST", endpoint, bodywriter.FormDataContentType(), body.Bytes(), endpoint, 0)
+	} else {
+		response, err = s.RequestWithBucketID("POST", EndpointChannelMessages(channelID), data, EndpointChannelMessages(channelID))
+	}
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


iopred:
Just put an io.Reader as an arg in SendMessageParams? We don't need another complex method.

legolord208:
I would agree, but it uses a completely different `request` method, and content type.

iopred:
Then message send complex needs to get a bit more complex :)

legolord208:
@iopred You mean detecting if file is set, and if so use a completely different method? o_o

Though, how would the name be set? Have a file struct?

iopred:
We're likely to have a SetFile method on SendParams like we did on edit. So
I think it's fine to have Filename and FileReader on the params

legolord208:
I'll see what I can do tomorrow

legolord208:
@iopred :bell:

restapi.go
```diff
@@ -1311,7 +1311,49 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
...
+		bodywriter := multipart.NewWriter(body)
+
+		// What's a better way of doing this? Reflect? Generator? I'm open to suggestions
+
+		if err = bodywriter.WriteField("content", data.Content); err != nil {
+			return nil, err
+		}
+		var embed []byte
+		embed, err = json.Marshal(data.Embed)
+		if err != nil {
```
iopred: You're only writing the embed on failure, this embed write shouldn't be in an if scope.

legolord208: Oh oops! Nice catch!

restapi.go
```diff
@@ -1311,7 +1311,52 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
...
+			return
+		}
+
+		err = bodywriter.Close()
+		if err != nil {
+			return
+		}
+
+		endpoint := EndpointChannelMessages(channelID)
```
iopred: Sorry I'm being really pedantic here, could you move this up outside the if, and use it in the else below as well, I'm being all weird and finicky :)

legolord208: No problem! Fixed!

restapi.go
```diff
@@ -1311,7 +1311,53 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
...
+	endpoint := EndpointChannelMessages(channelID)
+
+	var response []byte
+	if data.File != nil {
+		body := &bytes.Buffer{}
+		bodywriter := multipart.NewWriter(body)
+
+		// What's a better way of doing this? Reflect? Generator? I'm open to suggestions
+
+		if err = bodywriter.WriteField("content", data.Content); err != nil {
```
iopred: You should only do this if content is != ""

restapi.go
```diff
@@ -1311,7 +1311,53 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
...
+			return
+		}
+
+		var embed []byte
+		embed, err = json.Marshal(data.Embed)
+		if err != nil {
+			return
+		}
+		err = bodywriter.WriteField("embed", string(embed))
```
iopred: Only if embed != nil

restapi.go
```diff
@@ -1311,7 +1311,53 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
...
+		if err != nil {
+			return
+		}
+
+		if err = bodywriter.WriteField("tts", strconv.FormatBool(data.Tts)); err != nil {
```
iopred: Probably only if tts is true.

(also, line 1337 👍  )

restapi.go
legolord208: I felt some sending, even if useless, parameters wasn't the biggest deal. I'll fix it though!

restapi.go
```diff
@@ -1311,7 +1311,59 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
...
+
+		if data.Embed != nil {
+			var embed []byte
+			embed, err = json.Marshal(data.Embed)
+			if err != nil {
+				return
+			}
+			err = bodywriter.WriteField("embed", string(embed))
+			if err != nil {
+				return
```
legolord208: Pointing out line 1337, @iopred? `return` where you came from... :rofl:

PR #371, opened 2017-04-30
Add Presence Tracking to State

Necroforger:
The presences in the state didn't seem to be updating so I added the methods
`func PresenceAdd`
`func PresenceRemove`
`func Presence`
and the bool
`TrackPresences`
to the State struct

and now update presences on the `PresenceUpdate` event
```go
case *PresenceUpdate:
	if s.TrackPresences {
		s.PresenceAdd(t.GuildID, &t.Presence)
        }
```

I used and adjusted the code suggested from
 [Handle PresenceUpdate in state tracking #345](https://github.com/bwmarrin/discordgo/pull/345)

### DIFF
--- a/state.go
+++ b/state.go
@@ -34,6 +34,7 @@ type State struct {
 	TrackMembers    bool
 	TrackRoles      bool
 	TrackVoice      bool
+	TrackPresences  bool
 
 	guildMap   map[string]*Guild
 	channelMap map[string]*Channel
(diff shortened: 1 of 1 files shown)


state.go
```diff
@@ -147,6 +149,107 @@ func (s *State) Guild(guildID string) (*Guild, error) {
...
+			}
+			if presence.User.Username != "" {
+				guild.Presences[i].User.Username = presence.User.Username
+			}
+
+			return nil
+		}
+	}
+
+	guild.Presences = append(guild.Presences)
```
iopred: append(guild.Presences, presence)

PR #374, opened 2017-05-01
Added state_test.go to test state functions.

Kminttech:
Wrote test cases for state_test.go


iopred:
Pleas run this through gofmt

iopred:
:( But tests were good.

PR #375, opened 2017-05-04
ContentWithMentionReplaced on roles and nicks (Fixes #208)

### LINKED ISSUE #208: Message.ContentWithMentionsReplaced() replace role and channel mentions.
Currently `Message.ContentWithMentionsReplaced()` only replaces the User mentions, but not Channel and/or Role mentions.
It maybe nice to have them replaced too.

legolord208:
Haven't tried if it works, so ¯\\\_(ツ)\_/¯


message.go
```diff
@@ -168,13 +167,42 @@ type MessageReactions struct {
 
 // ContentWithMentionsReplaced will replace all @<id> mentions with the
 // username of the mention.
-func (m *Message) ContentWithMentionsReplaced() string {
-	if m.Mentions == nil {
-		return m.Content
+func (m *Message) ContentWithMentionsReplaced(s *Session) (content string, err error) {
```
iopred: This is breaking the API, we should have a ContentWithMentionsReplacedWithSession and document that this will replace more data than above.

message.go
```diff
@@ -168,13 +167,42 @@ type MessageReactions struct {
...
+		if nick == "" {
+			nick = user.Username
+		}
+		content = regexp.MustCompile("<@!?"+regexp.QuoteMeta(user.ID)+">").ReplaceAllString(content, "@"+nick)
+	}
+	for _, role := range guild.Roles {
+		if !role.Mentionable {
+			continue
+		}
+		content = regexp.MustCompile("<@&"+regexp.QuoteMeta(role.ID)+">").ReplaceAllString(content, "@"+role.Name)
```
iopred: This is really unperformant, instead of iterating through all the roles on each message we should use regexp.ReplaceAllStringFunc instead here, search for <@&[.*?]> and then search for the correct role name only when a mention happens.

legolord208: Well, I just kept the same syntax as it previously used.

message.go
legolord208: True

message.go
iopred: The difference is before the mentions array tells us which things are mentioned, in this case we don't know that theres a role mention, and so we're iterating through each role and doing a regex.

legolord208: Oh... yeah that's true. Nice catch!

message.go
```diff
@@ -168,13 +167,56 @@ type MessageReactions struct {
...
+
+	if !s.StateEnabled {
+		content = m.ContentWithMentionsReplaced()
+		return
 	}
-	content := m.Content
+
+	channel, err := s.State.Channel(m.ChannelID)
+	if err != nil {
+		return
```
iopred: Should probably return ContentWithMentionsReplaced.

message.go
```diff
@@ -168,13 +167,56 @@ type MessageReactions struct {
...
+		nick := member.Nick
+		if nick == "" {
+			nick = user.Username
+		}
+		content = regexp.MustCompile("<@!?"+regexp.QuoteMeta(user.ID)+">").ReplaceAllString(content, "@"+nick)
 	}
-	return content
+
+	if patternRoles == nil {
+		patternRoles = regexp.MustCompile("<&[^>]*>")
```
iopred: Only compile this once, where you define the var, also I prefer a lazy regex:

<&[0-9]*>

message.go
```diff
@@ -168,13 +167,56 @@ type MessageReactions struct {
...
+	}
+
+	patternRoles.ReplaceAllStringFunc(content, func(mention string) string {
+		role, err := s.State.Role(channel.GuildID, mention)
+		if err != nil || !role.Mentionable {
+			return mention
+		}
+
+		return "<@&" + role.ID + ">"
```
iopred: Don't you want `return "@"+role.Name` here?

message.go
```diff
@@ -168,13 +167,56 @@ type MessageReactions struct {
...
+		content = regexp.MustCompile("<@!?"+regexp.QuoteMeta(user.ID)+">").ReplaceAllString(content, "@"+nick)
 	}
-	return content
+
+	if patternRoles == nil {
+		patternRoles = regexp.MustCompile("<&[^>]*>")
+	}
+
+	patternRoles.ReplaceAllStringFunc(content, func(mention string) string {
+		role, err := s.State.Role(channel.GuildID, mention)
```
iopred: This won't work right? You're passing `<&40394820394209>` here as the role id instead of just `40394820394209`, in bruxism I just do mention[3 : len(mention)-1]

message.go
legolord208: True, but what if Discord decides not to use 0-9 or something weird like that?

message.go
legolord208: Right, yeah, that's true. Will correct!

message.go
legolord208: lol oops

message.go
iopred: id's are actually numbers, we just use strings internally.

legolord208:
It does require a session to get the channel :thinking:

iopred:
Oh, there is! Yeah definitely use MentionRoles instead of the regexp, but use similar regexp code for MentionRooms.

You can make a fake Session/State:

fakeState = &State{}
fakeState.Add.....
fakeSession = &Session{State:fakeState}

message.ContentWithMoreMentionsReplaced(fakeSession)

legolord208:
Yep, trying to fake a session and state already :stuck_out_tongue:

message.go
legolord208: I know. But we never know if they decide to change them. Or, we do know that that's not gonna happen. Still. Better be paranoid :stuck_out_tongue:

message.go
```diff
@@ -168,13 +168,68 @@ type MessageReactions struct {
...
+		if err != nil || !role.Mentionable {
+			continue
+		}
+
+		content = strings.Replace(content, "<&"+role.ID+">", "@"+role.Name, -1)
 	}
-	return content
+
+	if patternChannels == nil {
+		patternChannels = regexp.MustCompile("<#[^>]*>")
```
iopred: Compile this once.

message.go
```diff
@@ -168,13 +168,68 @@ type MessageReactions struct {
...
 // username of the mention.
-func (m *Message) ContentWithMentionsReplaced() string {
-	if m.Mentions == nil {
-		return m.Content
+func (m *Message) ContentWithMentionsReplaced() (content string) {
+	content = m.Content
+
+	for _, user := range m.Mentions {
+		content = strings.NewReplacer(
+			"<@"+user.ID+">", "@"+user.Username,
```
iopred: This should be user.Username:
https://discordapp.com/developers/docs/resources/channel#message-formatting

Only the @! replaces with nick.

message.go
iopred: Nope, being paranoid means you litter your code with defensiveness that is unwarranted.

Discord uses snowflakes as their ID's which are 64bit ints and have huge amounts of infrastructure around it. In fact we are actually the bad guys because we don't define our ID's as strings (which python does for example).

At worst you should be using (.*?), but [0-9]* is correct.

message.go
legolord208: Realized that, but didn't bother to fix it, because for some reason I thought I'd have to make a bunch of workarounds. Didn't realize it was as easy as changing that :stuck_out_tongue:

message.go
```diff
@@ -168,13 +168,68 @@ type MessageReactions struct {
...
+		content = m.ContentWithMentionsReplaced()
+		return
 	}
-	content := m.Content
+
 	for _, user := range m.Mentions {
-		content = regexp.MustCompile(fmt.Sprintf("<@!?(%s)>", user.ID)).ReplaceAllString(content, "@"+user.Username)
+		member, err := s.State.Member(channel.GuildID, user.ID)
+		nick := user.Username
+		if (err == nil && member.Nick != "") || nick == "" {
```
iopred: I think the `nick == ""` check is unneeded, a Member will always have a Username.

iopred:
Thanks for all the work on this guy <3

message.go
legolord208: That's true. No idea why I put that there :laughing:

message.go
iopred: I think you missed this, move this to line 183 and use `"<#[0-9]*>"`

legolord208:
Reason I didn't move it at first was because I thought only initiating it if necessary would be better. But I guess all global variables in Go are lazily initiated anyways, so... ¯\_(ツ)_/¯

And then now I realized it could cause race conditions. Better just move it where it should be

iopred:
Thanks again!

PR #376, opened 2017-05-05
Prevent UserColor from modifying the guild.Roles slice

Necroforger:
Edit: I don't think I PR'd this correctly, but copying the roles like this fixed the issue for me.
Its at line 931
```go
roles := make([]*Role, len(guild.Roles))
copy(roles, guild.Roles)
sort.Sort(Roles(roles))
```

Input
```go
	case ">usercolor":
		guild, _ := b.Guild(m)
		fmt.Println("--------Roles before calling UserColor-------")
		for _, v := range guild.Roles {
			fmt.Println(v.Position)
		}
		b.DG.State.UserColor(m.Author.ID, m.ChannelID)
		fmt.Println("--------Roles after calling UserColor--------")
		for _, v := range guild.Roles {
			fmt.Println(v.Position)
		}
```
Output:
image

In this example, calling UserColor has the side effect of also sorting the roles in a guild.

### DIFF
--- a/state.go
+++ b/state.go
@@ -34,6 +34,7 @@ type State struct {
 	TrackMembers    bool
 	TrackRoles      bool
 	TrackVoice      bool
+	TrackPresences  bool
 
 	guildMap   map[string]*Guild
 	channelMap map[string]*Channel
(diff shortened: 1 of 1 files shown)


iopred:
Yeah, you will need to repush this branch with the new updates from develop.

Is there an issue with this? I struggled with this when sorting them originally, but it seemed fine.

PR #381, opened 2017-05-08
Fix presence unmarshal with game names as numbers

jonas747:
When this fails decoding, it wouldn't even set the id of the guild for me, thought i was crazy at first when i saw guilds without ids, heh

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -304,7 +304,7 @@ type Game struct {
 // UnmarshalJSON unmarshals json to Game struct
 func (g *Game) UnmarshalJSON(bytes []byte) error {
 	temp := &struct {
-		Name string          `json:"name"`
+		Name json.RawMessage `json:"name"`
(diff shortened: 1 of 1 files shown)


structs.go
```diff
@@ -304,16 +304,16 @@ type Game struct {
 // UnmarshalJSON unmarshals json to Game struct
 func (g *Game) UnmarshalJSON(bytes []byte) error {
 	temp := &struct {
-		Name string          `json:"name"`
+		Name json.Number     `json:"name"`
```
iopred: This will fail on actual strings though? I think some RawMessage shenanigans is whats needed.

jonas747: image
it won't fail

iopred: Put "Dota 2" as one of those fields.

jonas747: all good 
image

PR #388, opened 2017-05-29
Methods to obtain a user's avatar URL

Necroforger:
Makes obtaining the avatar URL of a user easier.

### DIFF
--- a/user.go
+++ b/user.go
@@ -1,6 +1,9 @@
 package discordgo
 
-import "fmt"
+import (
+	"fmt"
+	"strings"
+)
 
 // A User stores all data for an individual Discord user.
 type User struct {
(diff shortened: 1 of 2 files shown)


user.go
```diff
@@ -24,3 +27,21 @@ func (u *User) String() string {
...
+// UserAvatarURL returns a URL to the requested user's avatar
+//		userID:   The ID of the user to get
+//		avatarID: The ID of the user's avatar
+//		size:     The size of the user's avatar as a power of two
+func UserAvatarURL(userID, avatarID, size string) string {
+	extension := ".jpg"
+	if strings.HasPrefix(avatarID, "a_") {
+		extension = ".gif"
+	}
+	return EndpointCDNAvatars + userID + "/" + avatarID + extension + "?size=" + size
```
iopred: Hmm, I would prefer:
```
func (u *User) AvatarURL(size string) string {
  var url string
  if strings.HasPrefix(avatarID, "a_) {
    url = EndpointUserAvatarAnimated(userID, avatarID)
  } else {
    url = EndpointUserAvatar(userID, avatarID)
  }
  return url + "?size=" + size
}
```

We already support EndpointUserAvatar, adding an animated one seems reasonable.

Necroforger: I noticed that the `EndpointUserAvatar` function requests a png image. Does the file format change to a png when you request it as one?

Necroforger: Updated the code things you suggested, apparently it does return a PNG file.

PR #392, opened 2017-06-04
Add GatewayBot

jonas747:


jonas747:
so the test fails because it's a bot only endpoint, there is some other bot only and user only endpoints as well though, so probably gonna need 2 sessions for testing. should i remove the test for now?

restapi.go
```diff
@@ -1685,6 +1685,28 @@ func (s *Session) Gateway() (gateway string, err error) {
 	return
 }
 
+// Gateway returns the websocket Gateway address and the reccomended number of shards
```
acdenisSK: s/reccomended/recommended

jonas747: No Hablo Ingles

iopred:
@bwmarrin could you create a new bot account so we can add this test?

bwmarrin:
I can do that :)

bwmarrin:
I just added a Bot account under the dgtest account.  I added the token to travis with the env `DGB_TOKEN`

jonas747:
It's still complaining it's a bot only endpoint

bwmarrin:
I think the code needs to change so that it uses the dgbot session that your other PR added.

bwmarrin:
Hmn, well that token is definitely a Bot account.  I added the "Bot " prefix to the travis settings and now it's got a new error.  HTTP 401 Unauthorized, {"code": 0, "message": "401: Unauthorized"}

jonas747:
Have you connected to the gateway once with that token?

bwmarrin:
I've ran the discordgo ping pong example to test it.. You can of course try it locally, it shows the token in the Travis details link for the failed build

jonas747:
Ran fine using that token on my machine 
image

PR #396, opened 2017-06-28
Add heartbeat ACK response and error handling

blackhole12:
- Error when sending a heartbeat now triggers a reconnection
- Op7 now triggers a reconnection
- Session now tracks the last heartbeat ACK that was recieved. If the
last ACK is more than FailedHeartbeatAcks*heartbeatinterval in the past,
this is treated as a dead connection and a reconnection is forced.

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -131,6 +131,7 @@ func (s *Session) Open() (err error) {
 	// lock.
 	s.listening = make(chan interface{})
 	go s.listen(s.wsConn, s.listening)
+	s.LastHeartbeatAck = time.Now().UTC()
 
 	s.Unlock()
 
@@ -199,6 +200,9 @@ type helloOp struct {
 	Trace             []string      `json:"_trace"`
 }
 
+// Number of heartbeat intervals to wait until forcing a connection restart.
+const FailedHeartbeatAcks time.Duration = 5
+
 // heartbeat sends regular heartbeats to Discord so it knows the client
 // is still connected.  If you do not send these heartbeats Discord will
 // disconnect the websocket connection after a few seconds.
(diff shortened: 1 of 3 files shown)


wsapi.go
```diff
@@ -215,16 +219,22 @@ func (s *Session) heartbeat(wsConn *websocket.Conn, listening <-chan interface{}
...
 		s.log(LogInformational, "sending gateway websocket heartbeat seq %d", sequence)
 		s.wsMutex.Lock()
 		err = wsConn.WriteJSON(heartbeatOp{1, sequence})
 		s.wsMutex.Unlock()
-		if err != nil {
-			s.log(LogError, "error sending heartbeat to gateway %s, %s", s.gateway, err)
-			s.Lock()
-			s.DataReady = false
-			s.Unlock()
+		if err != nil || time.Now().UTC().Sub(last) > (i*FailedHeartbeatAcks*time.Millisecond) {
```
iopred: Time.Sub() returns time in nanoseconds. At a glance, multiplying by time.Milliseconds looks wrong. Is there a way that you can make this i*FailedHeartbeatAcks and pull up the Millisecond value up into the constant, also verify/test this works.

wsapi.go
```diff
@@ -215,16 +219,22 @@ func (s *Session) heartbeat(wsConn *websocket.Conn, listening <-chan interface{}
 	defer ticker.Stop()
 
 	for {
+		s.RLock()
```
iopred: You need a write lock here.

wsapi.go
```diff
@@ -398,7 +408,15 @@ func (s *Session) onEvent(messageType int, message []byte) {
...
 	// Must immediately disconnect from gateway and reconnect to new gateway.
 	if e.Operation == 7 {
-		// TODO
+		s.log(LogInformational, "Closing and reconnecting in response to Op7")
+		err := s.Close()
+		if err != nil {
+			s.log(LogWarning, "error closing session connection, %s", err)
+		}
+
+		s.log(LogInformational, "calling reconnect() now")
```
iopred: Not sure this is needed, the log above seems like enough.

wsapi.go
blackhole12: This has been tested and it works. The reason is that because this is a duration, multiplying by Millisecond actually turns converts the value into the appropriate number of nanoseconds, which is why you can compare any duration to any other duration and always have it work (because they're ALL represented in nanoseconds). I can pull the millisecond amount up into the constant for readability, though.

wsapi.go
blackhole12: This is not writing anything, it's reading a value. Why would i need a write lock?

iopred: Doh, my bad.

wsapi.go
iopred: Hmm, looks like `i` is not in nanoseconds, we're just Unmarshalling into it from a msec value.

Your change seems right, but would you mind renaming the argument in this method to `heartbeatIntervalMsec` instead of i, thanks so much.

iopred:
Thanks for the PR!

PR #397, opened 2017-06-28
Develop

blackhole12:
DataReady was only used in a voice reconnect function. This seemed unnecessary, because a failure to reconnect would result in retrying anyway, but **I have not been able to test this**. Of more concern is the fact that the function itself is extremely hacky and probably needs
to be fixed up anyway.

This pull request demonstrates what is necessary to remove DataReady, but it might be more appropriate to bundle this change with a pull request that cleans up voice connections in general.


blackhole12:
Canceling because github wants me to make a new branch for a new pull request

PR #398, opened 2017-07-02
Hardcoded reactions ratelimit

Necroforger:
Sets a hardcoded ratelimit for reactions.

```go
 func NewRatelimiter() *RateLimiter {
 
  	return &RateLimiter{
  		buckets: make(map[string]*Bucket),
  		global:  new(int64),
 		customRateLimits: []*customRateLimit{
 			&customRateLimit{
 				suffix:   "//reactions//",
 				requests: 1,
 				reset:    200 * time.Millisecond,
 			},
 		},
  	}
  }
```

### DIFF
--- a/ratelimit.go
+++ b/ratelimit.go
@@ -1,7 +1,6 @@
 package discordgo
 
 import (
-	"errors"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


iopred:
End users should never be setting ratelimit timings, why don't we just change the rate limit for reactions?

ratelimit.go
```diff
@@ -1,28 +1,89 @@
...
+		}
+	}
+
+	if !found {
+		return errors.New("err: custom rate limit not found")
 	}
+
+	for _, b := range r.buckets {
+		if b.customRateLimit != nil && b.customRateLimit.suffix == suffix {
+			b.customRateLimit = nil
```
jonas747: Race condition here, write to customRatelimit can occur at the same time as a read on L160

iopred:
Sorry, I should have been more explicit with my last comment.

What problem is this PR trying to solve?

Necroforger:
Just wanted to add a way to fix the reactions ratelimit.

The ratelimit discord gives for reactions is incorrect.
it sends the ratelimit as 1/1s when it is actually 1/250ms.
So its a pain to add reactions to a message in bulk if you wanted to create something like a paginator with reactions.

Necroforger:
Another solution, rather than implementing custom rate limits, would be just to add this line to the `Release `function in `ratelimit.go`.

```go
// Release unlocks the bucket and reads the headers to update the buckets ratelimit info
// and locks up the whole thing in case if there's a global ratelimit.
func (b *Bucket) Release(headers http.Header) error {
	defer b.Unlock()

	if strings.HasSuffix(b.Key, "//reactions//") {
		b.reset = time.Now().Add(200 * time.Millisecond)
		b.remaining = 0
		return nil
	}
```

bwmarrin:
I'm pretty conflicted with this. 

I definitely do not want to expose any ability to interact with ratelimits to the enduser.  I also really hate the idea of setting a custom hard-coded rate limit but since apparently Discord isn't going to do anything to fix this (https://github.com/hammerandchisel/discord-api-docs/issues/182)... that's the only plausible option. 

So, I will accept a PR that cleanly hard-codes the reactions ratelimit :)

iopred:
Thank you!

PR #403, opened 2017-07-08
Add bot account to testing

jonas747:
Felt like this should be in its own pr instead of jamming it into the gatewaybot one


bwmarrin:
Thanks @jonas747  :)

PR #411, opened 2017-07-24
Add memberMap to State to speed up member queries

blackhole12:
This adds `memberMap` to the State object, in the same vein as the existing `guildMap` and `channelMap`, except it's a map of maps, mapping each guild ID to a map of member IDs paired to members. The map is replaced if necessary in `GuildAdd` so that all the pointers are always up-to-date, and is created in the `OnReady` function.

The Add, Remove, and Get functions have been updated to use the map, but still modify the slice if necessary, so the Member slice is still kept up to date. Speed tests indicate this approach is roughly 4-5 times faster for servers with 500+ members in them.

### DIFF
--- a/state.go
+++ b/state.go
@@ -89,7 +89,6 @@ func (s *State) GuildAdd(guild *Guild) error {
 
 	// If this guild contains a new member slice, we must regenerate the member map so the pointers stay valid
 	if guild.Members != nil {
-		delete(s.memberMap, guild.ID)
 		s.createMemberMap(guild)
 	}
 
@@ -349,7 +348,7 @@ func (s *State) Member(guildID, userID string) (*Member, error) {
 
 	members, ok := s.memberMap[guildID]
 	if !ok {
-		return nil, errors.New("guild not found")
+		return nil, ErrStateNotFound
 	}
 
 	m, ok := members[userID]
--- a/state.go
+++ b/state.go
@@ -42,6 +42,7 @@ type State struct {
 
 	guildMap   map[string]*Guild
 	channelMap map[string]*Channel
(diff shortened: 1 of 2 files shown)


state.go
```diff
@@ -77,6 +87,12 @@ func (s *State) GuildAdd(guild *Guild) error {
 		s.channelMap[c.ID] = c
 	}
 
+	// If this guild contains a new member slice, we must regenerate the member map so the pointers stay valid
+	if guild.Members != nil {
+		delete(s.memberMap, guild.ID)
```
iopred: Probably no benefit to deleting it first, it'll get overwritten in createMemberMap.

state.go
```diff
@@ -312,18 +344,17 @@ func (s *State) Member(guildID, userID string) (*Member, error) {
...
 	s.RLock()
 	defer s.RUnlock()
 
-	for _, m := range guild.Members {
-		if m.User.ID == userID {
-			return m, nil
-		}
+	members, ok := s.memberMap[guildID]
+	if !ok {
+		return nil, errors.New("guild not found")
```
iopred: Return ErrNotFound here I guess.

iopred:
Awesome, thanks for this PR! 💓

state.go
blackhole12: This was debug code used to differentiate the guild not found error during testing, reverting it to ErrNotFound.

blackhole12:
@iopred The test was adding a guild via GuildAdd() without having initialized it via OnReady, which wasn't properly handled. I added a check that initializes the member map if it doesn't exist in GuildAdd to ensure there is no possible case where it isn't initialized.

iopred:
Thanks so much!

PR #416, opened 2017-07-30
Add option for calling event handlers sync or async

jonas747:


structs.go
```diff
@@ -50,6 +50,10 @@ type Session struct {
 	// active guilds and the members of the guilds.
 	StateEnabled bool
 
+	// Wether to call event handlers synchronously or not
```
iopred: nit: Whether or not to call event handlers synchronously.

jonas747:
and done

PR #424, opened 2017-08-13
Fix voice connections not decoding data properly

### LINKED ISSUE #423: Voice connections not receiving data
After some trial and error, I noticed that the header type is now 0x90 for some reason... This issue pretty much sums it up: https://github.com/hammerandchisel/discord-api-docs/issues/231

This is from the official desktop client too, so there's something going on here. For now, the simplest fix is to allow 0x90 alongside 0x80, and skip 8 bytes off p.Opus if the type is 0x90. If there's no better idea, I'll submit a PR if someone would like to confirm this issue.

tystuyfzand:
For some reason the type was updated to 0x90, likely due to a Chromium update (the issue was present in Discord with Chrome/browsers before, just never patched). This will ensure that we handle the data correctly.

This fixes #423.

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -796,7 +796,7 @@ func (v *VoiceConnection) opusReceiver(udpConn *net.UDPConn, close <-chan struct
 		}
 
 		// For now, skip anything except audio.
-		if rlen < 12 || recvbuf[0] != 0x80 {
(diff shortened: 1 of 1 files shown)


tystuyfzand:
Not sure why Travis failed this, looks like a linting issue.

iopred:
Please PR this into develop. Thank you!

Issue #379, opened 2017-05-07
Race condition


legolord208:
```
==================
WARNING: DATA RACE
Read at 0x00c42018e7d1 by goroutine 38:
  golang.org/x/crypto/salsa20/salsa.HSalsa20()
      /home/legolord208/Coding/Go/src/golang.org/x/crypto/salsa20/salsa/hsalsa20.go:28 +0x52a
  golang.org/x/crypto/nacl/secretbox.setup()
      /home/legolord208/Coding/Go/src/golang.org/x/crypto/nacl/secretbox/secretbox.go:34 +0xb5
  golang.org/x/crypto/nacl/secretbox.Open()
      /home/legolord208/Coding/Go/src/golang.org/x/crypto/nacl/secretbox/secretbox.go:111 +0xd4
  github.com/bwmarrin/discordgo.(*VoiceConnection).opusReceiver()
      /home/legolord208/Coding/Go/src/github.com/bwmarrin/discordgo/voice.go:814 +0x6ad

Previous write at 0x00c42018e7d1 by goroutine 39:
  reflect.Value.SetUint()
      /usr/local/go/src/reflect/value.go:1506 +0x141
  encoding/json.(*decodeState).literalStore()
      /usr/local/go/src/encoding/json/decode.go:994 +0x1e3b
  encoding/json.(*decodeState).literal()
      /usr/local/go/src/encoding/json/decode.go:799 +0x170
  encoding/json.(*decodeState).value()
      /usr/local/go/src/encoding/json/decode.go:405 +0x186
  encoding/json.(*decodeState).array()
      /usr/local/go/src/encoding/json/decode.go:555 +0x246
  encoding/json.(*decodeState).value()
      /usr/local/go/src/encoding/json/decode.go:399 +0xc1
  encoding/json.(*decodeState).object()
      /usr/local/go/src/encoding/json/decode.go:733 +0x163c
  encoding/json.(*decodeState).value()
      /usr/local/go/src/encoding/json/decode.go:402 +0x149
  encoding/json.(*decodeState).unmarshal()
      /usr/local/go/src/encoding/json/decode.go:184 +0x2e8
  encoding/json.Unmarshal()
      /usr/local/go/src/encoding/json/decode.go:104 +0x20a
  github.com/bwmarrin/discordgo.(*VoiceConnection).onEvent()
      /home/legolord208/Coding/Go/src/github.com/bwmarrin/discordgo/voice.go:435 +0x1075

[...]

PR #426, opened 2017-08-16
Fix concurrrency in voice connect/disconnect

Shywim:
Background: if a bot has to connect to a voice channel just after disconnecting from one, it "dies" in the channel (it can't play sound and `voice.waitUntilConnected` error with *"timeout waiting for voice"*).

I tested it with the airhorn example, you have to send *"!airhorn"* the moment the bot leaves and it will connect to the channel and stay in an error state indefinitely. You can make it change channel by typing *"!airhorn"* while being in another channel but it stays in this "errored" state. (the manipulation sounds far fetched, but it happens quite often naturally with the users on my server and I have to restart the bot manually)

Enabling logging, it seems to happens because waiting first for the **wsapi lock** let discordgo create the VoiceConnection with the new information, then subsequently delete it once the **session lock** is released from the connect side, meanwhile connecting with wsapi.

So inverting lock order seems to do the trick, without apparent performance loss or API problem (but I may have overlooked something). 😃

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -140,6 +140,14 @@ func (v *VoiceConnection) ChangeChannel(channelID string, mute, deaf bool) (err 
 // !!! NOTE !!! this function may be removed in favour of ChannelVoiceLeave
 func (v *VoiceConnection) Disconnect() (err error) {
 
+	// Close websocket and udp connections
+	v.Close()
+	v.log(LogInformational, "Deleting VoiceConnection %s", v.GuildID)
+
+	v.session.Lock()
+	delete(v.session.VoiceConnections, v.GuildID)
+	v.session.Unlock()
+
 	// Send a OP4 with a nil channel to disconnect
 	if v.sessionID != "" {
 		data := voiceChannelJoinOp{4, voiceChannelJoinData{&v.GuildID, nil, true, true}}
@@ -148,15 +156,6 @@ func (v *VoiceConnection) Disconnect() (err error) {
 		v.session.wsMutex.Unlock()
 		v.sessionID = ""
 	}
-
-	// Close websocket and udp connections
-	v.Close()
-
-	v.log(LogInformational, "Deleting VoiceConnection %s", v.GuildID)
-
-	v.session.Lock()
-	delete(v.session.VoiceConnections, v.GuildID)
-	v.session.Unlock()
 
 	return
 }


remd:
What's needed to get this merged @bwmarrin ? I can confirm tonight but I believe this issue is affecting my bot as well.

@Shywim Is there an issue filed for this?

Shywim:
There weren't when I posted this PR and I though it would be more straightforward to explain the issue with a PR with a solution that works (for my use case, at least).

remd:
I merged this to the current HEAD (c4a08d) and can still reproduce the issue by spamming my bot with chat messages and/or working my best to time a message right as the bot exits the channel.

```
$ pwd
/home/emd/.go/src/github.com/bwmarrin/discordgo
$ git branch -vv
* confirm-426 8195967 [Shywim/develop: ahead 10] Merge branch 'develop' of https://github.com/Shywim/discordgo into confirm-426
  develop     c4a08de [origin/develop] Remove a misspelled copy of EndpointGuildInvites. Issue #440 (#441)
  master      2fda7ce [origin/master] Bump to v0.17.0
$ git log | grep a0f58d4 -B1 -A 5
commit 8195967365bf19d8c08f44e216e7634422056eb0
Merge: c4a08de a0f58d4
Author: Eric Dostie <ericmdostie@gmail.com>
Date:   Wed Sep 20 19:18:00 2017 -0400

    Merge branch 'develop' of https://github.com/Shywim/discordgo into confirm-426

--

commit a0f58d447fbfd1778a94b3e5d4d48ecfc91b19da
Author: Matthieu Harlé <harle.matthieu@gmail.com>
Date:   Wed Aug 16 14:41:32 2017 +0200

    Fixed concurrrency in voice connect/disconnect

```

```
$ go run main.go
...
playing sound testsound.dca
playing sound testsound.dca
Failed to retrieve voice connection. Close and retry.
2017/09/20 19:28:42 timeout waiting for voice
2017/09/20 19:28:53 timeout waiting for voice
```

Shywim:
Weird it didn't happen in my tests and not "accidentally" anymore since I run my bots with this code. I guess we can close this and open an issue if the problem is deeper.

robbix1206:
I think this is related a bit too #379 most problems comes from the op4 packet without mutex or only partial ones

remd:
Nice find @robbix1206. I believe you are correct regarding the data race in #379 potentially causing this behavior.

robbix1206:
@remd Can you give me your testing script ? I will try to find out from where does this bug comes

remd:
I'm running https://github.com/mkolas/discord-soundboard. I can reproduce the bug on the `master` branch of that bot. I added some `log.Println(err)` lines in some spots to determine that the voice connection was timing out, but no other functional changes. 

There are instructions for setting up the bot in the README.md. Let me know if you need any more help or information @robbix1206.

lon9:
I've solved the problem. This isn't a problem of the library, it's problem of airhornbot.
The cause of it is calling `vc.Disconnect()` after deleting of queue.
https://github.com/discordapp/airhornbot/blob/master/cmd/bot/bot.go#L453

`vc.Disconnect()` takes about 1 second to execute, but the queue was already deleted. So it try to make new `VoiceConnection` but `vc.Disconnect()` is executing yet, and then cause that error `timeout waiting for voice`.

So, we have to put `delete(queues, play.GuildID)` after `vc.Disconnect()`, but it causes another problem that it try to push new sound to queue during `vc.Disconnect()`, and then queue will be deleted, so the sound will be never played.

https://gist.github.com/lon9/42e342893ffa82ebc7688a86a12c4f11

The gist is working example refer to https://github.com/jonas747/yagpdb/blob/master/soundboard/player.go

bwmarrin:
As best I can tell from reading this, the proposed PR here doesn't fix the issue.  Because of that I'm closing this.  There does seem to still be a potential issue here or race condition when joining and disconnecting from channels at the same time.     @lon9 you seem to have a good idea what's going on, if you want to open a issue for the problem we can try to see if there's something we should change in the library to help prevent it from happening.

Issue #429, opened 2017-08-18
WebhookDelete always returns an error


qaisjp:
https://discordapp.com/developers/docs/resources/webhook#delete-webhook

> Returns a 204 NO CONTENT response on success.

the method still tries to unmarshal on success, but since there is no content, it returns an unmarshal error

PR #432, opened 2017-08-19
#429-Fix

robbix1206:


qaisjp:
Uh, not sure if this is a proper fix. shouldn't it  unmarshal an error if `204 NO CONTENT` is not provided?

robbix1206:
@qaisjp 
I do exactly the same as other functions that have 204 in response.
If you feel that this is not the good solution please open an issue

Issue #380, opened 2017-05-07
Use uint64s instead of strings for IDs


ducc:

bwmarrin:
I think this change would be dependant on switching to ETF https://github.com/bwmarrin/discordgo/issues/96

NatoBoram:
For more context : https://discordapp.com/developers/docs/reference#snowflakes

> Discord utilizes Twitter's snowflake format for uniquely identifiable descriptors (IDs). These IDs are guaranteed to be unique across all of Discord, except in some unique scenarios in which child objects share their parent's ID. Because Snowflake IDs are up to 64 bits in size (e.g. a **uint64**), they are always returned as **strings** in the HTTP API to prevent integer overflows in some languages. See Gateway ETF/JSON for more information regarding Gateway encoding.

jonas747:
i opened a pr for this a while ago (#433), but the branch is a bit out of date now, I'm using a variation of it in production for my own bots.

robbix1206:
A problem for me is that we cannot check when we have a possible null value like in [this structure](https://discordapp.com/developers/docs/resources/channel#channel-object-channel-structure)
 in the parent id for example.
How to manage them ? Use 0 or negative value and check the validity of snowflakes values based on that criterion ?

jonas747:
That's what i've been doing, have had no problems so far, dont think there is any 0 snowflake, unless they sent a message at the exact time the discord epoch started, or something like that.

CarsonHoffman:
The library is quite reliant on strings at this point, and given that the API only deals with strings and there hasn't been much movement on this front, I'm going to go ahead and close this.

Issue #232, opened 2016-07-22
Review all uses of "int".


bwmarrin:
They should probably be int32 or int64.

dmportella:
go int64 and all should be oke this app doesnt need to be really memory efficient.

bwmarrin:
Yeah, I mean using just "int" makes it inconsistent across platforms.  So the plan is to change all instances of "int" to either int32 or int64 depending on which is actually needed.

tsdtsdtsd:
@bwmarrin 
> to either int32 or int64

Is there a reason not to use smaller int or unsigned int if applicable? 
I was already working on this one, but I was using the smaller sizes too. E.g.:

``` 
type Event struct {
	Operation uint8           `json:"op"`
```

I chose uint8 because the api documentation says it's a value between 0-11 (https://discordapp.com/developers/docs/topics/gateway#gateway-op-codespayloads)

dmportella:
no there is no reason other then as you dev things switching context between int8, int16, int32, int64, int 128 and the unsigned versions of it is really annoying. plus since we dont control discords public api they could one day say "right everyone all ints are bigints now" then we have to go change a ton of code.

tsdtsdtsd:
It's not that much code but well, I'm fine with int32/64

robbix1206:
Sorry bad reference

tsdtsdtsd:
> Yeah, I mean using just "int" makes it inconsistent across platforms. So the plan is to change all instances of "int" to either int32 or int64 depending on which is actually needed.

Sorry for this off-topic question, but I have to ask:
if "int" is inconsistent across platforms in a library (logic says you're right), does that mean that Go's stdlib is also inconsisten across platforms? I mean because stdlib funcs nearly always use "int", even in cases where a int8 would be enough.

jrbury:
Under the hood `int` is essentially `int32` on 32 bit machines, and `int64` on 64 bit machines. Go recommends that you just use `int` unless you have a very specific reason to use sized int type

CarsonHoffman:
This seems superfluous when Discord doesn't give specific sizes for types.

PR #433, opened 2017-08-24
Change all string ids to int64 ids

jonas747:
So this probably breaks 90% of the bots out there, and the gains are not major but i still feel like it needed to be done.
I expect this to not be merged in a while, if it's even merged at all because of the large breakage and small gains.

The main benefit i believe is less memory usage. A string id is 18 characters atm and go string headers are 2 integers, thats 34 bytes on a 64bit system, 26 bytes ovearhead per id.

There is some slightly more cpu used when decoding events however. 

I'm not done with this pr, i'm gonna use this branch in my fork and integrate my bot with it to do some real world testing and see what the real differences in performance, memory usage and usability actually are.

I will update this pr with the results of said tests.

### Notes:

 - All id's are represented as int64, using the `string` tag for decoding/encoding from/to json strings
 - This `string` tag did not work for slices :( so i had to make a new type, IDSlice.
 - For the api calls that has @me as a possibility as the id, i created a seperate function for.

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 	"time"
 )
 
@@ -143,4 +144,8 @@ func New(args ...interface{}) (s *Session, err error) {
 	// It is recommended that you now call Open() so that events will trigger.
 
 	return
+}
+
+func StrID(id int64) string {
+	return strconv.FormatInt(id, 10)
 }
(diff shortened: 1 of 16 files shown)


robbix1206:
Refer to #232

bwmarrin:
This PR is pretty out of date and the original issue https://github.com/bwmarrin/discordgo/issues/380 is on hold until we (if we) move to using ETF instead of JSON.  So I'm going to close this but keep the issue itself open.

PR #435, opened 2017-08-29
Added GameType

jD91mZM2:


structs.go
```diff
@@ -312,41 +312,28 @@ type Presence struct {
...
-	Name string `json:"name"`
-	Type int    `json:"type"`
-	URL  string `json:"url,omitempty"`
+	Name string   `json:"name"`
+	Type GameType `json:"type"`
+	URL  string   `json:"url,omitempty"`
 }
 
 // UnmarshalJSON unmarshals json to Game struct
 func (g *Game) UnmarshalJSON(bytes []byte) error {
```
iopred: This method shouldn't be needed anymore.

legolord208: Very true. Will delete it!

Issue #440, opened 2017-09-07
[Typo] var "EndpointGuildInvites" in discordgo/endpoints.go


jcmdln:
It appears that EndpointGuildInvites is declared twice, though there is a typo in the first declaration on line 74: 
```
...
	EndpointGuildInivtes         = func(gID string) string { return EndpointGuilds + gID + "/invites" }
```

Later on line 86:
```
...
	EndpointGuildInvites         = func(gID string) string { return EndpointGuilds + gID + "/invites" }
...
```

It appears that the only other mention of this variable is on line 910 of discordgo/restapi.go, but if I'm not mistaken this is in fact a typo and should be corrected?
```
	body, err := s.RequestWithBucketID("GET", EndpointGuildInvites(guildID), nil, EndpointGuildInivtes(guildID))
```

iopred:
Definitely should be fixed.

jcmdln:
It appears that this was fixed in a PR. I will close this issue for the time being.

PR #441, opened 2017-09-07
Remove a misspelled copy of EndpointGuildInvites. Issue #440

therealfakemoot:
Reference Issue: https://github.com/bwmarrin/discordgo/issues/440
This endpoint was declard twice in endpoints.go, and used in the
GuildInvites() method on Session values. I replaced the misspelled
invocation.

All tests pass.


bwmarrin:
Thanks @therealfakemoot  :)

PR #444, opened 2017-09-20
Fix #443 (recipients not recipient)

### LINKED ISSUE #443: Channel struct's JSON tag for Recipients is incorrect
The _Channel_ struct has a JSON tag for the field _Recipients_ that currently says "recipient", while the present API uses a field named "recipient**s**", causing it to go unfilled by the unmarshaller.  
See https://discordapp.com/developers/docs/resources/channel

robbix1206:
Fix #443


bwmarrin:
Thanks @robbix1206 :)

PR #454, opened 2017-10-10
Add members from GuildMembersChunk to state

MaxBosse:
Adds members from the GuildMembersChunk event to state.

Results in Members being in both, guild.Members + the private members-map compared to just the guild.Members slice when doing it "manually" in your application which can cause issues.


iopred:
Thank you for this PR.

I merged it into master by accident, sorry!

Could you PR this into the develop branch, sorry to bother you!

PR #458, opened 2017-10-16
Rich Presence

SilverCory:
Not perfect, from what I can see there's still fields missing
- partySize
- partyMax
- partyId
- matchSecret
- joinSecret
- spectateSecret
- instance

Whilst I could stab in the dark guessing where these go I've not the competence to check.

I'm going to be checking these, but no guarantee.

Feedback appreciated, I'll fix docs ect when we establish whether there's a better way about this rather than UpdateStatusRaw

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -21,7 +21,7 @@ import (
 )
 
 // VERSION of DiscordGo, follows Semantic Versioning. (http://semver.org/)
-const VERSION = "0.17.0-dev"
+const VERSION = "0.17.0"
 
 // ErrMFA will be risen by New when the user has 2FA.
 var ErrMFA = errors.New("account has 2FA enabled")
--- a/state.go
+++ b/state.go
@@ -816,6 +816,13 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 		if s.TrackMembers {
 			err = s.MemberRemove(t.Member)
 		}
+	case *GuildMembersChunk:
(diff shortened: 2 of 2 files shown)


SilverCory:
Those test structs don't work I'll revert those soon.

iopred:
I would like to see the status updating from https://github.com/bwmarrin/discordgo/pull/459 added to this PR.

I would prefer we create a `UpdateStatusComplex(data UpdateStatusData)` so that we future proof the API for next time too :)

SilverCory:
I agree with that, do you want me to merge #459 to this then?

iopred:
So I'd like to change UpdateStatusRaw in this PR to:

`UpdateStatusComplex(data UpdateStatusData)`

Exporting updateStatusData seems like it's more future proof.

SilverCory:
Yeah I can see that being the case. I'll do this within an 2 hours if that's okay

iopred:
No rush, please take your time!

iopred:
Could you make this change into the `develop` branch too? We don't PR into master.

SilverCory:
Would you like UpdateStatus and UpdatingStremingStatus to fall back to UpdateStatusComplex, aswell then? and yeah sure.

iopred:
Yeah they should just call into UpdateStatusComplex.

SilverCory:
#462

PR #459, opened 2017-10-19
Added ability to change the online status

Ikkerens:
As far as I could tell, there was no way of changing the users status through the API so I added a few methods to accomodate users wishing to do this.


iopred:
We've had a lot of duplicated API's over time for this. With status gaining rich data I think we need to combine this PR with https://github.com/bwmarrin/discordgo/pull/458/files

I would prefer we create a `UpdateStatusComplex(data UpdateStatusData)` which includes the rich data also.

iopred:
Sorry to close this, but I think we will go with the other PR on this one, but thank you very much for bringing the status to our attention, it'll get added in https://github.com/bwmarrin/discordgo/pull/458/files

PR #460, opened 2017-10-20
added clarification when initializing discordgo

spencer1573:
from the code I ran on my system with the latest changes this seems to be the syntax for the authentication tokens. I'm guessing it was just never updated.
Let me know if this is incorrect. 
Thanks!


bwmarrin:
Hey that's a good idea :) However if you're not using a Bot account, you would omit the "Bot " section.  Maybe you could also add a tiny mention that this example is for connecting as a bot user?

PR #462, opened 2017-10-20
Implement Raw sending of status (Rich Presence and Online Status)

SilverCory:
@Ikkerens had his cloned and Rich Presence was implemented on top of this.

- #458 is the original.
- #459 was added as per request.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -324,7 +324,7 @@ const (
 // A Game struct holds the name of the "playing .." game for a user
 type Game struct {
 	Name          string     `json:"name"`
-	Type          int        `json:"type"`
+	Type          GameType   `json:"type"`
 	URL           string     `json:"url,omitempty"`
 	Details       string     `json:"details,omitempty"`
 	State         string     `json:"state,omitempty"`
--- a/structs.go
+++ b/structs.go
@@ -323,9 +323,30 @@ const (
 
 // A Game struct holds the name of the "playing .." game for a user
 type Game struct {
-	Name string   `json:"name"`
-	Type GameType `json:"type"`
-	URL  string   `json:"url,omitempty"`
+	Name string `json:"name"`
+	Type int    `json:"type"`
+	URL  string `json:"url,omitempty"`
+	Details       string     `json:"details,omitempty"`
+	State         string     `json:"state,omitempty"`
+	TimeStamps    TimeStamps `json:"timestamps,omitempty"`
+	Assets        Assets     `json:"assets,omitempty"`
+	ApplicationID string     `json:"application_id,omitempty"`
+	Instance      int8       `json:"instance,omitempty"`
+	// TODO: Party and Secrets (unknown structure)
+}
+
+// A TimeStamps struct contains start and end times used in the rich presence "playing .." Game
+type TimeStamps struct {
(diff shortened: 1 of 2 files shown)


wsapi.go
```diff
@@ -267,6 +267,30 @@ type updateStatusOp struct {
...
+// If game!="" then set game.
+// if otherwise, set status to active, and no game.
+func (s *Session) UpdateStatus(idle int, game string) (err error) {
+	return s.UpdateStreamingStatus(idle, game, "")
+}
+
+// UpdateOnlineStatus is used to update the user's online status
+// If idle>0 and status=="online" then set status to idle
+// If otherwise, set status to the given value and no game
+func (s *Session) UpdateOnlineStatus(status string, afk bool, idle int) (err error) {
```
iopred: I don't think we need to bother with this now.

wsapi.go
```diff
@@ -267,6 +267,30 @@ type updateStatusOp struct {
...
+	return s.UpdateOnlineStreamingStatus(status, afk, idle, "", "")
+}
+
+// UpdateOnlineStreamingStatus is used to update the user's online status and game/streaming status
+// If idle>0 and status=="online" then set status to idle
+// If game!="" then set game
+// If game!="" and url!="" then set the status type to streaming with the URL set.
+// If otherwise, set status to the given value and no game
+func (s *Session) UpdateOnlineStreamingStatus(status string, afk bool, idle int, game string, url string) (err error) {
```
iopred: I don't think we need this either, or at least make this private.

structs.go
```diff
@@ -323,9 +323,30 @@ const (
 
 // A Game struct holds the name of the "playing .." game for a user
 type Game struct {
-	Name string   `json:"name"`
-	Type GameType `json:"type"`
-	URL  string   `json:"url,omitempty"`
+	Name          string     `json:"name"`
+	Type          int        `json:"type"`
```
iopred: Change Type back to GameType, changing to int is an api change.

wsapi.go
```diff
@@ -276,7 +276,7 @@ func (s *Session) UpdateStreamingStatus(idle int, game string, url string) (err
 		return ErrWSNotFound
```
iopred: Move this block into UpdateStatusComplex, it should be thread safe too! :)

wsapi.go
```diff
@@ -291,11 +291,16 @@ func (s *Session) UpdateStreamingStatus(idle int, game string, url string) (err
 		}
 		usd.Game = &Game{
 			Name: game,
-			Type: gameType,
+			Type: int(gameType),
```
iopred: This shouldn't be needed.

SilverCory:
https://travis-ci.org/bwmarrin/discordgo/jobs/290566359#L483

iopred:
Oops, my apologies, looks like we need the cast when creating the Game, but we should keep the API as using GameType.

wsapi.go
iopred: Sorry, I think you missed this, we should move the RLock/wsConn check to UpdateStatusComplex.

SilverCory: Yeah sorry I missed that.

iopred:
Thanks so much for this <3

PR #464, opened 2017-10-23
Use v.log rather than log for consistency.

NamedKitten:
Change made to keep the use of logging consistent.


bwmarrin:
Thanks @NamedKitten

PR #465, opened 2017-10-23
Add support for custom logging behaviour.

NamedKitten:
I would like this PR to be merged because I can use it to replace the log what discordgo does with something like logrus.
Feel free to add comments for adjustments that might need to be made.


iopred:
Please PR into develop.

NamedKitten:
Fixed.

PR #467, opened 2017-10-29
Allow size parameter to be omitted in User.AvatarURL

Necroforger:
Allows you to leave the size parameter as an empty string


iopred:
Please patch this into develop

Necroforger:
oops

PR #470, opened 2017-10-30
Change uint to float64 for TimeStamps

MaxBosse:
Hey,

currently the develop branch will throw the following error when parsing a PRESENCE_UPDATE containing a Game object:
`2017/10/30 14:52:48 [DG0] wsapi.go:477:onEvent() error unmarshalling PRESENCE_UPDATE event, json: cannot unmarshal number 1509371565406.0 into Go struct field TimeStamps.start of type uint`

This change will fix the error by changing the `TimeStamps.start` and `TimeStamps.end` values from an uint to a float64.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -337,8 +337,8 @@ type Game struct {
 
 // A TimeStamps struct contains start and end times used in the rich presence "playing .." Game
 type TimeStamps struct {
-	EndTimestamp   uint `json:"end,omitempty"`
-	StartTimestamp uint `json:"start,omitempty"`
+	EndTimestamp   float64 `json:"end,omitempty"`
+	StartTimestamp float64 `json:"start,omitempty"`
 }
 
 // An Assets struct contains assets and labels used in the rich presence "playing .." Game


njhanley:
Why use a float64 when the payload is an int64?
https://discordapp.com/developers/docs/topics/rich-presence#updating-presence-update-presence-payload-fields

iopred:
Breaking the API is not the right way to fix this, writing a custom UnmarshalJSON seems right here.

njhanley:
@iopred As I see it, the API itself is broken. Discord returns startTimestamp and endTimestamp as signed 64-bit integers, but TimeStamps.start and TimeStamps.end are 32-bit integers on some platforms, thus causing the unmarshalling error.

njhanley:
My previous comment was incorrect in assuming the Discord Rich Presence SDK payload was reflective of the Discord API. I see now that Discord returns the "timestamps" object's fields as integers encoded as floats.

I agree that a custom UnmarshalJSON makes sense but the API still must be changed as a uint is 32-bit on some platforms. I've created a pull request (#475) with the changes I believe are necessary.

PR #471, opened 2017-11-03
Add custom dialer option

cking:
This change allows the dialer to be replaced with a custom version.
You can replace the general websocket dialer and if you want, you can use a different dialer for voice connections.

As an example, you could enable SOCKS5 proxying for voice requests with the following code and `x/net/proxy`:
```go
	socksDialer, err := proxy.SOCKS5("tcp", "localhost:31337", nil, &net.Dialer{})
	if err != nil {
		fmt.Println("Error creating socks proxy", err)
		return
	}

	session.VoiceDialer = &websocket.Dialer{
		NetDial: socksDialer.Dial,
	}
```

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -301,7 +301,15 @@ func (v *VoiceConnection) open() (err error) {
 	// Connect to VoiceConnection Websocket
 	vg := fmt.Sprintf("wss://%s", strings.TrimSuffix(v.endpoint, ":80"))
 	v.log(LogInformational, "connecting to voice endpoint %s", vg)
-	v.wsConn, _, err = websocket.DefaultDialer.Dial(vg, nil)
+	dialer := v.session.VoiceDialer
+	if dialer == nil {
+		if v.session.Dialer != nil {
+			dialer = v.session.Dialer
+		} else {
+			dialer = websocket.DefaultDialer
+		}
+	}
+	v.wsConn, _, err = dialer.Dial(vg, nil)
 	if err != nil {
 		v.log(LogWarning, "error connecting to voice endpoint %s, %s", vg, err)
 		v.log(LogDebug, "voice struct: %#v\n", v)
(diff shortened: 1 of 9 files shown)


cking:
By the way, the test `TestOpenClose` works locally for me.

iopred:
Please update the PR to the `develop` branch.

voice.go
```diff
@@ -302,7 +302,15 @@ func (v *VoiceConnection) open() (err error) {
 	// Connect to VoiceConnection Websocket
 	vg := fmt.Sprintf("wss://%s", strings.TrimSuffix(v.endpoint, ":80"))
 	v.log(LogInformational, "connecting to voice endpoint %s", vg)
-	v.wsConn, _, err = websocket.DefaultDialer.Dial(vg, nil)
+	dialer := v.session.VoiceDialer
+	if dialer == nil {
+		if v.session.Dialer != nil {
+			dialer = v.session.Dialer
```
iopred: Is there a good reason to default back to session.Dialer as well? It's not really documented above and seems like an easy thing to set in the application:

```
someDialer := Dialer{}
session.Dialer = someDialer
session.VoiceDialer = someDialer
```

cking: I left it in because it made sense in my eyes.
Specifying nothing will fallback to the default dialer. So wouldn't it make more sense to fall back to the global dialer for voice comms if one is specified.
Makes more sense to me this way

iopred:
Then it should be documented on the struct, I think the documentation on the struct needs to be a little stronger anyways.

cking:
i did rebase it now (aka its running from develop) - patch works for master and develop

bwmarrin:
This PR has to changed to develop to be accepted, and it's been awhile..  So I'm going to close it for now.  If you want to update it to merge against develop we can look at it again.

PR #475, opened 2017-11-09
Fix TimeStamps unmarshalling (#474)

njhanley:
TimeStamps used in rich presence currently cause the following error:
`[DG0] wsapi.go:477:onEvent() error unmarshalling PRESENCE_UPDATE event, json: cannot unmarshal number 1510183304574.0 into Go struct field TimeStamps.start of type uint`

This change fixes the issue by implementing UnmarshalJSON on TimeStamps and changing startTimestamp and endTimestamp from uint (which may be too small on some platforms) to int64.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -12,6 +12,7 @@
 package discordgo
 
 import (
+	"encoding/json"
 	"net/http"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


iopred:
I hate breaking the API, but this hasn't made it's way to master yet. If it had I would prefer not breaking users of the library and just deal with uint.

robbix1206:
Fix #474

PR #478, opened 2017-11-16
Added JetBrains GoLang Editor

ianwijma:
Added the GoLand editor


bwmarrin:
Hey not a bad idea to add GoLand on there.  However all of those files are generated by mkdocs from the markdown files in the docs folder of the main repo.  So any changes made to these files would be overwrote when the docs were regenerated.

You could pose this PR to the docs markdown file

PR #480, opened 2017-11-17
Implement new status types

PixeLInc:
Implement the new Listening & Watching status types.
Also, moved UpdateStreamStatus to a new method to accept the new GameType parameter and made the old method use the new method.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -320,6 +320,8 @@ type GameType int
 const (
 	GameTypeGame GameType = iota
 	GameTypeStreaming
+	GameTypeListening
+	GameTypeWatching
 )
 
 // A Game struct holds the name of the "playing .." game for a user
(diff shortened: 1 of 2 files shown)


iopred:
Don't you just want UpdateStatusComplex?

PixeLInc:
Well, sure you could use that but I felt like it would be easier to just use the defined method for it just like the UpdateStatus/UpdateStreamingStatus method where it explicitly sets it so you don't have to know the type number or anything of that sort, no?

iopred:
The problem with adding new methods is that the discord API is always changing, the thing we try to do the least is break current users of the API. The complex method's tend to be a little bit more resilient to API changes in the future, which is why we tend to prefer them.

PixeLInc:
Okay. Well then I'll just close up this PR.

PR #484, opened 2017-11-20
Fix MessageEmbed Timestamp

vkhashimoto:
Fix Timestamp type from string to Time, since with string Discord would return a Bad Requesst.

Haven't tested other structs with Timestamp.


qaisjp:
Maybe they are submitting a malformed timestamp

PR #488, opened 2017-12-07
adds default avatar generation to AvatarURL method

Seklfreak:
AvatarURL will return a link to the default avatars instead of an invalid link if the user has no avatar set.

As described here: https://discordapp.com/developers/docs/reference#image-formatting-cdn-endpoints

### DIFF
--- a/endpoints.go
+++ b/endpoints.go
@@ -63,8 +63,9 @@ var (
 	EndpointUser               = func(uID string) string { return EndpointUsers + uID }
 	EndpointUserAvatar         = func(uID, aID string) string { return EndpointCDNAvatars + uID + "/" + aID + ".png" }
 	EndpointUserAvatarAnimated = func(uID, aID string) string { return EndpointCDNAvatars + uID + "/" + aID + ".gif" }
-	EndpointDefaultUserAvatar  = func(uDiscriminator int) string {
-		return EndpointCDN + "embed/avatars/" + strconv.Itoa(uDiscriminator%5) + ".png"
+	EndpointDefaultUserAvatar  = func(uDiscriminator string) string {
+		uDiscriminatorInt, _ := strconv.Atoi(uDiscriminator)
+		return EndpointCDN + "embed/avatars/" + strconv.Itoa(uDiscriminatorInt%5) + ".png"
 	}
 	EndpointUserSettings      = func(uID string) string { return EndpointUsers + uID + "/settings" }
 	EndpointUserGuilds        = func(uID string) string { return EndpointUsers + uID + "/guilds" }
--- a/endpoints.go
+++ b/endpoints.go
@@ -11,6 +11,8 @@
 
 package discordgo
 
+import "strconv"
+
 // APIVersion is the Discord API version used for the REST and Websocket API.
 var APIVersion = "6"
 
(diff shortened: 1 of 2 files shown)


user.go
```diff
@@ -34,7 +35,10 @@ func (u *User) Mention() string {
 //             be added to the URL.
 func (u *User) AvatarURL(size string) string {
 	var URL string
-	if strings.HasPrefix(u.Avatar, "a_") {
+	if u.Avatar == "" {
+		discriminatorInt, _ := strconv.Atoi(u.Discriminator)
```
iopred: Instead of throwing away this error, just use Discriminator as a string in endpoints.go

Seklfreak: Wasn't sure how much logic I should put into endpoints.go, I will change it 👍

iopred: Endpoints shouldn't need any real logic, if you just use strings (it's kinda gross, but we don't use discriminators as ints anywhere else in the codebase)

endpoints.go
```diff
@@ -63,8 +63,9 @@ var (
 	EndpointUser               = func(uID string) string { return EndpointUsers + uID }
 	EndpointUserAvatar         = func(uID, aID string) string { return EndpointCDNAvatars + uID + "/" + aID + ".png" }
 	EndpointUserAvatarAnimated = func(uID, aID string) string { return EndpointCDNAvatars + uID + "/" + aID + ".gif" }
-	EndpointDefaultUserAvatar  = func(uDiscriminator int) string {
-		return EndpointCDN + "embed/avatars/" + strconv.Itoa(uDiscriminator%5) + ".png"
+	EndpointDefaultUserAvatar  = func(uDiscriminator string) string {
+		uDiscriminatorInt, _ := strconv.Atoi(uDiscriminator)
+		return EndpointCDN + "embed/avatars/" + strconv.Itoa(uDiscriminatorInt%5) + ".png"
```
iopred: Oh you need the int for modulus, guh.

I regret asking you for the previous change now, sorry!

But now that we need to do the Atoi, can you check the error in user.go and return "" if it fails?

Seklfreak: if it fails it returns 0, the final url will be https://cdn.discordapp.com/embed/avatars/0.png, which is a valid url. I think that might be better to return instead of just ""?

endpoints.go
```diff
@@ -61,14 +63,18 @@ var (
...
-	EndpointUserGuilds         = func(uID string) string { return EndpointUsers + uID + "/guilds" }
-	EndpointUserGuild          = func(uID, gID string) string { return EndpointUsers + uID + "/guilds/" + gID }
-	EndpointUserGuildSettings  = func(uID, gID string) string { return EndpointUsers + uID + "/guilds/" + gID + "/settings" }
-	EndpointUserChannels       = func(uID string) string { return EndpointUsers + uID + "/channels" }
-	EndpointUserDevices        = func(uID string) string { return EndpointUsers + uID + "/devices" }
-	EndpointUserConnections    = func(uID string) string { return EndpointUsers + uID + "/connections" }
-	EndpointUserNotes          = func(uID string) string { return EndpointUsers + "@me/notes/" + uID }
+	EndpointDefaultUserAvatar  = func(uDiscriminator string) string {
+		uDiscriminatorInt, _ := strconv.Atoi(uDiscriminator)
+		return EndpointCDN + "embed/avatars/" + strconv.Itoa(uDiscriminatorInt%5) + ".png"
```
iopred: Are you missing the uID here somewhere? embed/avatars/1682.png doesn't seem right for my avatar?

Seklfreak: It's the default avatar url, which is used if you don't have a custom avatar set. And it's discriminator modulo 5, not just the discriminator in the url. 1682 modulo 5 = 2: Your default avatar is https://cdn.discordapp.com/embed/avatars/2.png .

Seklfreak: There are in total five default avatars:
https://cdn.discordapp.com/embed/avatars/0.png
https://cdn.discordapp.com/embed/avatars/1.png
https://cdn.discordapp.com/embed/avatars/2.png
https://cdn.discordapp.com/embed/avatars/3.png
https://cdn.discordapp.com/embed/avatars/4.png

Necroforger: I think the endpoint should be named EndpointUserAvatarDefault so it sits next to the rest of the avatar endpoints.

PR #492, opened 2017-12-13
Modified ChannelEdit to accept a ChannelParams object

psheets:
Modified ChannelEdit to accept a ChannelParams object.

This allows for management such as topics to be updated using ChannelEdit.


iopred:
The way this has usually been done is to add a new ChannelEditComplex method, that takes a channel object.
We should aim to not break the API for existing users.
You're changing the encoding of all the files so please make sure that isn't done.
Also please make PR's into the develop branch.

PR #493, opened 2017-12-14
Add ChannelEditComplex

psheets:
Add ChannelEditComplex function and ChannelEdit struct.

Allows updating a channel's fields such as topic.

Implemented as suggested by @iopred in [#322](https://github.com/bwmarrin/discordgo/pull/322) and [#492](https://github.com/bwmarrin/discordgo/pull/492)


Seklfreak:
I think all possible attributes should be added. ( https://discordapp.com/developers/docs/resources/channel#modify-channel )

Missing right now: `position`, `user_limit`, `permission_overwrites` and `parent_id`.

psheets:
ChannelEdit struct has been updated to include all fields per @Seklfreak comment.

iopred:
2 things:
1. The file encodings changed again
2. Modify the existing ChannelEdit method to call the new ChannelEditComplex method

Thank you so much for this PR!

Strum355:
Re. point 1, I havent heard them being called "file encodings", so for those who havent heard this, its the files permissions (as per *nix) that have been changed from 644 to 755.

iopred:
Ah, sorry I didn't look closely, I've also seen this when the newline character (\n -> \r\n) changes, which is why I said encoding.

Strum355:
I ran the files and its current live one, no difference in encoding, so it must be just a permissions thing ^^

psheets:
Reverted permissions back to 644 and Modified ChannelEdit to use ChannelEditComplex per @iopred and @Strum355

restapi.go
```diff
@@ -1221,17 +1221,19 @@ func (s *Session) Channel(channelID string) (st *Channel, err error) {
...
+func (s *Session) ChannelEdit(channelID, name string) (*Channel, error) {
+	return s.ChannelEditComplex(&ChannelEdit{
+		Name: name,
+		ID:   channelID,
+	})
+}
 
-	body, err := s.RequestWithBucketID("PATCH", EndpointChannel(channelID), data, EndpointChannel(channelID))
+// ChannelEditComplex edits an existing channel, replacing the parameters entirely with ChannelEdit struct
+func (s *Session) ChannelEditComplex(c *ChannelEdit) (st *Channel, err error) {
```
iopred: Sorry to be a pain, could we change this method signature to:

ChannelEditComplex(channel ID string, c *ChannelEdit)

We can then remove the ID field from the struct.

This would bring this API closer to the ChannelMessageSendComplex method.

Thanks!

psheets: Not a pain, thank you for the feedback. This is a learning experience for me.

The current signature is mimicking ChannelMessageEditComplex:
`func (s *Session) ChannelMessageEditComplex(m *MessageEdit) (st *Message, err error)`

and the MessageEdit struct:
```
type MessageEdit struct {
	Content *string       `json:"content,omitempty"`
	Embed   *MessageEmbed `json:"embed,omitempty"`

	ID      string
	Channel string
}
```

Should ChannelMessageEditComplex be rewritten to be inline with your suggestion?
Adding channelID seems to match the other functions.

psheets: I see that changing ChannelMessageEditComplex will break the existing API.

iopred:
Thanks so much!

PR #496, opened 2017-12-22
Adds support for animated emojis

Seklfreak:
This adds the newly added field to the emoji struct identifying animated emojis.


iopred:
Thanks!

PR #506, opened 2018-01-26
travis: update go versions

dvrkps:


bwmarrin:
I'm guessing, but does this change it to test vs the latest dot release of each of those version?

dvrkps:
@bwmarrin .x match latest patch subversion of each 1.7, 1.8 and 1.9.

bwmarrin:
Sounds like a good idea :)

PR #508, opened 2018-02-07
Allow setting of Game to nil, and allow the game type to be "listening"

dannyt66:
This code serves to allow the library to set the user's Game to nil, or not currently playing a game, and also allows for the use of Discord's new "Listening to" function to be used in place of "Playing" and "Streaming".

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -21,7 +21,7 @@ import (
 )
 
 // VERSION of DiscordGo, follows Semantic Versioning. (http://semver.org/)
-const VERSION = "0.17.0"
+const VERSION = "0.18.0"
 
 // ErrMFA will be risen by New when the user has 2FA.
 var ErrMFA = errors.New("account has 2FA enabled")
@@ -50,7 +50,7 @@ func New(args ...interface{}) (s *Session, err error) {
 	// Create an empty Session interface.
 	s = &Session{
 		State:                  NewState(),
-		ratelimiter:            NewRatelimiter(),
+		Ratelimiter:            NewRatelimiter(),
 		StateEnabled:           true,
 		Compress:               true,
 		ShouldReconnectOnError: true,
--- a/discord.go
+++ b/discord.go
@@ -21,7 +21,7 @@ import (
 )
 
 // VERSION of DiscordGo, follows Semantic Versioning. (http://semver.org/)
-const VERSION = "0.18.0"
+const VERSION = "0.18.0-alpha"
 
 // ErrMFA will be risen by New when the user has 2FA.
 var ErrMFA = errors.New("account has 2FA enabled")
--- a/discord.go
+++ b/discord.go
@@ -21,7 +21,7 @@ import (
 )
 
 // VERSION of DiscordGo, follows Semantic Versioning. (http://semver.org/)
-const VERSION = "0.17.0-dev"
+const VERSION = "0.17.0"
 
 // ErrMFA will be risen by New when the user has 2FA.
 var ErrMFA = errors.New("account has 2FA enabled")
--- a/wsapi.go
+++ b/wsapi.go
@@ -15,6 +15,7 @@ import (
 	"compress/zlib"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"runtime"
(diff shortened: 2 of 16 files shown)


dannyt66:
This code is breaking the build as it breaks compatibility with the old UpdateStatus method - I'd propose fixing this by having two instances of the method, one which uses the old system, however I do not know if this is acceptable to the way this library maintains compatibility. Please advise on the best fix for this.

wsapi.go
```diff
@@ -328,7 +328,7 @@ type updateStatusOp struct {
 // If game!="" then set game.
 // If game!="" and url!="" then set the status type to streaming with the URL set.
 // if otherwise, set status to active, and no game.
-func (s *Session) UpdateStreamingStatus(idle int, game string, url string) (err error) {
+func (s *Session) UpdateStreamingStatus(idle int, game string, url string, isSong bool) (err error) {
```
iopred: This will break existing users of the API, I think just the change to line 337 is enough to allow this.

If you want a simple way to update the status, I suggest just adding an 'UpdateListeningStatus' method, but I would shy away from that personally.

dannyt66: Branch changed, apoligies. I think the ability to send the listening status should be added, however I'm not sure how you'd best want me to add this if not by adding an 'UpdateListeningStatus' method.

iopred: UpdateStatusComplex would be the way to call this, adding some additional comments on the method could explain this to users.

dannyt66:
New commits pushed to maintain compatibility, whilst still allowing the update of listening status.

discord.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 // VERSION of DiscordGo, follows Semantic Versioning. (http://semver.org/)
-const VERSION = "0.18.0-alpha"
+const VERSION = "0.18.0"
```
iopred: please undo this change.

wsapi.go
```diff
@@ -351,6 +351,9 @@ func (s *Session) UpdateStreamingStatus(idle int, game string, url string) (err
 			URL:  url,
 		}
 	}
+	if game == "" {
+		usd.Game = nil
```
iopred: This shouldn't be needed, the default is nil.

wsapi.go
```diff
@@ -379,6 +382,22 @@ func (s *Session) UpdateStatus(idle int, game string) (err error) {
...
+//If game!="" then set to what user is listening to
+//Else, set user to active and no game.
+func (s *Session) UpdateListeningStatus(idle int, game string) (err error) {
+	gameType := GameTypeListening
+	usd := UpdateStatusData{
+		Status: "online",
+	}
+	usd.Game = &Game{
+		Name: game,
+		Type: gameType,
```
iopred: This probably reads a bit better as:


return s.UpdateStatusComplex(UpdateStatusData{
  Status: "online",
  Game: &Game {
    Name: game,
    Type: GameTypeListening,
    URL: "",
  }
})

wsapi.go
dannyt66: This is definitely not the case, if I've sent "" without this change, the game has been left as it was previously set to.

discord.go
dannyt66: Sorry, this change was inadvertent. Will fix.

wsapi.go
iopred: Hmmmmm, I think I see what's going on, this check should be in UpdateListeningStatus then, sorry for the back and forth on this one!

wsapi.go
```diff
@@ -379,6 +382,20 @@ func (s *Session) UpdateStatus(idle int, game string) (err error) {
 	return s.UpdateStreamingStatus(idle, game, "")
 }
 
+//UpdateListeningStatus is used to set the user to "Listening to..."
+//If game!="" then set to what user is listening to
+//Else, set user to active and no game.
+func (s *Session) UpdateListeningStatus(idle int, game string) (err error) {
+	return s.UpdateStatusComplex(UpdateStatusData{
```
iopred: Sorry, I was misunderstanding what you're trying here, this should be:

var g Game
if game != "" {
  g = &Game{
    Name: game,
    Type: GameTypeListening,
    URL:  "",
  }
}
return s.UpdateStatusComplex(UpdateStatusData{
  Status: "online",
  Game: g,
})

wsapi.go
dannyt66: This behaviour exhibits itself when setting an actual game too, which is why I put the check in there. I do agree that I probably need a similar check in UpdateListeningStatus as well, however.

wsapi.go
```diff
@@ -379,6 +379,23 @@ func (s *Session) UpdateStatus(idle int, game string) (err error) {
 	return s.UpdateStreamingStatus(idle, game, "")
 }
 
+//UpdateListeningStatus is used to set the user to "Listening to..."
+//If game!="" then set to what user is listening to
+//Else, set user to active and no game.
+func (s *Session) UpdateListeningStatus(idle int, game string) (err error) {
```
iopred: Let's make this:

UpdateListeningStatus(game)

Idle seems kinda useless and people can do it with Complex.

wsapi.go
```diff
@@ -379,6 +379,23 @@ func (s *Session) UpdateStatus(idle int, game string) (err error) {
 	return s.UpdateStreamingStatus(idle, game, "")
 }
 
+//UpdateListeningStatus is used to set the user to "Listening to..."
```
iopred: Could you put a space between // and the comments on these lines.

Issue #386, opened 2017-05-26
Add Audit Log reason support


nullpixel:
Via `X-Audit-Log-Reason` header

bwmarrin:
Maybe you can add a bit more context here on what you want, how it improves the library, and maybe a bit more detail on how you're envisioning it being implemented?

nullpixel:
Well, just add a reason parameter to every auditable action. It allows the end user to specify reasons and state who does things, which eliminates the need for a modlog.

Seklfreak:
The implementation is described here: https://discordapp.com/developers/docs/resources/audit-log

And here an example how it looks in action: 
<img width="924" alt="screen shot 2017-07-26 at 04 08 09" src="https://user-images.githubusercontent.com/1156226/28601394-4519e760-71b8-11e7-9358-1a5eaf09ea04.png">

It's useful for Bots so they can attach who actually caused the, for example, kick. Without Audit Log Reason users can call bot commands and then remove their messages making it impossible to trace it back (without other mod logs as nullpixel mentioned).

Seklfreak:
Ban and Kick already have a function for adding a Reason. Is there a list with all endpoints that support audit log reasons?

bwmarrin:
That's a pretty good question and I'm not sure I know the answer even at this point.  I'll try to research it a bit though and see - or anyone else that already knows off the top of their head are free to jump in :)

JurrijnP:
Here 356 days later, I tested this and this is what I have:
When adding **X-Audit-Log-Reason** in the request header and giving someone a role through a bot it also adds the reason in the audit log, it seems it also has no inpact on a endpoint like getting channel data.

I don't know though what would be a good way to implement this, I think we should avoid adding a ton of _WithReason_ functions unless there is no other way.

J0eppp:
If I'm correct #785 implemented this, so maybe this issue can be closed?

Cyb3r-Jak3:
I don't believe this has been fully addressed. While functions such as Kick and Ban have a separate WithReason, functions such ChannelMessagePin and GuildMemeberRoleAdd do not. Not sure if I should make it a separate issue.

bwmarrin:
@Cyb3r-Jak3 If you can, lets just make a ticket specific to what you've found.

PR #512, opened 2018-02-27
Add Audit Log Reason Support (#386)

rei2hu:
This pull request adds audit log reason support.

The new `OptionalRequestHeaders` struct can be extended to cover future headers that may want to be needed in the future.

Also added versions of functions so that they take a reason argument.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -47,6 +47,11 @@ func (s *Session) Request(method, urlStr string, data interface{}) (response []b
 
 // RequestWithBucketID makes a (GET/POST/...) Requests to Discord REST API with JSON data.
 func (s *Session) RequestWithBucketID(method, urlStr string, data interface{}, bucketID string) (response []byte, err error) {
+	return s.RequestWithHeaders(method, urlStr, data, bucketID, OptionalRequestHeaders{})
+}
+
+// RequestWithHeaders makes a (GET/POST/...) Requests to Discord REST API with JSON data and headers
+func (s *Session) RequestWithHeaders(method, urlStr string, data interface{}, bucketID string, optHeaders OptionalRequestHeaders) (response []byte, err error) {
 	var body []byte
 	if data != nil {
 		body, err = json.Marshal(data)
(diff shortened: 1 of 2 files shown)


Seklfreak:
Maybe I missed something, but I think you removed the ability to set the Days to delete at the `GuildBanCreateWithReason` method, don't see `days` used anywhere.

rei2hu:
My bad, I changed the method to set the header instead of set query parameters to pass the reason for a little consistency and missed it.

jonas747:
IIRC there is a difference between the audit log reason header and the reason query param in behaviour

The audit log reason wont show in the ban/kick event for example, i'm gonna do a couple tests in a little bit to see if this is still the case

rei2hu:
I did test if the setting the header would work for the ban method before I made the PR and it worked:
image
I also tested a few of the other methods like unban, role edit, and role create:
image
image
image

jonas747:
Yeah i think i remember wrong, disregard what i said.

jonas747:
ah yes i remember now, does it show up here aswell? in the bans? 
image

rei2hu:
It does not show up there; I will revert the ban change.

Do you think I should do the same for the GuildMemberDeleteWithReason method? As far as I know, the only place the reason for a kick shows up in is the audit logs, but seeing as how setting the header and passing the query parameter do not do exactly the same thing for ban, I think maybe it should be changed back, too.

iopred:
This PR looks pretty good but it drastically increases the complexity of the API. What do you think about making the API:

session.SetNextRequestHeaders(headers OptionalRequestHeaders)

Necroforger:
RequestWithBucketID accepts an interface for the data struct.
You could modify just the request and requestWithLockedBucket functions
to accept a `*RequestOptions` struct to avoid changing most of the API.

Then do a type check to decide how to treat the data.

```go
// RequestOptions allows you to supply additional information to
// the request functions
type RequestOptions struct {
	Headers [][]string
	Body    []byte
}
```

PR #514, opened 2018-03-07
Added UserConnections function and structs

Necroforger:
Adds a method to fetch your own connections.
Only works on a user account.

To get another user's connections you would need to fetch their profile while on a user account.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -112,6 +112,37 @@ type Session struct {
 	wsMutex sync.Mutex
 }
 
+// Connection is a Connection returned from the UserConnections endpoint
+type Connection struct {
+	ID           string         `json:"id"`
+	Name         string         `json:"name"`
+	Type         string         `json:"type"`
+	Revoked      bool           `json:"revoked"`
+	Integrations []*Integration `json:"integrations"`
+}
+
+// Integration stores integration information
+type Integration struct {
+	ID                string             `json:"id"`
+	Name              string             `json:"name"`
+	Type              string             `json:"type"`
+	Enabled           bool               `json:"enabled"`
+	Syncing           bool               `json:"syncing"`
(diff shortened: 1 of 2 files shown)


structs.go
```diff
@@ -112,6 +112,37 @@ type Session struct {
 	wsMutex sync.Mutex
 }
 
+// Connection is a Connection returned from the UserConnections endpoint
+type Connection struct {
```
iopred: Connection is pretty generic, could we call this UserConnection

Necroforger: GuildConnection and GuildAccount share the same fields, would it be possible to rename them to use the same structs

iopred: Guh, don't really want to break the API, would be happy embedding:

type GuildIntegration struct {
  Integration
}

structs.go
```diff
@@ -477,23 +508,7 @@ type GuildBan struct {
...
-	ID                string                   `json:"id"`
-	Name              string                   `json:"name"`
-	Type              string                   `json:"type"`
-	Enabled           bool                     `json:"enabled"`
-	Syncing           bool                     `json:"syncing"`
-	RoleID            string                   `json:"role_id"`
-	ExpireBehavior    int                      `json:"expire_behavior"`
-	ExpireGracePeriod int                      `json:"expire_grace_period"`
-	User              *User                    `json:"user"`
-	Account           *GuildIntegrationAccount `json:"account"`
```
iopred: Guh...... We break the API here too.. I think maybe we just break the API, sorry to make you undo your change here, maybe just change Guild to use Integration and be done with it. Searching seems to show that noone is using GuildIntegration on github.

Necroforger: Should I revert the last commit

PR #533, opened 2018-04-12
Changed State.MemberAdd to set fields individually

### LINKED ISSUE #532: Member.JoinedAt is wiped out by GuildMemberUpdate events
`Member.JoinedAt` (retrieved from `State`) seems to be getting wiped out by (some) GuildMemberUpdate events. I noticed this because when the bot first starts up, it can successfully retrieve the `JoinedAt` time for all guild members, but later, for any guild member whose roles are updated while the bot is running, the field becomes an empty string.

I've attached a screenshot of how this manifests with the particular bot I'm working on. The Member object is simply being retrieved like so:

```
	member, err := s.State.Member(guild.ID, userID)
	switch err {
	case nil:
		ui.Nickname = member.Nick
	case discordgo.ErrStateNotFound:
		ui.OnServer = false
	default:
		s.ChannelMessageSend(m.ChannelID, fmt.Sprintf("Error retrieving member info: %s", err))
	}
	//...
	fmt.Fprintf(b, "Joined At: %s\n", member.JoinedAt)

```

screen shot 2018-04-11 at 4 16 05 pm

CarsonHoffman:
This fixes #532. Though it's not mentioned on the Discord docs, it seems `Member.JoinedAt` isn't sent in member updates (it *is* sent in member chunks), which caused the value to be overwritten to an empty value after member updates. This pull request changes `State.MemberAdd` to handle each field individually, similarly to other function in the State that have similar dilemmas, checking for an empty `JoinedAt` field.

### DIFF
--- a/state.go
+++ b/state.go
@@ -299,7 +299,12 @@ func (s *State) MemberAdd(member *Member) error {
 		members[member.User.ID] = member
 		guild.Members = append(guild.Members, member)
 	} else {
-		*m = *member // Update the actual data, which will also update the member pointer in the slice
+		// We are about to replace `m` in the state with `member`, but first we need to
+		// make sure we preserve any fields that the `member` doesn't contain from `m`.
+		if member.JoinedAt == "" {
+			member.JoinedAt = m.JoinedAt
+		}
+		*m = *member
 	}
 
 	return nil


CarsonHoffman:
It seems the check failed to a network timeout completely unrelated to the code changed. :(

fuzzyfabricator:
Confirmed, this fixed the issue I reported in #532 for me. (And, just FTR, I built it fine with 1.10.1.)

state.go
```diff
@@ -299,7 +299,17 @@ func (s *State) MemberAdd(member *Member) error {
 		members[member.User.ID] = member
 		guild.Members = append(guild.Members, member)
 	} else {
-		*m = *member // Update the actual data, which will also update the member pointer in the slice
```
iopred: Could you implement this in the same way as we do in Guild, it will mean that we don't have to update 2 places whenever the Member struct changes, ie:

```
// We are about to replace `m` in the state with `member`, but first we need to
// make sure we preserve any fields that the `member` doesn't contain from `m`.
if member.JoinedAt == "" {
  member.JoinedAt = m.JoinedAt
}
*m = *member
```

CarsonHoffman: That looks like a better solution; will do.

PR #545, opened 2018-05-09
Added copy of previous Message state to MessageUpdate

ikkerens:
It is currently impossible to obtain a copy of the previous message content when receiving `MessageUpdate`, as the changes are always written to state before the message is sent to the event handler, even when SyncEvents is used.

This PR fixes that by introducing a (shallow) copy of the Message object, allowing comparison of the old and the new message.

### DIFF
--- a/events.go
+++ b/events.go
@@ -162,6 +162,8 @@ type MessageCreate struct {
 // MessageUpdate is the data for a MessageUpdate event.
 type MessageUpdate struct {
 	*Message
+	// BeforeUpdate will be nil if the Message was not previously cached in the state cache.
+	BeforeUpdate *Message `json:"-"`
 }
 
 // MessageDelete is the data for a MessageDelete event.
--- a/state.go
+++ b/state.go
@@ -863,6 +863,12 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 		}
 	case *MessageUpdate:
 		if s.MaxMessageCount != 0 {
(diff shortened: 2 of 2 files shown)


state.go
```diff
@@ -863,6 +863,13 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 		}
 	case *MessageUpdate:
 		if s.MaxMessageCount != 0 {
+			var old *Message
+			old, err = s.Message(t.ChannelID, t.ID)
+			if err == nil {
+				oldCopy := *old
+				t.BeforeUpdate = &oldCopy
```
iopred: `t.BeforeUpdate = old` should work here too, then you can get rid of oldCopy.

events.go
```diff
@@ -162,6 +162,8 @@ type MessageCreate struct {
 // MessageUpdate is the data for a MessageUpdate event.
 type MessageUpdate struct {
 	*Message
+	// BeforeUpdate will be nil if the Message was not previously cached in the state cache.
+	BeforeUpdate *Message `json:"-"`
```
iopred: PreviousMessage?

state.go
ikkerens: No? Using old would cause it to use a copy of the pointer, which is pointing to a struct that is about to be updated. By doing this I'm creating a shallow copy of the struct.

events.go
ikkerens: Would be potentially confusing, as people might think that points to the message before this one.

PR #550, opened 2018-05-20
Update guild object

Seklfreak:
I made a few changes to the `Guild` struct and `GuildParams` struct to stay up to date with the API.

I put them all in one PR to prevent merge conflicts. Please tell me if I should do it differently.

Some might break backwards compatibility. Please tell me if I should use a different approach.

Changes:
- add VerificationLevel Very High (4)
- add DefaultMessageNotificationLevel
  - use it in Guild Struct
  - use it in GuildParams (for modifying Guilds)
  - ⚠️ This might break backwards compatibility. 
- add ExplicitContentFilterLevel to GuildParams (for modifying Guilds)
- removed `omitempty` for `AfkChannelID` and `AfkTimeout` to be able to reset it.
  - ⚠️ This might break backwards compatibility. 
  - ⚠️ I forgot to put that in a separate commit. (it's in 09ff136f06ff22fb7b7ca34d6d7185aa7bf5f186)

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -288,12 +288,13 @@ func (e *Emoji) APIName() string {
 // VerificationLevel type definition
 type VerificationLevel int
 
-// Constants for VerificationLevel levels from 0 to 3 inclusive
+// Constants for VerificationLevel levels from 0 to 4 inclusive
 const (
 	VerificationLevelNone VerificationLevel = iota
 	VerificationLevelLow
 	VerificationLevelMedium
 	VerificationLevelHigh
+	VerificationLevelVeryHigh
 )
 
 // ExplicitContentFilterLevel type definition
--- a/structs.go
+++ b/structs.go
@@ -316,6 +316,15 @@ const (
 	MfaLevelElevated
 )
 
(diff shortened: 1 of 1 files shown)


iopred:
The afkChannelID Change worries me a bit because anyone relying on it being empty by default will have their AfkChannel wiped out.

The other changes seem fine because I can't find any usages of DefaultMessageNotifications around the web.

Seklfreak:
Thanks for checking it out!

I see two solutions for the concern you raised:

1.
We could use `*string` as the type for `AfkChannelID ` instead of `string`, and `*int` as the type for `AfkTimeout ` instead of `int`, and add `omitempty ` back for both. This way not setting the field will not send it to discord, and setting the field with `""` or `0` will send the empty value to discord (to reset it). It might break some projects however, because they need to set the changed types now. It won't break it if they haven't set the field.

2. 
We will just add `omitempty` back for both. Nothing will break, but it will be impossible to reset the fields using the discordgo API.

I'm happy to implement both, but I don't know which one to choose. Maybe you can decide it. :)

PR #557, opened 2018-06-07
Add userlimit to the channel struct

SilverCory:
I'd add the ApplicationID, OwnerID and LastPinTimeStamp but I'm doing this in the web editor.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -239,6 +239,9 @@ type Channel struct {
 	// A list of permission overwrites present for the channel.
 	PermissionOverwrites []*PermissionOverwrite `json:"permission_overwrites"`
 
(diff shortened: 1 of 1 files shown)


structs.go
```diff
@@ -238,6 +238,9 @@ type Channel struct {
 
 	// A list of permission overwrites present for the channel.
 	PermissionOverwrites []*PermissionOverwrite `json:"permission_overwrites"`
+
```
iopred: This extra tab breaks the build :|

SilverCory: Should be resolved and amended waiting for new build to complete

iopred:
Thanks for the PR!

PR #561, opened 2018-06-20
add Webhooks Update event

Seklfreak:
This adds support for the event fired when Webhooks get created, removed, or updated. It is documented here https://discordapp.com/developers/docs/topics/gateway#webhooks-update .


qaisjp:
thanks!

PR #567, opened 2018-07-25
Remove redundant line

Fogapod:
s.StateEnabled variable already set to true above


bwmarrin:
Good catch @Fogapod, thanks.

PR #569, opened 2018-07-31
Added .MessageFormat() to emoji struct

ikkerens:
.APIName currently covers all REST url cases, but with the introduction of animated Emoji the chat usage of Emoji has become slightly less obvious as well, despite bots being able to use them.

This new func makes usage of custom emoji (including animated ones) easier.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -281,8 +281,8 @@ type Emoji struct {
 	Animated      bool     `json:"animated"`
 }
 
-// ChatName returns a correctly formatted Emoji for use in Message content and embeds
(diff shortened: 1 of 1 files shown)


structs.go
```diff
@@ -281,6 +281,19 @@ type Emoji struct {
 	Animated      bool     `json:"animated"`
 }
 
+// ChatName returns a correctly formatted Emoji for use in Message content and embeds
+func (e *Emoji) ChatName() string {
```
iopred: MessageName ?

ikkerens: Pushed change.

PR #570, opened 2018-07-31
Changed Member.JoinedAt to type Timestamp

pixelrazor:
It was previously just string, but has been updated to be consistent with the other structs.


bwmarrin:
Thanks @pixelrazor

PR #576, opened 2018-08-20
Expose SessionID in VoiceConnection struct

TurtleGamingFTW:


bwmarrin:
I believe this was requested due to @TurtleGamingFTW trying to get Lavalink working with DiscordGo, but from further conversation in Discord exposing this may not be needed.  Another solution is proposed here https://github.com/bwmarrin/discordgo/pull/578 which the user states is tested and working with Lavalink.

So I am going to close this out, if there's another reason to expose this then feel free to comment/reopen it.

PR #588, opened 2018-09-26
Add LastHeartbeatSent to Session struct

cxnky:
This will be used for the ``Latency()`` method which will be added to ``Session``


cxnky:
#589 will fix this build

cxnky:
going to resubmit as one

PR #592, opened 2018-09-26
Add Latency method

cxnky:


iopred:
Could you unify all these PR's into a single one that works (add method and properties), and could you make it into the develop branch.

Thank you so much!

Issue #587, opened 2018-09-24
implement "heartbeat latency"


cxnky:
hi, 

i was told to submit an issue here reminding you to create a thing which allows for heartbeat latency (or something) to be implemented, so we could get that through discordgo

PR #593, opened 2018-09-26
Add latency method

cxnky:
This adds a method which will be updated whenever a heartbeat is sent, to retrieve the latency in ``ms``.

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -267,9 +267,16 @@ type helloOp struct {
 // FailedHeartbeatAcks is the Number of heartbeat intervals to wait until forcing a connection restart.
 const FailedHeartbeatAcks time.Duration = 5 * time.Millisecond
 
+// latency returns the latency between heartbeat acknowledgement and heartbeat send
+func (s *Session) Latency() int64 {
+
+	return s.LastHeartbeatAck.Sub(s.LastHeartbeatSent).Nanoseconds() / 1000000
+
+}
+
 // heartbeat sends regular heartbeats to Discord so it knows the client
 // is still connected.  If you do not send these heartbeats Discord will
-// disconnect the websocket connection after a few seconds.
+// disconnect the websocket connection after a few seconds
 func (s *Session) heartbeat(wsConn *websocket.Conn, listening <-chan interface{}, heartbeatIntervalMsec time.Duration) {
 
 	s.log(LogInformational, "called")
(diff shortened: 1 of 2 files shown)


cxnky:
resolves #587

wsapi.go
```diff
@@ -267,9 +267,16 @@ type helloOp struct {
 // FailedHeartbeatAcks is the Number of heartbeat intervals to wait until forcing a connection restart.
 const FailedHeartbeatAcks time.Duration = 5 * time.Millisecond
 
+// latency returns the latency between heartbeat acknowledgement and heartbeat send
+func (s *Session) Latency() int64 {
```
iopred: Can this method simply return the time.Time?

wsapi.go
```diff
@@ -267,9 +267,16 @@ type helloOp struct {
 // FailedHeartbeatAcks is the Number of heartbeat intervals to wait until forcing a connection restart.
 const FailedHeartbeatAcks time.Duration = 5 * time.Millisecond
 
+// latency returns the latency between heartbeat acknowledgement and heartbeat send
```
iopred: Perhaps we can name this HeartbeatLatency? Latency would lead me to believe this is the round trip/ping time.

wsapi.go
cxnky: It could, but I just thought returning milliseconds would make things easier.

wsapi.go
cxnky: Sure, sounds good.

wsapi.go
iopred: Returning time.Time is a little more standard and explicit, I don't need to guess at the unit of the value, I didn't realise you were returning millis for example.

cxnky: Ah, I see where you're coming from now. So just return ``s.LastHeartbeatAck.Sub(s.LastHeartbeatSent)``?

iopred: Yeah I think that would be great.

wsapi.go
```diff
@@ -267,9 +267,16 @@ type helloOp struct {
...
+func (s *Session) HeartbeatLatency() time.Duration {
+
+	return s.LastHeartbeatAck.Sub(s.LastHeartbeatSent)
+
+}
+
 // heartbeat sends regular heartbeats to Discord so it knows the client
 // is still connected.  If you do not send these heartbeats Discord will
-// disconnect the websocket connection after a few seconds.
+// disconnect the websocket connection after a few seconds
```
iopred: Sorry, last change, you should undo this change (comments should end in periods), you should also add a period in the comment for your change! Thanks so much.

wsapi.go
cxnky: It's gonna have to return ``time.Duration``, since ``Sub()`` returns a Duration

wsapi.go
cxnky: Will do

PR #603, opened 2018-10-24
travis: to use official go lint path

roberthmiller:
Saw that Travis was failing and realized that the problem is with the go lint path. I had a look at the reasoning and the path has changed. This change can be viewed in the readme: https://github.com/golang/lint/commit/ead987a65e5c7e053cf9633f9eac1f734f6b4fe3


iopred:
Please PR this into the develop branch, thank you!

roberthmiller:
This is I guess still a WIP due to travis still not passing. I am working to make golint optional in the build process. Would this be something that is wanted?

PR #608, opened 2018-10-28
Feat(MessageReactions): afterID and beforeID

xDimGG:
API support for the afterID and beforeID of reactions.


CarsonHoffman:
This functionality is now present on `develop`.

PR #609, opened 2018-11-03
check for empty gID/cID

tyhi:
Sending a empty string causes the session to disconnect for then reconnect a few seconds later. Sending a null/nil string pointer fixes this and gives the expected result of just disconnecting from the voice channel.

Also added a check for gID as there's no point to send with no guild ID and it will just temporarily disconnect the session.

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -649,6 +649,17 @@ func (s *Session) ChannelVoiceJoinManual(gID, cID string, mute, deaf bool) (err 
 
 	// Send the request to Discord that we want to join the voice channel
 	data := voiceChannelJoinOp{4, voiceChannelJoinData{&gID, &cID, mute, deaf}}
+	
+	// If a empty string is sent the session will temporarily disconnect, unless it's null(nil)
+	if cID == "" {
+		data.Data.ChannelID = nil
+	}
+	
+	// If the gID is empty the session will temporarily dissconnect, returning prevents this.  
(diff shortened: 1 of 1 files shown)


robbix1206:
Shouldn't that be some error cases ? Returning nothing in the gID case is a bad idea from my point of view as it is an error to send an empty ID. And isn't that the case for invalid snowflake ?
An other thing is that you should operate on gID and cID before creating the voiceChannelJoinOp I think.
And I think there is the same problem in the ChannelVoiceJoin.

soowan:
> Returning nothing in the gID case is a bad idea from my point of view as it is an error to send an empty ID.

Very true, I tried looking through and seeing how to return the error but I wasn't able to compare it with anything? Sorry, still pretty new to Go.

> An other thing is that you should operate on gID and cID before creating the voiceChannelJoinOp I think.

How would you go about assigning nil to the string before it's converted to a pointer? From what I understand nil isn't a type for strings, that's why I did it after creating voiceChannelJoinOp.

> And I think there is the same problem in the ChannelVoiceJoin.

I can add the same nil check for the cID, once I know how I should return the error from no gID if necessary.

Thanks!

bwmarrin:
@tyhi does https://github.com/bwmarrin/discordgo/pull/672 take care of things for this?

tyhi:
Yeah this can be closed, thanks!

PR #615, opened 2018-11-12
Escape # in emoji IDs passed to reaction endpoints

### LINKED ISSUE #614: Can't get reactions for #⃣
Steps to reproduce: make a new channel, send a message, react to it with  #⃣ (`:hash:`), run the code below:
```
package main

import (
	"flag"
	"log"
	"os"
	"os/signal"
	"syscall"

)

var token = flag.String("t", "", "access token")

func do(d *discordgo.Session, _ *discordgo.Ready) {
	cid := "<replace with id of your channel>"
	msgs, err := d.ChannelMessages(cid, 100, "", "", "")
	if err != nil {
		log.Fatalf("error getting messages: %v", err)
	}

	for _, m := range msgs {
		for _, r := range m.Reactions {
			_, err = d.MessageReactions(cid, m.ID, r.Emoji.APIName(), 100)
			if err != nil {
				log.Fatalf("error getting message reactions for %s: %v", r.Emoji.APIName(), err)
			}
		}
	}

	os.Exit(0)
}

func main() {
	flag.Parse()

	d, err := discordgo.New(*token)
	if err != nil {
		log.Fatal("login failed:", err)
	}

	d.AddHandler(do)

	err = d.Open()
	defer d.Close()
	if err != nil {
		log.Fatal("opening the websocket connection failed:", err)
	}

	sc := make(chan os.Signal, 1)
	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM, os.Interrupt, os.Kill)
	<-sc
}
```

Output:
```
2018/11/12 18:04:37 error getting message reactions for #⃣: HTTP 404 Not Found, {"code": 0, "message": "404: Not Found"}
```

tsudoko:
Looks like `net/http` strips URL fragments, which causes emoji with `#` in names to be sent incorrectly to some endpoints.

Fixes #614.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2085,6 +2085,8 @@ func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *Webho
 // emojiID   : Either the unicode emoji for the reaction, or a guild emoji identifier.
 func (s *Session) MessageReactionAdd(channelID, messageID, emojiID string) error {
 
+	// emoji such as  #⃣ need to have # escaped
+	emojiID = strings.Replace(emojiID, "#", "%23", -1)
 	_, err := s.RequestWithBucketID("PUT", EndpointMessageReaction(channelID, messageID, emojiID, "@me"), nil, EndpointMessageReaction(channelID, "", "", ""))
 
 	return err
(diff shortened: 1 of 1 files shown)


iopred:
I think the correct thing to do is escape our strings when they are included as part of the URL.

PR #633, opened 2019-02-23
Add Pinned to Message struct

JurrijnP:


JurrijnP:
Can somebody explain why this failed?

Hinara:
You have a empty line with spaces which should not be there

message.go
```diff
@@ -80,6 +80,9 @@ type Message struct {
 	// A list of reactions to the message.
 	Reactions []*MessageReactions `json:"reactions"`
 
+	// Whether the message is pinned or not.
+	Pinned bool `json:"pinned"`
+
```
Hinara: The empty line there should not have spaces

JurrijnP: Ah thanks will edit that

PR #634, opened 2019-02-23
Enable ChannelVoiceJoinManual to leave channel

mtharp:
Currently if you try to leave voice by calling ChannelVoiceJoinManual with a channelID of "" it invalidates the session and forces a reconnect (Op9). This fixes it so it sends the null that discord is expecting.


bwmarrin:
This was resolved with https://github.com/bwmarrin/discordgo/pull/672

However I do really appreciate you helping with the project!

Issue #629, opened 2019-02-07
Unable to retrieve bot's oauth application


hassieswift621:
Hi, doing the following gives me an error:
```go
// Get bot app and bot user.
app, err := command.client.Connection().Application(command.client.Connection().State.User.ID)
```
2019/02/07 21:21:48 Failed to get bot app info. HTTP 403 Forbidden, {"code": 20001, "message": "Bots cannot use this endpoint"}

Fetching the bot account works in other API libraries such as Discord4J and Eris JS, however with this lib, Im getting an error. Any workarounds?

hassieswift621:
Fix: Use /oauth2/applications/@me endpoint
https://discordapp.com/developers/docs/topics/oauth2#get-current-application-information

bwmarrin:
Maybe we should add some comments in the code to help make this more clear.

PR #647, opened 2019-05-14
Fix for fetching application info for current bot

hassieswift621:
This is a fix for issue #629, which now allows the user to fetch application info for the bot.

### DIFF
--- a/oauth2.go
+++ b/oauth2.go
@@ -33,6 +33,11 @@ type Application struct {
 //   appID : The ID of an Application
 func (s *Session) Application(appID string) (st *Application, err error) {
 
+	// If the app ID is for the current bot, use @me as the ID to allow the bot to fetch its application.
+	if appID == s.State.User.ID && s.State.User.Bot {
+		appID = "@me"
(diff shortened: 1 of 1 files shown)


bwmarrin:
I mixed on this.  This is a bit of the type of "magic" I'd prefer to keep out of the library.  We expose as close a match to discord's interface without magic as we can.  I'll think about it.  Open to comments by others.

oauth2.go
```diff
@@ -33,6 +33,11 @@ type Application struct {
 //   appID : The ID of an Application
 func (s *Session) Application(appID string) (st *Application, err error) {
 
+	// If the app ID is for the current bot, use @me as the ID to allow the bot to fetch its application.
+	if appID == s.State.User.ID && s.State.User.Bot {
```
iopred: Does this work if State is disabled? It seems like it might actually panic.

bwmarrin: Good point.

hassieswift621: Well spotted

PR #648, opened 2019-05-15
Expose all Emojis and allow a member dump

dancodes:
I've added two methods that allow users to

* Get a list of all emojis added to the Guild
* Get a deep copy of the memberMap used in State

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1310,6 +1310,22 @@ func (s *Session) GuildAuditLog(guildID, userID, beforeID string, actionType, li
 	return
 }
 
+// GuildEmojis returns an array of Emoji structures for all emojis of a
+// given guild.
+// guildID : The ID of a Guild.
+// name    : The Name of the Emoji.
+// image   : The base64 encoded emoji image, has to be smaller than 256KB.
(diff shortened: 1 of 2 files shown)


restapi.go
```diff
@@ -1310,6 +1310,22 @@ func (s *Session) GuildAuditLog(guildID, userID, beforeID string, actionType, li
 	return
 }
 
+// GuildEmojis returns an array of Emoji structures for all emojis of a
+// given guild.
+// guildID : The ID of a Guild.
+// name    : The Name of the Emoji.
+// image   : The base64 encoded emoji image, has to be smaller than 256KB.
+// roles   : The roles for which this emoji will be whitelisted, can be nil.
+func (s *Session) GuildEmojis(guildID string) (emojis []*Emoji, err error) {
```
Strum355: the comment doesnt seem to match the method signature, not sure what name, image and roles are for given theyre not in the parameter list

CarsonHoffman:
Given that this PR is opened against `master`, it's going to be unmergeable no matter what changes are made. If you would like these changes in the library, please open a new PR against the `develop` branch, incorporating the changes mentioned above.

Issue #649, opened 2019-05-26
Rethink `New` function API


UlisseMini:
# Problem
The `New` function does not express its intent in its type. it is also very unintuitive to use.
```go
func New(args ...interface{}) (s *Session, err error)
```
The type signature tells the user nothing, i think this API was picked because of different ways to authenticate. I think this is a bad idea and it is more intuitive to use a struct.

# Solution
Add a new more intuitive struct based API. something along the lines of this
```go
dg, err := discordgo.WithAuth {
    Email: "foo@bar.com",
    Pass: "foobar",
    Token: "abc123",
}.Verify()
```
I'm using a method here to prevent me from having to type in`discordgo` again. you could also do it like this
```go
dg, err := discordgo.New(discordgo.WithAuth{
    Email: "foo@bar.com",
    Pass: "foobar",
    Token: "abc123",
})
```
In my opinion both are better then the current API, and could be added while remaining backwards compatible.

If this gets approved and i have time i'll implement this (it should not be that hard).

iopred:
I like where this is heading, an API where you can pass in a WithAuth
option, or WithToken option sounds great to me

diamondburned:
-1, You don't always need an email, a password and a token. I think the current system is fine as it is.

Luckily, since the current `New` function takes in an `interface{}`, you can add this without breaking, and I'm all up for that.

UlisseMini:
@diamondburned If you don't specify a field then it will be set to its zero value, exactly what we want.

for example, not specifying the `Bot` field would leave it to a default of `false`
not specifying an email and password would leave them to `""` and we could check for that in the `New` / `Verify` functions

bwmarrin:
Personally, if we're to break the current API on everyone - I'm more in favor of dropping the Email/Pass entirely and just accepting a single token string that we pass along to Discord.  It's a closer match to the goal of just being a wrapper around Discord's API and it's simple and easy to use.

UlisseMini:
I agree with @bwmarrin, it would also make the API much simpler, something like
```go
func New(token string) (*Session, error)
```
would be desirable

pollen5:
You could also remove the error return value entirely in that case, since currently it's only used for paramater validation and fetching token from email/pass, if it takes only token no error would be returned.

diamondburned:
My PR should add something like this without any breaking changes: https://github.com/bwmarrin/discordgo/pull/725

PR #651, opened 2019-05-27
Add auth struct to use with `New`

UlisseMini:
WIP This is a proof of concept and not done yet! We need discussion about the API and whether to create a new function / method or to just add it to `New`.

I think we should deprecate `New` and add a new function (or method on the `Auth` struct) to get a session.

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -26,6 +26,17 @@ const VERSION = "0.19.0"
 // ErrMFA will be risen by New when the user has 2FA.
 var ErrMFA = errors.New("account has 2FA enabled")
 
+// Auth represents authentication options to use with new.
+type Auth struct {
+	Email    string // Email to use when authenticating to a user account.
+	Password string // Password to use when authenticating to a user account.
+
+	// Token to use when authenticating with discord.
+	// If the token is for a bot then the token must be prefixed with "Bot"
+	// eg "Bot <token>"
+	Token string
+}
+
 // New creates a new Discord session and will automate some startup
 // tasks if given enough information to do so.  Currently you can pass zero
 // arguments and it will return an empty Discord session.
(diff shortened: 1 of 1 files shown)


discord.go
```diff
@@ -26,6 +26,17 @@ const VERSION = "0.19.0"
 // ErrMFA will be risen by New when the user has 2FA.
 var ErrMFA = errors.New("account has 2FA enabled")
 
+// Auth represents authentication options to use with new.
+type Auth struct {
```
iopred: I like the idea of this being a PasswordAuth struct, and then having 2 fields. Then have a TokenAuth struct with just one field.

No reason to overload the two methods of authentication again, it was confusing as arguments, I don't want it to be confusing with struct members.

UlisseMini: I just thought it should have all three, since in the docs for `New` it says
```
With an email, password and auth token - Discord will verify the auth
token, if it is invalid it will sign in with the provided
credentials. This is the Discord recommended way to sign in.
```
So i thought it would be bad to remove the option of using both.

Soumil07:
IMO getting rid of email/password login would be a good thing, now that Discord doesn't support it.

CarsonHoffman:
I personally think that this is over-engineering a solution to support a feature that nobody really uses. Email+password login has been [officially unsupported](https://github.com/discordapp/discord-api-docs/issues/69#issuecomment-223886862) for three years now, and other major libraries have dropped support for it long ago (see [discord.py](https://github.com/Rapptz/discord.py/commit/4b6b5bd35eb59a0dac72039ce5bbd21c2ad26d6a): 2017, [discord.js](https://github.com/discordjs/discord.js/commit/5afd77ab730f413d0090be6ce0b0a7083ef15f7c#diff-49433f8a8f4ae91e442a338fb263e967): 2018). I have yet to run into a single person seriously attempting to use it in the time that I've been around this library. I think it would be much simpler to just support a string token, which can be used for bots, bearer tokens, and, *yes*, self-bots, as although they are against TOS they still hold the same major authentication scheme, and hard-coding a `Bot` token scheme would break the use of bearer tokens, and people wanting to self-bot are just going to edit the library anyways.

bwmarrin:
Personally, if we're to break the current API on everyone - I'm more in favor of dropping the Email/Pass entirely and just accepting a single token string that we pass along to Discord. It's a closer match to the goal of just being a wrapper around Discord's API and it's simple and easy to use.

The only reason this hasn't been done is because of not wanting to break the API on everyone for such a minor thing.  Now with modules and versions and such, I do feel better about making breaking changes to help clean up this library.

UlisseMini:
I agree with @bwmarrin little point to support email/pass auth, it adds a lot more complexity for little gain.

I'm closing, we can continue this in #649

PR #652, opened 2019-05-27
[Feature] Add injectable logger

Soumil07:
Makes it easier to use a custom logger by defining a Logger interface, and allowing users to inject their own.


bwmarrin:
@MostHated,

You need only override the Logger variable here to use your own custom logger.

https://github.com/bwmarrin/discordgo/blob/master/logging.go#L37

PR #653, opened 2019-05-28
Added user-agent

datutbrus:


bwmarrin:
Just curious, what's the reason for exposing the UserAgent? I though Discord wanted it to be presented in the specific format we were using.  What use case would an end-user of the library need to use a different user agent?

trvv:
you're supposed to be able to add more metadata to the user agent than just the project URL and version. whether or not this would actually be necessary for anything is up for debate but it should still be something people can do if they want.

PR #663, opened 2019-06-17
Fix #622: support for WebhookExecute msg response

### LINKED ISSUE #622: ExecuteWebhook (and similar functions) return value?
The `ExecuteWebhook` function should return a value describing the message sent.  Here is an example response body:

```
{
    "attachments": [], 
    "author": {
        "avatar": null, 
        "bot": true, 
        "discriminator": "0000", 
        "id": "5316xxxxxxxxxxxxxxx", 
        "username": "matterbridge"
    }, 
    "channel_id": "5306xxxxxxxxxxxxxxx", 
    "content": "test", 
    "edited_timestamp": null, 
    "embeds": [], 
    "id": "5317xxxxxxxxxxxxxxx", 
    "mention_everyone": false, 
    "mention_roles": [], 
    "mentions": [], 
    "nonce": null, 
    "pinned": false, 
    "timestamp": "2019-01-07T08:00:00.000000+00:00", 
    "tts": false, 
    "type": 0, 
    "webhook_id": "5316xxxxxxxxxxxxxxx"
}
```

Here is the relevant code:

https://github.com/bwmarrin/discordgo/blob/8325a6bf6dd6c91ed4040a1617b07287b8fb0eba/restapi.go#L2067-L2080

qaisjp:
- Should the API be changed?
- Is the "wait" doc described fine?

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2067,14 +2067,20 @@ func (s *Session) WebhookDeleteWithToken(webhookID, token string) (st *Webhook, 
 // WebhookExecute executes a webhook.
 // webhookID: The ID of a webhook.
 // token    : The auth token for the webhook
-func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *WebhookParams) (err error) {
+// wait     : Waits for server confirmation of message send and ensures that the return struct is populated (it is nil otherwise)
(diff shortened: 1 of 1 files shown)


iopred:
This is a breaking change, but I really can't argue for it. My apologies to anyone who has to add a  `_,` to their code.

PR #666, opened 2019-06-21
Added Asset and ApplicationAssets

datutbrus:
The `/assets` endpoint is used in `https://discordapp.com/developers/applications/$APP_ID/rich-presence/assets` for Rich Presence assets.

### DIFF
--- a/oauth2.go
+++ b/oauth2.go
@@ -105,6 +105,25 @@ func (s *Session) ApplicationDelete(appID string) (err error) {
 	return
 }
 
+// Asset struct stores values for an asset of an application
+type Asset struct {
+	Type int    `json:"type"`
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+// ApplicationAssets returns an application's assets
+func (s *Session) ApplicationAssets(appID string) (ass []*Asset, err error) {
+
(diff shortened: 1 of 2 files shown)


Strum355:
Might be a good idea to cleanup the commits to get rid of the 2nd one. For future reference, use rebasing instead

datutbrus:
Can you give me the commands ~~please~~

datutbrus:
Ayy I did it

oauth2.go
```diff
@@ -105,6 +105,25 @@ func (s *Session) ApplicationDelete(appID string) (err error) {
...
 
+// Asset struct stores values for an asset of an application
+type Asset struct {
+	Type int    `json:"type"`
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+// ApplicationAssets returns an application's assets
+func (s *Session) ApplicationAssets(appID string) (ass []*Asset, err error) {
```
mlvzk: Why is Asset a pointer here? On each index access there's gonna be a double pointer dereference, and since it's a slice the contents won't get copied anyway.
```suggestion
func (s *Session) ApplicationAssets(appID string) (ass []Asset, err error) {
```

iopred: Slice of pointers is consistent with the rest of the library.

PR #667, opened 2019-06-21
Update handlers to receive an interface for the Session

### LINKED ISSUE #662: proposal: update handler function signatures to use an interface for the session
I've started to use Discord Go to write a bot, and have found it somewhat difficult to properly test without writing my own shimmed handlers.

This challenge exists because the first argument to handlers is a struct type and not an interface, so a mock implementation can't be provided to assert proper behaviors. The standard library has solved for this exact problem as the first argument to a handler (`http.ResponseWriter`) is an interface, so that alternative implementations can be provided for testing.

I'd like to propose a breaking change to the library to update the handler signature to go from `discordgo.Session` being a struct to it becoming an interface. So from:

```go
func handler(s *discordgo.Session, <eventType>) {}
```

to

```go
func handler(s discordgo.Session, <eventType>) {}
```

The benefit of making this change is that Discord Go handler functions / methods could be tested directly, without needing to write shims for each handler you implement. This would make it much easier to test, and the ease of which could be shown-off in the example projects. It also may not be a major _breaking_ change for some users, if they only use the methods on the type, which would be a good benefit.

One downside of this change is that the size of the `*discordgo.Session` method set is large today. Even using composition, a `discordgo.Session` interface would end up becoming quite large. We would also need to add more methods, getters and setters, to tweak the internal settings.

Overall, I think this would be a net-win for the project and would like to encourage its inclusion in the 0.20.0 development cycle.

theckman:
Please be aware, this is a breaking API change.

This change adds a new `discordgo` type, `Sessioner`, which is an interface that
describes the full method set for the `*Session` type. In addition to the new
type, this change updates the package to no longer send a concrete `*Session`
value in to handlers it invokes as the first argument but a `Sessioner` instead.
This is to make testing of handlers much easier, as mock implementations of the
Sessioner interface can be passed right in.

This also renames the internal state field from `State` to `state`, and adds a
new method to `*Session`, `State()`, to expose that internal state to consumers.
A test was added to ensure this method works as expected.

The example projects were updated where necessary to work with the new API

Fixes #662.

Signed-off-by: Tim Heckman <t@heckman.io>

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -49,7 +49,7 @@ func New(args ...interface{}) (s *Session, err error) {
 
 	// Create an empty Session interface.
 	s = &Session{
-		State:                  NewState(),
+		state:                  NewState(),
 		Ratelimiter:            NewRatelimiter(),
 		StateEnabled:           true,
 		Compress:               true,
--- a/discord_test.go
+++ b/discord_test.go
@@ -137,17 +137,17 @@ func TestOpenClose(t *testing.T) {
 func TestAddHandler(t *testing.T) {
 
 	testHandlerCalled := int32(0)
-	testHandler := func(s *Session, m *MessageCreate) {
+	testHandler := func(s Sessioner, m *MessageCreate) {
 		atomic.AddInt32(&testHandlerCalled, 1)
 	}
 
 	interfaceHandlerCalled := int32(0)
-	interfaceHandler := func(s *Session, i interface{}) {
+	interfaceHandler := func(s Sessioner, i interface{}) {
 		atomic.AddInt32(&interfaceHandlerCalled, 1)
 	}
 
 	bogusHandlerCalled := int32(0)
-	bogusHandler := func(s *Session, se *Session) {
+	bogusHandler := func(s Sessioner, se *Session) {
 		atomic.AddInt32(&bogusHandlerCalled, 1)
 	}
 
@@ -181,7 +181,7 @@ func TestAddHandler(t *testing.T) {
 func TestRemoveHandler(t *testing.T) {
 
 	testHandlerCalled := int32(0)
-	testHandler := func(s *Session, m *MessageCreate) {
+	testHandler := func(s Sessioner, m *MessageCreate) {
 		atomic.AddInt32(&testHandlerCalled, 1)
 	}
 
--- a/event.go
+++ b/event.go
@@ -8,7 +8,7 @@ type EventHandler interface {
 	// Handle is called whenever an event of Type() happens.
 	// It is the receivers responsibility to type assert that the interface
 	// is the expected struct.
-	Handle(*Session, interface{})
+	Handle(Sessioner, interface{})
 }
 
 // EventInterfaceProvider is an interface for providing empty interfaces for
@@ -27,7 +27,7 @@ type EventInterfaceProvider interface {
 const interfaceEventType = "__INTERFACE__"
 
 // interfaceEventHandler is an event handler for interface{} events.
-type interfaceEventHandler func(*Session, interface{})
+type interfaceEventHandler func(Sessioner, interface{})
 
 // Type returns the event type for interface{} events.
 func (eh interfaceEventHandler) Type() string {
@@ -35,7 +35,7 @@ func (eh interfaceEventHandler) Type() string {
 }
 
 // Handle is the handler for an interface{} event.
-func (eh interfaceEventHandler) Handle(s *Session, i interface{}) {
+func (eh interfaceEventHandler) Handle(s Sessioner, i interface{}) {
 	eh(s, i)
 }
 
@@ -233,7 +233,7 @@ func (s *Session) onInterface(i interface{}) {
 	case *VoiceStateUpdate:
 		go s.onVoiceStateUpdate(t)
 	}
-	err := s.State.OnInterface(s, i)
+	err := s.state.OnInterface(s, i)
 	if err != nil {
 		s.log(LogDebug, "error dispatching internal event, %s", err)
 	}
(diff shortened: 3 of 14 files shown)


theckman:
There's one thing I'm not sure of with this change, and it's that it's not a compilation breaking change. If someone pulled in the new version of the library, their code would compile fine but not work as expected. To make that happen, we'd need to rename the `Session` type to something else. Thoughts?

Grafikart:
Maybe, if you want to keep compatibility with previous versions is it possible to duplicate every type in the switch ? 

```go
func handlerForInterface(handler interface{}) EventHandler {
	switch v := handler.(type) {
	case func(*Session, *ChannelCreate):
		return channelCreateEventHandler(v)
	case func(Sessioner, *ChannelCreate):
		return channelCreateEventHandler(v)
```

otothea:
> Maybe, if you want to keep compatibility with previous versions is it possible to duplicate every type in the switch ?
> 
> ```go
> func handlerForInterface(handler interface{}) EventHandler {
> 	switch v := handler.(type) {
> 	case func(*Session, *ChannelCreate):
> 		return channelCreateEventHandler(v)
> 	case func(Sessioner, *ChannelCreate):
> 		return channelCreateEventHandler(v)
> ```

@theckman I completely agree that it would be better for the lib to use interfaces instead of concrete types. I went to write tests and quickly decided not to mess with it since I cannot easily mock anything. However, why does it need to make the state private and change State to a function? If you revert that change _and_ @Grafikart's suggestion works, I think this could be a non-breaking change.

theckman:
@otothea if you look at the examples this updates, you'll see how that's sometimes used. I opted to add it because someone may want to use it, and if someone is going to have to do a type assertion to get at that field that code will have similar issues with testing we have today. By adding the methods, those who need that field can still get it and the code will be testable.

otothea:
@theckman ok, I see what you mean. Maybe we could leave `State` as is and add `GetState()` or `StateV2()` instead to avoid breaking things? Either way, I like this PR, thanks for taking the time to do it. I may merge this into my fork and start using it.

theckman:
@otothea I had considered that first suggestion, but things named `GetX` in Go are generally not considered idiomatic..

Thank you so much, I hope you can make good use of this code!

bwmarrin:
This is an interesting idea, but maybe it would be better suited for a v2 of the library.  I don't want to make such a substantial change to the library on all the users currently using it.

theckman:
@bwmarrin Thank you for your reply. Are you able to fill in a bit more context to help me? When is V2 of the library due? How is that being tracked? What's the largest blocker for leaning in to a major version bump for a package improvements?

bwmarrin:
I love the idea of working on a V2 that cleans up and refactors a lot of the mess that this library has grown over the years of Discord API changes - but I'm not sure what the realistic timeline would be based on available time.

This issue with this PR is it's pretty substantial change and there's thousands of people using the library.  I hate to make such a large change - however I'm open to additional comments and feedback from members of the community.  If it's there's a large number of people that need this - then maybe we could revisit making the change sooner than later.

ewohltman:
+1 for changes that can help developers write unit tests.

I've been getting by with writing a "mock" `Session` and replacing the embedded `Session.Client` with an `http.Client` using a custom `http.RoundTripper` to avoid making real API calls. While this works for a smaller project like my own, I admit this could potentially get unwieldy to mock all endpoint responses if this approach was provided by `discordgo`.

Example: https://github.com/ewohltman/ephemeral-roles/blob/master/pkg/mock/mock.go#L60

dcarriger:
+1 from me as well. I'm a fairly new Go developer, I'm trying to learn best practices for all of my private projects and using TDD....and writing tests for my Discord bot is proving to be a pain because Sessions aren't sufficiently abstracted.

CarsonHoffman:
Hi! This was unfortunately part of a mistake on my end—in the process of streamlining the PR submission process by eliminating the `develop` branch (due to the versioning capabilities of modules that didn't exist when the branch was created), all PRs against the branch were instantly closed without much warning; in hindsight this was a poor decision on my end, but there is unfortunately no remedy. I'm sorry that you got caught in this.

That said, I would suggest inverting the relationship between the interface and your code to aid in testing—I've found that it usually works best when *your* code defines the functionality that it needs by defining its *own* interfaces describing its required functionality which the `Session` type might satisfy. This prevents the `Session` type and a hypothetical `Sessioner` from falling out of sync, and means that a mock implementation only needs to provide the functionality actually required by your code, rather than the entire `Sessioner` interface. A succinct description of this pattern that I often fall back to is [this tweet](https://twitter.com/davecheney/status/942593128355192832) from Dave Cheney.

theckman:
@CarsonHoffman I agree with you. When you consider that I am writing these functions (handlers) to be injected into `discordgo`, doesn't that mean `discordgo` is the consumer and should defined the interface? I am not the consumer technically, I'm just implementing the required function structure that you declare in your API for dependency injection.

I am not trying to _consume_ `discordgo`, I'm trying to make it so that I can test my handler function that I am providing as a dependency to you.

Really we are 100% on the same page, which is why I think `discordgo` needs to implement this as an interface to allow much easier testing for everyone in the community.

CarsonHoffman:
The way that I’d typically architect this is to have the DiscordGo handlers serve as “shallow” entry points into your actual logic, with the logic being the consumer of the interface with the functionality it needs. This has the added benefit of completely decoupling your logic from the library, as well, as it only depends on the methods in the interface it requires, and a completely different library (or a mock implementation) could feasibly be used as long as it satisfies those interfaces by changing out that shallow entry point.

theckman:
@CarsonHoffman the issue there is I still cannot test the part that actually pulls the stuff I need from `discordgo` to pass to my logic, because of your `*Session` struct. So yes I can test most of my logic, but not the point that integrates with your package. That limitation is not ideal.

I'm not trying to be inflammatory, but instead figure out the path forward. Is my only option to fork this project as you have no desire to support testing those integration points?

PR #670, opened 2019-07-06
Adding ability to redirect discordgo generated logs

TheWithz:
This PR aims to make it easier for developers to redirect log output from discordgo. It uses a map to redirect each log level individually and some helper functions to expose underlying `log` functionality to the developers. Here is an example usage, the code is simply re-purposed from the `myToken` example

>
    file, err := os.OpenFile("debug.log", os.O_APPEND|os.O_WRONLY, 0644)
    if err != nil {
        fmt.Println("error creating log file.", err)
        return
    }

    // Create a new Discord session using the provided login information.
    dg, err := discordgo.New(Email, Password)
    if err != nil {
        fmt.Println("error creating Discord session,", err)
        return
    }

    dg.LogLevel = 2

    discordgo.SetOutFile(discordgo.LogError, file)
    discordgo.SetOutFile(discordgo.LogWarning, file)
    discordgo.SetOutFile(discordgo.LogInformational, file)

    discordgo.SetLogPrefix("[example] ")

    discordgo.SetLogFlags(Ldate|Ltime|Lmicroseconds)

There are two issues that aren't resolved currently. 
- redirecting the logs is very verbose as you need to specify the package name each time. 
  - this isn't really a huge issue, just looks weird
- Not all of the print statements in the library use the logger, for example voice.go
https://github.com/bwmarrin/discordgo/blob/8d8906ce4b959dacabfefbad9bacf11b6ffd8035/voice.go#L95

My go isn't the greatest, but if you have some suggestions on how to solve those two problems I can work on fixing it.

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -12,6 +12,8 @@ package discordgo
 import (
 	"fmt"
 	"log"
+    "io"
+    "os"
 	"runtime"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


Hinara:
Shouldn't that be just another custom logger ? Which can be proposed  but I'm not sure if this should be implemented directly in discordgo

TheWithz:
My goal with this PR was mainly to open dialog about how discordgo handles their inhouse logs. My specific issue was using a CLI with no way to redirect certain debug messages away from stderr. I asked about this in the discord and people I spoke to said there wasn't currently a way to do that. Currently, I just forked the project and deleted the print statements I didn't like, but that's not really in the spirit of open source, so I decided to try and make a contribution.

iopred:
You should be able to do this without this PR just by changing discordgo.Logger to a custom method.

PR #683, opened 2019-08-09
Remove State fallback from Session#Guild

recapitalverb:
This keeps Session#Guild a low level method and help it stay consistent with other methods.


CarsonHoffman:
LGTM; this is something that has been mentioned a fair amount of times on the Discord server but has never been followed through on. This should serve to keep things consistent in the library's REST method semantics. Thanks!

PR #685, opened 2019-08-17
feature: Add support for configurable opus rate/size

otothea:
This PR adds support for setting the opus rate and size when joining a voice channel. To avoid breaking changes, I have created new methods to join a voice channel: `ChannelVoiceJoinV2(opts)` and `ChannelVoiceJoinManualV2(opts)`. The V2 methods accept a single options struct instead of multiple arguments to allow for adding more options in the future without having to introduce or avoid breaking changes.

v1
```
s.ChannelVoiceJoin(cid, gid, false, true)
```
v2
```
s.ChannelVoiceJoinV2(&discordgo.ChannelVoiceJoinOptions{
  ChannelID: cid,
  GuildID: gid,
  Deaf: true,
  Mute: false,
  OpusRate: 48000,
  OpusSize: 960,
})
```

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -42,6 +42,8 @@ type VoiceConnection struct {
 	speaking     bool
 	reconnecting bool // If true, voice connection is trying to reconnect
 
+	opusRate int          // Sample rate for opus, default is 48000
+	opusSize int          // Sample size for opus, default is 960
 	OpusSend chan []byte  // Chan for sending opus audio
 	OpusRecv chan *Packet // Chan for receiving opus audio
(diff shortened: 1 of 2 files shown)


bwmarrin:
Maybe to match the naming we have now, we should go with ChannelVoiceJoinComplex? It's not the best name either but it's better than V2.

Sector95:
I think this PR should be reconsidered.  Having an options struct is really a solid way to make parameters optional and extensible in the future.  V2 makes sense to me.

PR #687, opened 2019-08-25
Add function that checks voice channel for users

haddu:
In discord.js there's a function [member](https://discord.js.org/#/docs/main/stable/class/VoiceChannel?scrollTo=members) that returns and in discord.py there's [is](https://discordpy.readthedocs.io/en/latest/api.html?highlight=voice%20members#discord.VoiceChannel.members), we need something similiar for go, i wrote a small example code that accepts channelID and guildID and returns either empty array or array populated with userIDs, the code itself probably works poorly because i wrote it w/o understanding how exactly it should be implemented, but we still need something similiar.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1378,6 +1378,25 @@ func (s *Session) Channel(channelID string) (st *Channel, err error) {
 	return
 }
 
+// VoiceChannelUsers returns IDS of users present in given channelID
+func (s *Session) VoiceChannelUsers(channelID, guildID string) (st []string, err error) {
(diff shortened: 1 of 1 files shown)


bwmarrin:
This is probably outside the scope of this library.  DiscordGo is intended to be a minimal wrapper around the Discord API itself, without any added magic functions.  If there's not an API endpoint from Discord to provide this lookup, it's not really a good fit.  Maybe an example of how to do it could be added to the wiki if you'd like to contribute that.

PR #691, opened 2019-08-28
Updated Presence struct and members

JurrijnP:
I have updated the presence struct and game struct to the current Discord documentation.

### DIFF
--- a/state.go
+++ b/state.go
@@ -200,9 +200,6 @@ func (s *State) PresenceAdd(guildID string, presence *Presence) error {
 			if presence.Status != "" {
 				guild.Presences[i].Status = presence.Status
 			}
-			if presence.Nick != "" {
-				guild.Presences[i].Nick = presence.Nick
-			}
 
 			//Update the optionally sent user information
 			//ID Is a mandatory field so you should not need to check if it is empty
@@ -921,16 +918,11 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 				// Member not found; this is a user coming online
 				m = &Member{
 					GuildID: t.GuildID,
-					Nick:    t.Nick,
 					User:    t.User,
 					Roles:   t.Roles,
 				}
 
 			} else {
-
-				if t.Nick != "" {
-					m.Nick = t.Nick
-				}
 
 				if t.User.Username != "" {
 					m.User.Username = t.User.Username
--- a/structs.go
+++ b/structs.go
@@ -582,37 +582,76 @@ type VoiceState struct {
 
 // A Presence stores the online, offline, or idle and game status of Guild members.
 type Presence struct {
-	User   *User    `json:"user"`
-	Status Status   `json:"status"`
-	Game   *Game    `json:"game"`
-	Nick   string   `json:"nick"`
-	Roles  []string `json:"roles"`
-	Since  *int     `json:"since"`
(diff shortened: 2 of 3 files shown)


JurrijnP:
Ignore my desperate 4 attempts to fix spacing while forgetting diff tells you what to change.

structs.go
```diff
@@ -582,37 +582,79 @@ type VoiceState struct {
...
 // A Presence stores the online, offline, or idle and game status of Guild members.
 type Presence struct {
-	User   *User    `json:"user"`
-	Status Status   `json:"status"`
-	Game   *Game    `json:"game"`
-	Nick   string   `json:"nick"`
-	Roles  []string `json:"roles"`
-	Since  *int     `json:"since"`
+	// The user the presence updated for.
+	User *User `json:"user"`
```
svenwiltink: These comments don't add any real value

state.go
```diff
@@ -200,9 +200,6 @@ func (s *State) PresenceAdd(guildID string, presence *Presence) error {
 			if presence.Status != "" {
 				guild.Presences[i].Status = presence.Status
 			}
-			if presence.Nick != "" {
-				guild.Presences[i].Nick = presence.Nick
```
svenwiltink: we no longer track nicknames?

structs.go
```diff
@@ -582,37 +582,79 @@ type VoiceState struct {
...
+// ActivityType is the type of "activity" (see ActivityType* consts) in the Activity struct
+type ActivityType int
 
 // Valid GameType values
 const (
-	GameTypeGame GameType = iota
-	GameTypeStreaming
-	GameTypeListening
-	GameTypeWatching
+	ActivityTypeGame ActivityType = iota
```
svenwiltink: This is a breaking change, which we do not allow

structs.go
```diff
@@ -582,37 +582,79 @@ type VoiceState struct {
...
+	ActivityTypeStreaming
+	ActivityTypeListening
 )
 
-// A Game struct holds the name of the "playing .." game for a user
-type Game struct {
```
svenwiltink: The struct is called Game to reflect the same name as the api

state.go
JurrijnP: I logged a PresenceUpdate and noticed the field isn't being provided, same for "Since". Thats why I removed them from the struct and therefore also from the state.

structs.go
JurrijnP: I will change the names back to Game in that case.

structs.go
JurrijnP: As I wasn't aware breaking changes weren't allowed, I changed it to _Activity_ as the json field is still called _game_ but the type is referred to as an _activity object_. But as said earlier will change this back to _Game_.

state.go
svenwiltink: According to the discord api nick changes are only provided in GuildMemberUpdate events. Not in a presence update :thinking:

structs.go
svenwiltink: We don't allow breaking changes to make sure people can always update the lib to the latest version. Unless we release a new major version, in which case breaking changes are allowed

JurrijnP:
Breaking changes have been reversed.

JurrijnP:
Requesting this to be merged.

PR #694, opened 2019-09-11
Fix inconsistent TTS casing

### LINKED ISSUE #640: Inconsistent casing for TTS
`WebhookParams` has it as `TTS`, `Message` and others use `Tts`.
According to https://github.com/golang/go/wiki/CodeReviewComments#initialisms it should be `TTS`.

BigHeadGeorge:
Closes #640

### DIFF
--- a/message.go
+++ b/message.go
@@ -63,7 +63,7 @@ type Message struct {
 	MentionRoles []string `json:"mention_roles"`
 
 	// Whether the message is text-to-speech.
-	Tts bool `json:"tts"`
+	TTS bool `json:"tts"`
 
 	// Whether the message mentions everyone.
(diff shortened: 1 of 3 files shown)


bwmarrin:
@JurrijnP  technically following semver, we can break anything prior to 1.0 :)  With Go modules being a thing, people can lock their app to a specific version and deal with breaking changes as they upgrade. 

I'm in favour of getting this fixed, along with other issues like this that should be corrected.

CarsonHoffman:
LGTM. This is a fairly infrequently-used property, and even for those uses that will broken it's a simple fix. We shouldn't necessarily be afraid to fix small things like this at this stage in library development :)

Issue #401, opened 2017-07-05
Locks should not be public


nhooyr:
e.g. on https://godoc.org/github.com/bwmarrin/discordgo#Session why is the lock public? Why would a user need to access it?

And if there is a lock, why are so many of the fields public? They should have getter methods that use the lock to prevent race conditions.

In many other structures in the library, this is repeated. I think we should make the locks private and use getter/setter methods to prevent race conditions.

ammario:
Getter methods aren't sufficient when multiple operations need to be performed synchronously. Some operations (such as iterating VoiceConnections) isn't supported by any helper function.

iopred:
There's no reason for the locks to be public.

nhooyr:
I disagree, at least for this struct. I don't see any reason why multiple operations would need to be performed synchronously on the fields of a Session. Most of the public fields are configuration options and they should not be public in the first place. There should be a configuration struct or something else (see my link below) that is passed to New to configure the session. That would remove most of the fields on the struct.

As for iterating voice connections, we can just write a method to return a shallow copy of the map. The performance impact will be negligible but the API will be much more safe and clean.

The public lock also raises questions like if its safe to access the http client or state members without locking/unlocking because they are protected by their own lock. I don't see why the state member's lock is public either. Most of its fields are configuration too and they should be in a separate configuration struct or something else.

There is a post by Dave Cheney that goes over various options for configuration: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

ammario:
If the lock is removed, functionality such as iterating `VoiceConnections` should be carefully considered or we risk deserting users relying on it (including me).

bwmarrin:
Hmn, Why? Well probably just because that's how it happened at the start and nobody in the last year or so has ever said anything or objected to it.  But, the issue with making huge changes now is how will that effect the people using the API now and is the benefit worth the pain?  I'm definitely open to discussing this and see what the community as a whole thinks.

nhooyr:
> Hmn, Why? Well probably just because that's how it happened at the start and nobody in the last year or so has ever said anything or objected to it.

Ah sorry, I didn't mean to ask literally "Why?" but rather, I wanted to propose changes. I'll retitle.

> But, the issue with making huge changes now is how will that effect the people using the API now and is the benefit worth the pain?

I think the safety is worth it.

And you have the disclaimer on the README so the community was warned already that this library may have major changes.

nhooyr:
[...]

PR #695, opened 2019-09-13
Private locks on most structs with locks

BigHeadGeorge:
#401 
State still has a public lock for iterating over stuff like guilds and channels because something like putting a lock on Ready seems unnecessary

### DIFF
--- a/discord_test.go
+++ b/discord_test.go
@@ -103,12 +103,12 @@ func TestOpenClose(t *testing.T) {
 	// this is totally gross.
 	start := time.Now()
 	for {
-		d.RLock()
+		d.mutex.RLock()
 		if d.DataReady {
-			d.RUnlock()
+			d.mutex.RUnlock()
 			break
 		}
-		d.RUnlock()
+		d.mutex.RUnlock()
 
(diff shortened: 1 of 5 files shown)


bwmarrin:
While I think moving the mutex locks to private is a good idea - this PR includes some other non-related changes.  Do you think you could clean up the PR to include only changes related to the mutex locks?

BigHeadGeorge:
Yup, lemme clean it up real quick

BigHeadGeorge:
I think that's the only unrelated thing, but without it accessing VoiceConnections won't be safe because of private locks. Maybe that's an issue for another PR though idk

bwmarrin:
@gogitops can you please remove all of your reviews from this repository, they're essentially spam.  Unfortunately github doesn't seem to provide me a way of doing it otherwise I would delete them all.

gogitops:
@bwmarrin deleted comments on all PRs in this org. Bot went a bit wild by accident. Fix has been implemented. Won't happen again.

bwmarrin:
Thank you @gogitops  :) Much appreciated.

PR #705, opened 2019-10-18
tools/eventhandlers: Don't pass Buffer by value

riking:
A copy of a bytes.Buffer will not function like the original. Additionally, it may have looked to the casual reader like log.Fatal was somehow writing to the buffer. Instead, the content is explicitly included at the top of the error, followed by a newline.


riking:
Travis failure is a flake.

CarsonHoffman:
PRs should be submitted against the `develop` branch, not `master`. If you'd still like to see this change, please open a new one against the correct branch.

PR #706, opened 2019-10-18
examples/{airhorn,pingpong}: os.Kill cannot be caught

riking:
os.Kill cannot be caught, only sent, so there's no point in passing it to signal.Notify.


CarsonHoffman:
This change LGTM, but we'd prefer that PRs are opened against the `develop` branch so that we can more easily manage versioning. If you could open another PR on `develop`, I'd be happy to merge!

PR #707, opened 2019-10-24
Add Audit Log Reason support

JurrijnP:
For this to work I had to change the request functions as well. I changed the request functions to support a slice of headers in case Discord decides to add more optional headers in the future.

This PR is only breaking for the `RequestWithLockedBucket` function as it is no longer exported.

Furthermore I have changed all functions that, to my knowledge, cause an audit log entry to support adding a reason as variadic argument.
With this PR #386 can also be closed.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -50,12 +50,13 @@ func (s *Session) Request(method, urlStr string, data interface{}) (response []b
 	return s.RequestWithBucketID(method, urlStr, data, strings.SplitN(urlStr, "?", 2)[0])
 }
 
-// RequestWithBucketID is the same as requestWithBucketID but does not support optional headers.
+// RequestWithBucketID makes a (GET/POST/...) Requests to Discord REST API with JSON data.
 func (s *Session) RequestWithBucketID(method, urlStr string, data interface{}, bucketID string) (response []byte, err error) {
 	return s.requestWithBucketID(method, urlStr, data, bucketID, nil)
 }
 
 // requestWithBucketID makes a (GET/POST/...) Requests to Discord REST API with JSON data.
+// optHeaders allows for optional headers to be set in the request.
 func (s *Session) requestWithBucketID(method, urlStr string, data interface{}, bucketID string, optHeaders []requestHeader) (response []byte, err error) {
 	var body []byte
 	if data != nil {
@@ -78,12 +79,13 @@ func (s *Session) request(method, urlStr, contentType string, b []byte, bucketID
 	return s.requestWithLockedBucket(method, urlStr, contentType, b, s.Ratelimiter.LockBucket(bucketID), sequence, optHeaders)
 }
 
-// RequestWithLockedBucket is the same as requestWithLockedBucket but without optional header support.
+// RequestWithLockedBucket makes a request using a bucket that's already been locked.
 func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b []byte, bucket *Bucket, sequence int) (response []byte, err error) {
 	return s.requestWithLockedBucket(method, urlStr, contentType, b, bucket, sequence, nil)
 }
 
 // requestWithLockedBucket makes a request using a bucket that's already been locked.
+// optHeaders allows for optional headers to be set in the request.
 func (s *Session) requestWithLockedBucket(method, urlStr, contentType string, b []byte, bucket *Bucket, sequence int, optHeaders []requestHeader) (response []byte, err error) {
 	if s.Debug {
 		log.Printf("API REQUEST %8s :: %s\n", method, urlStr)
--- a/restapi.go
+++ b/restapi.go
@@ -42,7 +42,7 @@ var (
 )
 
 type requestHeader struct {
-    key, val string
+	key, val string
 }
 
 // Request is the same as RequestWithBucketID but the bucket id is the same as the urlStr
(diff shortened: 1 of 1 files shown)


restapi.go
```diff
@@ -82,6 +91,12 @@ func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b
 		return
 	}
 
+	if optHeaders != nil {
```
CarsonHoffman: This is redundant. Iterating over a nil slice is a no-op

restapi.go
```diff
@@ -82,6 +91,12 @@ func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b
...
+		for opt := range optHeaders {
```
CarsonHoffman: If we're going to use a range-for loop, perhaps just going straight for the value version rather than the index would be a tad cleaner?

restapi.go
```diff
@@ -56,21 +65,21 @@ func (s *Session) RequestWithBucketID(method, urlStr string, data interface{}, b
...
 		bucketID = strings.SplitN(urlStr, "?", 2)[0]
 	}
-	return s.RequestWithLockedBucket(method, urlStr, contentType, b, s.Ratelimiter.LockBucket(bucketID), sequence)
+	return s.requestWithLockedBucket(method, urlStr, contentType, b, s.Ratelimiter.LockBucket(bucketID), sequence, optHeaders)
 }
 
-// RequestWithLockedBucket makes a request using a bucket that's already been locked
-func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b []byte, bucket *Bucket, sequence int) (response []byte, err error) {
+// requestWithLockedBucket makes a request using a bucket that's already been locked
+func (s *Session) requestWithLockedBucket(method, urlStr, contentType string, b []byte, bucket *Bucket, sequence int, optHeaders []requestHeader) (response []byte, err error) {
```
CarsonHoffman: I don't think we should completely un-export this method, as using other parts of the (exported) rate limiting code depends on this being here. I get the motivation for not exposing the optional headers outside of the library—perhaps take the existing `RequestWithLockedBucket` and have it be essentially a stub that calls a new `requestWithLockedBucket` with no headers?

restapi.go
```diff
@@ -41,13 +41,22 @@ var (
...
+	key, val string
+}
+
 // Request is the same as RequestWithBucketID but the bucket id is the same as the urlStr
 func (s *Session) Request(method, urlStr string, data interface{}) (response []byte, err error) {
 	return s.RequestWithBucketID(method, urlStr, data, strings.SplitN(urlStr, "?", 2)[0])
 }
 
-// RequestWithBucketID makes a (GET/POST/...) Requests to Discord REST API with JSON data.
+// RequestWithBucketID is the same as requestWithBucketID but does not support optional headers.
```
CarsonHoffman: I think it's fine to leave the comment as-is—defining an exported method's behavior in terms of an un-exported one's behavior doesn't work very well for docs.

restapi.go
```diff
@@ -41,13 +41,22 @@ var (
...
 func (s *Session) RequestWithBucketID(method, urlStr string, data interface{}, bucketID string) (response []byte, err error) {
+	return s.requestWithBucketID(method, urlStr, data, bucketID, nil)
+}
+
+// requestWithBucketID makes a (GET/POST/...) Requests to Discord REST API with JSON data.
```
CarsonHoffman: Perhaps add something about the optional headers in this comment (going along with the above review comment)

restapi.go
```diff
@@ -56,21 +65,26 @@ func (s *Session) RequestWithBucketID(method, urlStr string, data interface{}, b
...
+func (s *Session) request(method, urlStr, contentType string, b []byte, bucketID string, sequence int, optHeaders []requestHeader) (response []byte, err error) {
 	if bucketID == "" {
 		bucketID = strings.SplitN(urlStr, "?", 2)[0]
 	}
-	return s.RequestWithLockedBucket(method, urlStr, contentType, b, s.Ratelimiter.LockBucket(bucketID), sequence)
+	return s.requestWithLockedBucket(method, urlStr, contentType, b, s.Ratelimiter.LockBucket(bucketID), sequence, optHeaders)
 }
 
-// RequestWithLockedBucket makes a request using a bucket that's already been locked
+// RequestWithLockedBucket is the same as requestWithLockedBucket but without optional header support.
```
CarsonHoffman: Same thing as above

restapi.go
JurrijnP: Fixed.

JurrijnP:
As I am still waiting on a review, maybe somebody else can review this?

restapi.go
```diff
@@ -56,21 +66,27 @@ func (s *Session) RequestWithBucketID(method, urlStr string, data interface{}, b
...
 
-// RequestWithLockedBucket makes a request using a bucket that's already been locked
+// RequestWithLockedBucket makes a request using a bucket that's already been locked.
 func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b []byte, bucket *Bucket, sequence int) (response []byte, err error) {
+	return s.requestWithLockedBucket(method, urlStr, contentType, b, bucket, sequence, nil)
+}
+
+// requestWithLockedBucket makes a request using a bucket that's already been locked.
+// optHeaders allows for optional headers to be set in the request.
+func (s *Session) requestWithLockedBucket(method, urlStr, contentType string, b []byte, bucket *Bucket, sequence int, optHeaders []requestHeader) (response []byte, err error) {
```
gogitops, line 89: warning: cyclomatic complexity 23 of function (*Session).requestWithLockedBucket() is high (> 15) (gocyclo)

FedorLap2006:
Haven't seen this one, due to the status. But just as a heads up to people who stumble upon this.
Audit log reasons are now implemented through `discordgo.WithReason` option.

PR #708, opened 2019-10-24
Reworked GuildMemberEdit

JurrijnP:
I have reworked the `GuildMemberEdit` function to use a struct to support all optional parameters.
On top of that I have added 2 new functions for (un)muting and (un)deafining members as they use the same struct as `GuildMemberEdit`.

**Note:**
This PR goes together with #700 as the `channel_id` parameter has the same issue if a string would be used. Furthermore this PR would conflict with #707, so for this PR to get merged I would have to change the functions to support adding an audit log reason.
Finally this PR is breaking.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -41,6 +41,22 @@ var (
 	ErrUnauthorized            = errors.New("HTTP request was unauthorized. This could be because the provided token was not a bot token. Please add \"Bot \" to the start of your token. https://discordapp.com/developers/docs/reference#authentication-example-bot-token-authorization-header")
 )
 
+
+// NullString is a custom String type to allow for optional modifying strings that can be reset
+// the JSON Marshalling encoding is unique:
+// if the string is empty, it will be encoded as null
+// if the string is not empty, it will be encoded as a string
+type NullString string
+
+// MarshalJSON returns the JSON encoding of the NullString
+func (s NullString) MarshalJSON() ([]byte, error) {
+	if s == "" {
+		return json.Marshal(nil)
+	}
+
+	return json.Marshal(string(s))
+}
+
 // Request is the same as RequestWithBucketID but the bucket id is the same as the urlStr
 func (s *Session) Request(method, urlStr string, data interface{}) (response []byte, err error) {
 	return s.RequestWithBucketID(method, urlStr, data, strings.SplitN(urlStr, "?", 2)[0])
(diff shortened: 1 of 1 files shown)


restapi.go
```diff
@@ -826,16 +841,20 @@ func (s *Session) GuildMemberDeleteWithReason(guildID, userID, reason string) (e
...
-// guildID  : The ID of a Guild.
-// userID   : The ID of a User.
-// roles    : A list of role ID's to set on the member.
-func (s *Session) GuildMemberEdit(guildID, userID string, roles []string) (err error) {
-
-	data := struct {
-		Roles []string `json:"roles"`
-	}{roles}
+// GuildMemberEditData holds all optional parameters for the GuildMemberEdit function.
+type GuildMemberEditData struct {
```
CarsonHoffman: The convention in the library for these types of structs is simply `GuildMemberEdit` (see `ChannelEdit`, `MessageEdit`, etc.)

restapi.go
```diff
@@ -851,10 +870,38 @@ func (s *Session) GuildMemberEdit(guildID, userID string, roles []string) (err e
...
+	ns := NullString(channelID)
+	data := GuildMemberEditData{ChannelID: &ns}
 
-	data := struct {
-		ChannelID string `json:"channel_id"`
-	}{channelID}
+	_, err = s.RequestWithBucketID("PATCH", EndpointGuildMember(guildID, userID), data, EndpointGuildMember(guildID, ""))
+	if err != nil {
+		return
+	}
```
CarsonHoffman: This error check is redundant. I know there are some of these in the REST code currently, but it'd be better to not add any more. :)

restapi.go
```diff
@@ -851,10 +870,38 @@ func (s *Session) GuildMemberEdit(guildID, userID string, roles []string) (err e
...
+//  guildID : The ID of a Guild.
+//  userID  : The ID of a User.
+//  mute    : Whether the user gets muted.
+func (s *Session) GuildMemberMute(guildID, userID string, mute bool) (err error) {
+	data := GuildMemberEditData{Mute: mute}
+
+	_, err = s.RequestWithBucketID("PATCH", EndpointGuildMember(guildID, userID), data, EndpointGuildMember(guildID, ""))
+	if err != nil {
+		return
+	}
```
CarsonHoffman: Again redundant

PR #709, opened 2019-10-25
Support for Batch Request Guild Members

mistahj67:
This is supported by the Discord API and will save requests.

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -421,7 +421,7 @@ func (s *Session) RequestGuildMembers(guildID string, query string, limit int) (
 		Limit:    limit,
 	}
 	err = s.requestGuildMembers(data)
-	return err
+	return
 }
 
 // RequestGuildMembersBatch requests guild members from the gateway
@@ -436,7 +436,7 @@ func (s *Session) RequestGuildMembersBatch(guildIDs []string, query string, limi
 		Limit:    limit,
 	}
 	err = s.requestGuildMembers(data)
-	return err
+	return
 }
 
 func (s *Session) requestGuildMembers(data requestGuildMembersData) (err error) {
@@ -452,7 +452,7 @@ func (s *Session) requestGuildMembers(data requestGuildMembersData) (err error) 
 	err = s.wsConn.WriteJSON(requestGuildMembersOp{8, data})
 	s.wsMutex.Unlock()
 
-	return err
+	return
 }
 
 // onEvent is the "event handler" for all messages received on the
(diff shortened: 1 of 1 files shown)


Bios-Marcel:
This is backwards compatibility breaking change. I'd suggest adding a new function.

mistahj67:
Updated to eliminate breaking issue

wsapi.go
```diff
@@ -399,9 +399,10 @@ func (s *Session) UpdateStatusComplex(usd UpdateStatusData) (err error) {
...
 
 type requestGuildMembersData struct {
-	GuildID []string `json:"guild_id"`
-	Query   string   `json:"query"`
-	Limit   int      `json:"limit"`
+	GuildID  string   `json:"guild_id"`
+	GuildIDs []string `json:"guild_id"`
+	Query    string   `json:"query"`
+	Limit    int      `json:"limit"`
 }
```
JurrijnP: You can leave the struct the same as before so:
```go
type requestGuildMembersData struct {
	GuildID []string `json:"guild_id"`
	Query   string   `json:"query"`
	Limit   int      `json:"limit"`
}
```

And then pass the argument in the original function to the slice.

wsapi.go
```diff
@@ -411,10 +411,35 @@ type requestGuildMembersOp struct {
...
+	data := requestGuildMembersData{
+		GuildIDs: guildIDs,
+		Query:    query,
+		Limit:    limit,
+	}
+	err = s.requestGuildMember(data)
+	return err
+}
+
+func (s *Session) requestGuildMember(data requestGuildMembersData) (err error) {
```
CarsonHoffman: `requestGuildMember` seems to be a typo here; `requestGuildMembers` would make more sense.

wsapi.go
```diff
@@ -423,17 +448,11 @@ func (s *Session) RequestGuildMembers(guildID, query string, limit int) (err err
...
-		Query:   query,
-		Limit:   limit,
-	}
-
 	s.wsMutex.Lock()
 	err = s.wsConn.WriteJSON(requestGuildMembersOp{8, data})
 	s.wsMutex.Unlock()
 
-	return
+	return err
```
CarsonHoffman: This is unnecessary as `err` is assigned on line 452.

mistahj67: It's required to return err as the function returns an error?

CarsonHoffman: Since `err` is a named return value, it's redundant. See: [named return values](https://tour.golang.org/basics/7)

mistahj67: TIL. Thanks.

PR #710, opened 2019-10-25
Support for Reasons in Actions

mistahj67:
Discord audit logs display reason for actions if provided in the header of the request.


JurrijnP:
I refer you to #707 that adds reason support to all actions that cause an audit log entry.

CarsonHoffman:
Looks like this functionality is being introduced in #707, so I'm going to close this.

PR #711, opened 2019-11-02
Add Before Update Support for More Events.

mistahj67:


CarsonHoffman:
I like the additional support for this functionality, however after some discussion in the Discord channel I think it might be more beneficial to further generalize this system rather than hard-coding it into each event that might benefit from it. I'll leave this open for now in the event that that system becomes infeasible to implement given the current state of the library (I'm not sure if it will be possible without tearing up the event handling code).

Issue #716, opened 2019-12-19
Discord API responds with 400 Bad Request


ewohltman:
Hello,

I opened an issue with Discord's API documentation about getting `400: Bad Request` responses from the API when attempting to create a new guild role: https://github.com/discordapp/discord-api-docs/issues/1267

The response I got was that if the `Content-Type` header of the request is set but the request body is empty, the API will respond with the `400: Bad Request`.

ewohltman:
With the PR merged (thanks @CarsonHoffman !), any chance we might see a merge to master and new tag/release in the near-ish future?

I'd like to keep my project's `go.mod` file clean with a nice `v0.20.2` or something instead of pinning against the develop branch :smile: 

Thanks very much for the cooperation!

ewohltman:
Another quick PR there for the develop branch to prep for a merge to master if that helps. Please feel free to decline it if it's not appropriate :smile:

PR #717, opened 2019-12-19
Fix for Discord's API returning a 400 Bad Request from empty request body

ewohltman:
Fix for Discord's API returning a `400: Bad Request` from empty request body when `Content-Type` header is set.

This PR is associated with issue https://github.com/bwmarrin/discordgo/issues/716

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -88,7 +88,12 @@ func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b 
 		req.Header.Set("authorization", s.Token)
 	}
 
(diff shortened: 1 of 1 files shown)


CarsonHoffman:
Thanks for the PR. After our initial conversation in the Discord server the other day I had come away with the idea that it would be a bit more sinister to fix due to how `application/json` is hard-coded in `RequestWithBucketID`, but now I honestly can't think of how I got that idea in my head. Sorry for the inconvenience, and thanks for the quick fix; LGTM!

PR #718, opened 2019-12-19
Prep develop branch to merge to master

ewohltman:
This PR is related to issue https://github.com/bwmarrin/discordgo/issues/716


CarsonHoffman:
I'm going to close this as we'll need to do a cherry-pick to get the fix into master (we're not looking to do a full merge from develop to master at this moment) and it would be cleaner to just keep everything in one commit, but thanks for pushing us to getting it into master quickly (and thanks again for the PR yesterday!) :)

PR #721, opened 2019-12-25
Add SendWSEvent

diamondburned:


bwmarrin:
What are some examples of WS events you would like to be able to send?

CarsonHoffman:
This PR definitely requires some more explanation, including what documented gateway payloads the library does not currently support. That said, given that the PR is opened against `master` rather than `develop`, it's unmergeable in any state, so I'm going to close it. If you'd still like to see this functionality, open a new PR against `develop`, and please provide some more description of why this would be useful for the library.

PR #723, opened 2019-12-31
Add guild folders

diamondburned:


structs.go
```diff
@@ -694,13 +694,22 @@ type Settings struct {
...
 	FriendSourceFlags      *FriendSourceFlags `json:"friend_source_flags"`
 	Status                 Status             `json:"status"`
 	DetectPlatformAccounts bool               `json:"detect_platform_accounts"`
 	DeveloperMode          bool               `json:"developer_mode"`
 }
 
+// GuildFolder holds a single folder that you see in the left guild panel.
+type GuildFolder struct {
+	Name     string   `json:"name"`
+	ID       int64    `json:"id"`
```
ewohltman: In other structs, the `ID` field is a string. Should this also be a string instead of an int64?

Bios-Marcel:
Is there no eventing or anything for this? What happens if a guild folder gets deleted or changed in another instance? Will the cache be inconsistent?

structs.go
diamondburned: This isn't a snowflake, it's rather just arbitrary numbers (e.g. `3185939939`)

diamondburned:
> Is there no eventing or anything for this? What happens if a guild folder gets deleted or changed in another instance? Will the cache be inconsistent?

Fair point, I'll add those

Issue #678, opened 2019-07-29
Unable to specify guild_subscriptions in Identify


bkuhl:
When identifying with the gateway, there's an option in the API to specify `guild_subscriptions` which disables receiving certain events.  It'd be fantastic if we had the ability to specify this in the `Identify` struct: https://github.com/bwmarrin/discordgo/blob/8d8906ce4b959dacabfefbad9bacf11b6ffd8035/wsapi.go#L735

bwmarrin:
I think maybe we could export an Identify struct that exposes all of these settings as some of them are exposed in other fashions now, and yet some others might be nice to be configurable.  We could default the values on New(), but allow you to customize it before calling Open()

bwmarrin:
@bkuhl I've created a proposed solution, please feel free to test it out and provide your feedback.

bkuhl:
Sorry I meant to follow up with this.  Thanks so much for tackling it.

PR #728, opened 2020-01-20
Expose Identify Packet

bwmarrin:
This exposes the Identify Packet, and adds missing Identify parameters.  This is a proposed solution to #678

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -925,22 +925,25 @@ type GatewayStatusUpdate struct {
 	AFK    bool     `json:"afk"`
 }
 
+// https://discordapp.com/developers/docs/topics/gateway#activity-object
+// Activity defines the Activity sent with GatewayStatusUpdate
 type Activity struct {
 	Name string
 	Type ActivityType
 	URL  string
 }
 
-// GameType is the type of "game" (see GameType* consts) in the Game struct
+// ActivityType is the type of Activity (see ActivityType* consts) in the Activity struct
 // https://discordapp.com/developers/docs/topics/gateway#activity-object-activity-types
 type ActivityType int
 
-// Valid GameType values
+// Valid ActivityType values
+//https://discordapp.com/developers/docs/topics/gateway#activity-object-activity-types
 const (
 	ActivityTypeGame GameType = iota
 	ActivityTypeStreaming
 	ActivityTypeListening
-	//	ActivityTypeWatching // not valid in the use?
+	//	ActivityTypeWatching // not valid in this use case?
 	ActivityTypeCustom = 4
 )
 
--- a/structs.go
+++ b/structs.go
@@ -29,8 +29,10 @@ type Session struct {
 	// General configurable settings.
 
 	// Authentication token for this session
+	// TODO: Remove Below, Deprecated, Use Identify struct
 	Token string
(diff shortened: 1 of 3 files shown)


wsapi.go
```diff
@@ -741,55 +740,42 @@ func (s *Session) onVoiceServerUpdate(st *VoiceServerUpdate) {
...
 type identifyOp struct {
-	Op   int          `json:"op"`
-	Data identifyData `json:"d"`
+	Op   int      `json:"op"`
+	Data Identify `json:"d"`
 }
 
 // identify sends the identify packet to the gateway
 func (s *Session) identify() error {
+	s.log(LogDebug, "called")
```
ewohltman: Is this debug line intended to be left in? If so, does the log message contain enough information for other developers to know it came from here?

structs.go
```diff
@@ -909,6 +916,59 @@ type GatewayBotResponse struct {
...
+type ActivityType int
+
+// Valid ActivityType values
+//https://discordapp.com/developers/docs/topics/gateway#activity-object-activity-types
+const (
+	ActivityTypeGame GameType = iota
+	ActivityTypeStreaming
+	ActivityTypeListening
+	//	ActivityTypeWatching // not valid in this use case?
+	ActivityTypeCustom = 4
```
ewohltman: According to the documentation you provided, it does seem that `ActivityTypeWatching` is not a supported activity type. Would it be worth moving this line to above the commented out line to continue the `iota` without hard-coding it to `4`?

structs.go
```diff
@@ -909,6 +916,59 @@ type GatewayBotResponse struct {
...
+	Type ActivityType
+	URL  string
+}
+
+// ActivityType is the type of Activity (see ActivityType* consts) in the Activity struct
+// https://discordapp.com/developers/docs/topics/gateway#activity-object-activity-types
+type ActivityType int
+
+// Valid ActivityType values
+//https://discordapp.com/developers/docs/topics/gateway#activity-object-activity-types
```
ewohltman: [Nit] This line could use a leading space to align nicely

wsapi.go
bwmarrin: Yeah, most the functions in the library have a similar one.  When that prints out (only with debug logging enabled) it includes the line number, filename, and function name.  It's just to help follow the code path when troubleshooting an issue.

structs.go
bwmarrin: Well, ActivityTypeCustom has to be a 4 (per Discord Docs) - so it has to be hard coded in there,  we can't just continue the iota because that line would get assigned a 3.  

The comment could possibly be removed but I liked it just for a bit of clarity.

structs.go
bwmarrin: I agree :)

structs.go
ewohltman: Ah good eye! I missed that 3 is skipped over in their doc. Probably a hint that will be supported in the future :)

bwmarrin:
We could definitely use more unit tests - most of the library has no test coverage.  In large part because it's somewhat difficult to test many of it offline.  Thanks a lot for helping review @ewohltman - great having a second set of eyes on changes.

PR #732, opened 2020-01-30
Adding millisecond precision to SnowflakeTimestamp utility

AndyTempel:
The current utility only allows second precision and maybe to some people millisecond precision will come in handy since Discord timestamps are in milliseconds.

### DIFF
--- a/util.go
+++ b/util.go
@@ -12,6 +12,6 @@ func SnowflakeTimestamp(ID string) (t time.Time, err error) {
 		return
 	}
 	timestamp := (i >> 22) + 1420070400000
-	t = time.Unix(timestamp/1000, (timestamp%1000) * 1000000)
+	t = time.Unix(timestamp/1000, (timestamp%1000)*1000000)
 	return
 }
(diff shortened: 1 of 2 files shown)


CarsonHoffman:
Instead of introducing the modulus operation and passing both parameters to `time.Unix`, it might just be easier to pass in `timestamp*1000000` to the nanoseconds field and `0` to the seconds field. This is the pattern you'll often see with this function—keeping the parameters mutually exclusive—and takes a bit of load off reading the function (as well as performance with the modulus, as subtle as it may be).

AndyTempel:
Yup, you're correct. I was going by (*outdated*) [benchmarks](https://github.com/Tigraine/go-timemilli#performance-considerations) and didn't test it my self if they still stand.

ewohltman:
Could this PR be changed to merge into the `develop` branch of this repo instead of `master` as per the [Contributing guidelines](https://github.com/bwmarrin/discordgo#contributing)?

AndyTempel:
Branch changed.

CarsonHoffman:
I went ahead and added a quick unit test since we were changing how the function worked (which probably should've been there from the start). Thanks for the PR!

PR #735, opened 2020-02-06
Add banner to GuildParams struct

samuel-hunter:
I've tested this on a server with Banner permissions, and it works as perfect as I can tell.


samuel-hunter:
I've made my own fork of this library with the change added so I can continue developing on my discord bot, and the operation has been working flawlessly for the past 3 weeks!

samuel-hunter:
Right, I missed that. It looks like the `develop` branch already has the needed change!

samuel-hunter:
Hi,
Checking back on this a few months later I noticed that the `develop` branch is missing, the feature doesn't look like it's added _yet_, and the Contributing guidelines points to the `master` branch, so I think this pull request is worth another look at!

I've rebased my fork and got no merge conflicts. It looks like everything should be OK!

CarsonHoffman:
LGTM; thanks for the PR! I'm sorry for the confusion—when ewohltman made their comment they were correct, but in the meantime we consolidated the codebase into a single branch to make things easier for contributors. We did perform a final merge of `develop` into `master` before making that change, however, so if it isn't there now I'm not sure it was actually ever present in `develop`. Either way, thanks!

PR #739, opened 2020-02-27
Change all to syscall signals

sschiz:
os.Interrupt is equal to syscall.SIGINT i. e. there's repeat

proof

### DIFF
--- a/examples/airhorn/main.go
+++ b/examples/airhorn/main.go
@@ -62,7 +62,7 @@ func main() {
 	// Wait here until CTRL-C or other term signal is received.
 	fmt.Println("Airhorn is now running.  Press CTRL-C to exit.")
 	sc := make(chan os.Signal, 1)
-	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM, os.Interrupt, os.Kill)
+	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL)
 	<-sc
 
 	// Cleanly close down the Discord session.
--- a/examples/airhorn/main.go
+++ b/examples/airhorn/main.go
@@ -62,7 +62,7 @@ func main() {
 	// Wait here until CTRL-C or other term signal is received.
 	fmt.Println("Airhorn is now running.  Press CTRL-C to exit.")
 	sc := make(chan os.Signal, 1)
-	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL)
+	signal.Notify(sc, syscall.SIGTERM, os.Interrupt, os.Kill)
 	<-sc
 
 	// Cleanly close down the Discord session.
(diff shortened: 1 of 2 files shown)


examples/airhorn/main.go
```diff
@@ -62,7 +62,7 @@ func main() {
 	// Wait here until CTRL-C or other term signal is received.
 	fmt.Println("Airhorn is now running.  Press CTRL-C to exit.")
 	sc := make(chan os.Signal, 1)
-	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM, os.Interrupt, os.Kill)
+	signal.Notify(sc, syscall.SIGTERM, os.Interrupt, os.Kill)
```
ewohltman: Since `syscall.SIGTERM` is being left in, would it be more consistent to instead change `os.Interrupt` to `syscall.SIGINT` and `os.Kill` to `syscall.SIGKILL`?

```suggestion
	signal.Notify(sc, syscall.SIGTERM, syscall.SIGINT, syscall.SIGKILL)
```

examples/pingpong/main.go
```diff
@@ -43,7 +43,7 @@ func main() {
 	// Wait here until CTRL-C or other term signal is received.
 	fmt.Println("Bot is now running.  Press CTRL-C to exit.")
 	sc := make(chan os.Signal, 1)
-	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM, os.Interrupt, os.Kill)
+	signal.Notify(sc, syscall.SIGTERM, os.Interrupt, os.Kill)
```
ewohltman: Since `syscall.SIGTERM` is being left in, would it be more consistent to instead change `os.Interrupt` to `syscall.SIGINT` and `os.Kill` to `syscall.SIGKILL`?

```suggestion
	signal.Notify(sc, syscall.SIGTERM, syscall.SIGINT, syscall.SIGKILL)
```

examples/airhorn/main.go
sschiz: Sigkill doesn’t work

sschiz: So I decided to use OS signal

ewohltman: Makes sense about SIGKILL, I should have looked closer because technically no process is able to act on SIGKILL anyway.

> The signals SIGKILL and SIGSTOP may not be caught by a program, and therefore cannot be affected by this package.
https://golang.org/pkg/os/signal/#hdr-Types_of_signals

This would mean that we should remove the SIGKILL from the list of signals passed to `signal.Notify` to be most technically correct. Though either way, these changes are for examples so IMO its not super critical that this be completely consistent.

bwmarrin:
It seems like it would make the most sense to just convert this to remove the os.X and use syscall.X identifies.  Also remove the ones that cannot be caught.

PR #741, opened 2020-02-27
String method for message

sschiz:
I have added  String method to Message struct so that I could get a message link

### DIFF
--- a/message.go
+++ b/message.go
@@ -304,6 +304,11 @@ type MessageReference struct {
 	GuildID   string `json:"guild_id"`
 }
 
+// String returns link to message
+func (m *Message) String() string {
+	return "https://discordapp.com/channels/" + m.GuildID + "/" + m.ChannelID + "/" + m.ID
+}
+
 // ContentWithMentionsReplaced will replace all @<id> mentions with the
 // username of the mention.
 func (m *Message) ContentWithMentionsReplaced() (content string) {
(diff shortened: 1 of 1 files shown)


message.go
```diff
@@ -304,6 +304,11 @@ type MessageReference struct {
 	GuildID   string `json:"guild_id"`
 }
 
+// String returns link to message
+func (m *Message) String() string {
+	return "https://discordapp.com/channels/" + m.GuildID + "/" + m.ChannelID + "/" + m.ID
```
ewohltman: If I were passing a `*Message` into something that accepts a `fmt.Stringer`, I think I would expect that it at least contains the content of the message instead of a URL.

Something like the following?

```suggestion
        return fmt.Sprintf("Message in %s (%s): %s", m.GuildID, m.ChannelID, m.Content)
```

sschiz: I think URL is better, ‘cause it can be used, for example, for logger bot

sschiz: even if you use it as a debug, what these numbers will give, so the URL will be more useful

bwmarrin:
I don't think this type of function is within scope for this library. The goals of the library is to provide (and only provide) a minimal wrapper around the Discord API itself.

JurrijnP:
If this does get implemented I would certainly not call it `String()` if it returns the URL. But I would not implement this anyway, does not fit the scope in my opinion.

sschiz:
@bwmarrin why? useful func, why not?

PR #744, opened 2020-03-03
Add support for null values

JurrijnP:
First, thanks to @Seklfreak for creating the initial PR (#700).

I created `NullInt` & `NullBool` in addition to `NullString`. This PR only contains the types and corresponding funcitons, no modified functions as it would be a massive undertaking. So feedback would be greatly appreciated.

### DIFF
--- a/types.go
+++ b/types.go
@@ -39,6 +39,25 @@ func (s NullString) MarshalJSON() ([]byte, error) {
 	return json.Marshal(string(s))
 }
 
+func nullString(v interface{}) *NullString {
+	switch v.(type) {
+	case string:
+		if v.(string) == "" {
+			return nil
+		} else {
+			ns := NullString(v.(string))
+			return &ns
+		}
+	case NullString:
+		ns := NullString(v.(NullString))
+		return &ns
+	case *NullString:
+		return v.(*NullString)
+	default:
+		return nil
+	}
+}
+
 // NullBool is a custom Bool type to allow for optional modifying booleans that can have false as value
 // the JSON Marshalling encoding is unique:
 // if the bool is false, it will be encoded as false
@@ -54,6 +73,25 @@ func (b NullBool) MarshalJSON() ([]byte, error) {
 	return json.Marshal(true)
 }
 
+func nullBool(v interface{}) *NullBool {
+	switch v.(type) {
+	case bool:
+		if v.(bool) == false {
+			return nil
+		} else {
+			nb := NullBool(true)
+			return &nb
+		}
+	case NullBool:
+		nb := NullBool(v.(NullBool))
+		return &nb
+	case *NullBool:
+		return v.(*NullBool)
+	default:
+		return nil
+	}
+}
+
 // NullInt is a custom Int type to allow for optional modifying integers that can have 0 as value
 // the JSON Marshalling encoding is unique:
 // if the integer is 0, it will be encoded as 0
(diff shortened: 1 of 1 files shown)


types.go
```diff
@@ -24,6 +24,108 @@ func (t Timestamp) Parse() (time.Time, error) {
...
+		return &ni
+	case NullInt:
+		ni := NullInt(v.(NullInt))
+		return &ni
+	case *NullInt:
+		return v.(*NullInt)
+	default:
+		return nil
+	}
+}
```
ewohltman: `switch` for type can create a variable of the case type automatically: https://tour.golang.org/methods/16

```suggestion
func nullInt(i interface{}) *NullInt {
	switch v := i.(type) {
	case int:
		if v == 0 {
			return nil
		}

		ni := NullInt(v)
		return &ni
	case NullInt:
		return &v
	case *NullInt:
		return v
	default:
		return nil
	}
}
```

types.go
```diff
@@ -24,6 +24,108 @@ func (t Timestamp) Parse() (time.Time, error) {
...
+		return &nb
+	case NullBool:
+		nb := NullBool(v.(NullBool))
+		return &nb
+	case *NullBool:
+		return v.(*NullBool)
+	default:
+		return nil
+	}
+}
```
ewohltman: `switch` for type can create a variable of the case type automatically: https://tour.golang.org/methods/16

```suggestion
func nullBool(i interface{}) *NullBool {
	switch v := i.(type) {
	case bool:
		if v {
			nb := NullBool(true)
			return &nb
		}

		return nil
	case NullBool:
		return &v
	case *NullBool:
		return v
	default:
		return nil
	}
}
```

types.go
```diff
@@ -24,6 +24,108 @@ func (t Timestamp) Parse() (time.Time, error) {
...
+		return &ns
+	case NullString:
+		ns := NullString(v.(NullString))
+		return &ns
+	case *NullString:
+		return v.(*NullString)
+	default:
+		return nil
+	}
+}
```
ewohltman: `switch` for type can create a variable of the case type automatically: https://tour.golang.org/methods/16

```suggestion
func nullString(i interface{}) *NullString {
	switch v := i.(type) {
	case string:
		if v == "" {
			return nil
		}

		ns := NullString(v)
		return &ns
	case NullString:
		return &v
	case *NullString:
		return v
	default:
		return nil
	}
}
```

types.go
```diff
@@ -24,6 +24,108 @@ func (t Timestamp) Parse() (time.Time, error) {
...
+type NullInt int
+
+// MarshalJSON returns the JSON encoding of the NullBool
+func (i NullInt) MarshalJSON() ([]byte, error) {
+	if i == 0 {
+		return json.Marshal(0)
+	}
+
+	return json.Marshal(int(i))
+}
```
ewohltman: If `i == 0`, should we marshal `nil`? If not, I'm not sure how this would operate differently from using a regular `int`.

```suggestion
func (i NullInt) MarshalJSON() ([]byte, error) {
	if i == 0 {
		return json.Marshal(nil)
	}

	return json.Marshal(i)
}
```

types.go
```diff
@@ -24,6 +24,108 @@ func (t Timestamp) Parse() (time.Time, error) {
...
+type NullBool bool
+
+// MarshalJSON returns the JSON encoding of the NullBool
+func (b NullBool) MarshalJSON() ([]byte, error) {
+	if b == false {
+		return json.Marshal(false)
+	}
+
+	return json.Marshal(true)
+}
```
ewohltman: Can this be simplified to the following?

```suggestion
func (b NullBool) MarshalJSON() ([]byte, error) {
	return json.Marshal(b)
}
```

Or should this be the following where if `b` is false, we marshal `nil`? If not, I'm not sure I understand how this would operate differently from a regular `bool`.

```suggestion
func (b NullBool) MarshalJSON() ([]byte, error) {
	if b {
		return json.Marshal(b)
	}

	return json.Marshal(nil)
}
```

Satty9361:
There are cases above where you are forced to marshal zero values as null. Although this might work out in this context , if one is to implement a more generic method, this is a very good nullable datatype implementation in golang: <https://github.com/volatiletech/null>.

PR #756, opened 2020-04-01
Add bitwise permission const PermissionViewChannel

ewohltman:
Based upon bitwise permissions defined at: https://discordapp.com/developers/docs/topics/permissions#permissions-bitwise-permission-flags

Note: The new `PermissionViewChannel` is technically the same offset as the existing `PermissionReadMessages` permission. However, the Discord documentation doesn't have a specific "READ_MESSAGES" permission, instead using "VIEW_CHANNEL" to govern visibility of both text and voice channels.

To maintain backward compatibility without breaking existing users, `PermissionReadMessages` has been left in because it's an exported const. We may want to somehow mark it as deprecated though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# IDE-specific metadata
+.idea/
--- a/structs.go
+++ b/structs.go
@@ -973,6 +973,7 @@ type IdentifyProperties struct {
 
 // Constants for the different bit offsets of text channel permissions
 const (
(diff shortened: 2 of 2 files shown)


bwmarrin:
How about you add a comment to the old one with the brief explanation and say it's deprecated? Otherwise, in a few months, we'll be wondering why we have two :)

ewohltman:
> Can you add a comment to the old one with the brief explanation and say it's deprecated? Otherwise, in a few months, we'll be wondering why we have two :)

Done!

Issue #754, opened 2020-03-29
No CA Certs results in a `null` discordgo response


centerorbit:
My application built and ran fine locally on both windows and linux, but once I put it into a container, the only response I got from Discord was `null`. Took me forever to track it down, but it turns out that discordgo doesn't error out or log anything if it can't establish the TLS, and instead just returns `null`.

Expected behavior:
An error is thrown or a log output that the API request failed due to TLS connection issues.

The fix was to copy the ca-certificates from another image (in this case buster) to the scratch image:
```
FROM golang:1.13-buster as go

COPY . /go/src/api
WORKDIR /go/src/api
ENV CGO_ENABLED=0
ENV GOOS=linux 
RUN go build -ldflags="-w -s" .


FROM scratch
COPY --from=go /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
COPY --from=go /go/src/api/api /bin/api
ENTRYPOINT ["api"]
```

bwmarrin:
Looking here,

https://github.com/bwmarrin/discordgo/blob/develop/wsapi.go#L82

For some reason that's set at a Warning Level so your logging level would need to be increased to see it.  I'm not sure why it's a warning instead of an error though.

If you do ever have weird problem, raising your log level to debug will net you a lot more logging details to help track down the issue.

CarsonHoffman:
The log level was changed to `LogError` in #760, which should make this error more obvious in the future.

PR #760, opened 2020-04-07
Change log level for websocket connection failure from warning to error

ewohltman:
This PR is in support of Issue https://github.com/bwmarrin/discordgo/issues/754.


CarsonHoffman:
Looking at our other uses of `LogError`, this definitely seemed like an appropriate change. Thanks for the PR!

Issue #742, opened 2020-02-29
Adding support for Gateway Intents, possibly resulting in majorly breaking code


JurrijnP:
Discord introduced a new way to limit your event data called *Gateway Intents*. You can read more about it here:
https://discordapp.com/developers/docs/topics/gateway#gateway-intents

You need to provide the gateway intents in the identify payload, which is not exposed in any way in DiscordGo. The gateway intents are optional in gateway version 6, but mandatory in version 7. The easy route would be to add another field in the `Session` struct to define the intents but that reveals an underlying issue which has been brought up before. Ideally you would be able to directly access the indentify function with the data you want to provide or something in that direction. @diamondburned & I discussed a bit about this issue and came the conclusion it would be best to make this change the breaking change that not only changes the identify function, but the whole structure of connecting all together that has been up for debate before.

I am mainly talking about the `New()` & `Open()` function.
As this would break the code of every single user of DiscordGo in the end I would like @bwmarrin to comment on this. :+1:

PR #763, opened 2020-04-18
Basic support for Gateway Intents

lukasz-horonziak:
Issue #742 

How to use?

`s.Identify.Intents =  discordgo.IntentsGuilds | discordgo.IntentsGuildMembers`

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -74,6 +74,7 @@ func New(args ...interface{}) (s *Session, err error) {
 	s.Identify.GuildSubscriptions = true
 	s.Identify.Properties.OS = runtime.GOOS
 	s.Identify.Properties.Browser = "DiscordGo v" + VERSION
+	s.Identify.Intents = MakeIntent(IntentsAllWithoutPrivileged)
 
 	// If no arguments are passed return the empty Session interface.
 	if args == nil {
--- a/discord.go
+++ b/discord.go
@@ -74,7 +74,6 @@ func New(args ...interface{}) (s *Session, err error) {
 	s.Identify.GuildSubscriptions = true
 	s.Identify.Properties.OS = runtime.GOOS
 	s.Identify.Properties.Browser = "DiscordGo v" + VERSION
-	s.Identify.Intents = IntentsAllWithoutPrivileged
 
 	// If no arguments are passed return the empty Session interface.
 	if args == nil {
--- a/discord.go
+++ b/discord.go
@@ -74,7 +74,7 @@ func New(args ...interface{}) (s *Session, err error) {
 	s.Identify.GuildSubscriptions = true
 	s.Identify.Properties.OS = runtime.GOOS
 	s.Identify.Properties.Browser = "DiscordGo v" + VERSION
-	s.Identify.Intents = 32767
+	s.Identify.Intents = IntentsAllWithoutPrivileged
 
 	// If no arguments are passed return the empty Session interface.
 	if args == nil {
--- a/discord.go
+++ b/discord.go
@@ -74,6 +74,7 @@ func New(args ...interface{}) (s *Session, err error) {
 	s.Identify.GuildSubscriptions = true
 	s.Identify.Properties.OS = runtime.GOOS
 	s.Identify.Properties.Browser = "DiscordGo v" + VERSION
+	s.Identify.Intents = 32767
 
 	// If no arguments are passed return the empty Session interface.
 	if args == nil {
--- a/structs.go
+++ b/structs.go
@@ -1139,5 +1139,5 @@ const (
 	IntentsAll = IntentsAllWithoutPrivileged |
 		IntentsGuildMembers |
 		IntentsGuildPresences
-    IntentsNone Intent = 0
+	IntentsNone Intent = 0
 )
--- a/structs.go
+++ b/structs.go
@@ -1103,3 +1103,38 @@ const (
 
 	ErrCodeReactionBlocked = 90001
 )
+
+const (
+	IntentsGuilds = 1 << iota
+	IntentsGuildMembers
+	IntentsGuildBans
+	IntentsGuildEmojis
+	IntentsGuildIntegrations
+	IntentsGuildWebhooks
+	IntentsGuildInvites
+	IntentsGuildVoiceStates
+	IntentsGuildPresences
+	IntentsGuildMessages
+	IntentsGuildMessageReactions
+	IntentsGuildMessageTyping
+	IntentsDirectMessagesMessages
+	IntentsDirectMessageReaction
+	IntentsDirectMessageTyping
+
+	IntentsAllWithoutPrivileged = IntentsGuilds |
+		IntentsGuildBans |
+		IntentsGuildEmojis |
+		IntentsGuildIntegrations |
+		IntentsGuildWebhooks |
+		IntentsGuildInvites |
+		IntentsGuildVoiceStates |
+		IntentsGuildMessages |
+		IntentsGuildMessageReactions |
+		IntentsGuildMessageTyping |
+		IntentsDirectMessagesMessages |
+		IntentsDirectMessageReaction |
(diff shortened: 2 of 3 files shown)


discord.go
```diff
@@ -74,6 +74,7 @@ func New(args ...interface{}) (s *Session, err error) {
 	s.Identify.GuildSubscriptions = true
 	s.Identify.Properties.OS = runtime.GOOS
 	s.Identify.Properties.Browser = "DiscordGo v" + VERSION
+	s.Identify.Intents = 32767
```
ewohltman, line 77: This value `32767` seems a bit like a magic number. How was it obtained? Can consts be used to define/build up to it? Do we also want it to be the default value rather than zero?

lukasz-horonziak: We can't use 0 cause no event will be received from discord - it will basicly 'broke' old implementation. 

32767 means all intents so it's eq to `1 << 0 | 1 << 1 | 1 << 2 | 1 << 3 | 1 << 4 | 1 << 5 | 1 << 6 | 1 << 7 | 1 << 8 | 1 << 9 | 1 << 10 | 1 << 11 | 1 << 12 | 1 << 13 | 1 << 14`

https://discordapp.com/developers/docs/topics/gateway#list-of-intents

I can move it to const as well.

structs.go
```diff
@@ -1102,3 +1103,38 @@ const (
...
+	IntentsGuildEmojis
+	IntentsGuildIntegrations
+	IntentsGuildWebhooks
+	IntentsGuildInvites
+	IntentsGuildVoiceStates
+	IntentsGuildPresences
+	IntentsGuildMessages
+	IntentsGuildMessageReactions
+	IntentsGuildMessageTyping
+	IntentsDirectMessagesMessages
```
CarsonHoffman, line 1120: This seems like an unintended stutter; the Discord docs list it as simply `DIRECT_MESSAGES`.

structs.go
```diff
@@ -1102,3 +1103,38 @@ const (
...
+	IntentsDirectMessageReaction
```
CarsonHoffman, line 1121: I don't mean to nit-pick too much, but this should probably be plural to match both the docs and `IntentsGuildMessageReactions`.

lukasz-horonziak: Great catch!

structs.go
lukasz-horonziak: Fixed

lukasz-horonziak:
@CarsonHoffman Is there anything else I should fix?

bwmarrin:
I think this looks pretty good.  I wouldn't mind if we could have a couple users just test out the PR and make sure it's working as intended ;)

CarsonHoffman:
My only concern would be that this is a breaking change for functions requiring the members and presences intents; because intents aren't required on v6, this change would force users into the intents system immediately even if they don't want to use it (and would thus require them to enable the intents in their portal and on their session). This becomes less of a concern as we get closer to intents becoming mandatory, but it's still something that I'm slightly unsure about.

Perhaps `Intents` could be marked as `omitempty`, so that a user could manually opt-out of the intents system? I think that would be an acceptable middle ground if enabling it by default. I'm not sure that specifying `0` for the intents field has any practical use.

lukasz-horonziak:
@bwmarrin I'm using it for a while on my bot and it works fine but I'm looking forward for more testing :-)

@CarsonHoffman as always great comments. I make intents `omitempty`. The 0 value is a valid option for intent so I make it available for users as well.

structs.go
```diff
@@ -959,7 +959,7 @@ type Identify struct {
 	Shard              *[2]int             `json:"shard,omitempty"`
 	Presence           GatewayStatusUpdate `json:"presence,omitempty"`
 	GuildSubscriptions bool                `json:"guild_subscriptions"`
-	Intents            int                 `json:"intents"`
+	Intents            *Intent             `json:"intents,omitempty"`
```
CarsonHoffman, line 962: I think leaving the field as an `int` as fine. `0` falls under the `omitempty` clause for ints; introducing it as a pointer makes doing initialization a bit more difficult.

discord.go
```diff
@@ -74,7 +74,6 @@ func New(args ...interface{}) (s *Session, err error) {
 	s.Identify.GuildSubscriptions = true
 	s.Identify.Properties.OS = runtime.GOOS
 	s.Identify.Properties.Browser = "DiscordGo v" + VERSION
-	s.Identify.Intents = IntentsAllWithoutPrivileged
```
CarsonHoffman: I think having this as a default is fine; it's just that having *no* option to opt-out concerned me.

structs.go
```diff
@@ -1102,3 +1103,41 @@ const (
...
+		IntentsGuildMessages |
+		IntentsGuildMessageReactions |
+		IntentsGuildMessageTyping |
+		IntentsDirectMessages |
+		IntentsDirectMessageReactions |
+		IntentsDirectMessageTyping
+	IntentsAll = IntentsAllWithoutPrivileged |
+		IntentsGuildMembers |
+		IntentsGuildPresences
+    IntentsNone Intent = 0
```
CarsonHoffman, line 1142: Again—this might be a bit of a nitpick, but make sure your code gets run through `gofmt` before committing :)

structs.go
riking: There are a few events that are still sent for a `0` intent, and `omitempty` on a pointer still allows a 0 to be sent over the wire.

Instead, add a utility function that heap-allocates the integer:

```
func MakeIntent(intents Intent) *Intent {
       return &intents
}
```

riking:
Excited to get this merged soon - my bot is having a bit of a time with overloaded network pipes!

structs.go
iopred: This seems fine for now, I like the idea of the Helper method.

iopred:
Could you gofmt and then I'll merge, thanks!

discord.go
lukasz-horonziak: I like that too but maybe we should set it when Intent will be required. I just thought to add it when it will be required by discord.

structs.go
lukasz-horonziak: Fixed. I will remember to use it :)

lukasz-horonziak:
> Excited to get this merged soon - my bot is having a bit of a time with overloaded network pipes!

You could always use `replace` in ur go mod and test it on ur bot.

structs.go
lukasz-horonziak: Ye, the 0 is valid intent thats why I used pointer. I added helper method to utils.

util.go
```diff
@@ -15,3 +15,7 @@ func SnowflakeTimestamp(ID string) (t time.Time, err error) {
 	t = time.Unix(timestamp/1000, 0)
 	return
 }
+
+func MakeIntent(intents Intent) *Intent {
```
riking, line 19: Put this right next to the enum definition instead.

Doc:

// MakeIntent helps convert a gateway intent value for use in the Identify structure.

lukasz-horonziak:
@riking @iopred @CarsonHoffman Is there anything more I should fix?

CarsonHoffman:
It looks like the CI build is failing on golint:

```
$ golint -set_exit_status ./...
8262structs.go:1107:6: exported type Intent should have comment or be unexported
8263structs.go:1110:2: exported const IntentsGuilds should have comment (or a comment on this block) or be unexported
8264Found 2 lint suggestions; failing.
8265The command "golint -set_exit_status ./..." exited with 1.
```

lukasz-horonziak:
> It looks like the CI build is failing on golint:
> 
> ```
> $ golint -set_exit_status ./...
> 8262structs.go:1107:6: exported type Intent should have comment or be unexported
> 8263structs.go:1110:2: exported const IntentsGuilds should have comment (or a comment on this block) or be unexported
> 8264Found 2 lint suggestions; failing.
> 8265The command "golint -set_exit_status ./..." exited with 1.
> ```

fixed

PR #764, opened 2020-04-19
Add allowed mentions functionality

CarsonHoffman:
A recent API addition allows developers to explicitly list the mentions that should be valid in the message, allowing bots to safely send user input as a message without worrying about escaping the message (which never really worked, due to all of the internal processing Discord does on messages). This PR adds this functionality, allowing the `MessageAllowedMentions` struct to be used on complex message sends and edits, giving the developer the choice to explicitly disallow all mentions as a base and then selectively whitelist some. Existing functionality doesn't change—when this struct isn't provided, all mention types are still allowed.

Due to the complex nature of how the fields are processed on the backend I didn't attempt to explain every use-case in comments and linked to the Discord docs for that, but I believe I covered the basic uses of the struct and its fields.

### DIFF
--- a/message.go
+++ b/message.go
@@ -129,10 +129,11 @@ type File struct {
 
 // MessageSend stores all parameters you can send with ChannelMessageSendComplex.
 type MessageSend struct {
-	Content string        `json:"content,omitempty"`
-	Embed   *MessageEmbed `json:"embed,omitempty"`
-	TTS     bool          `json:"tts"`
-	Files   []*File       `json:"-"`
+	Content         string                  `json:"content,omitempty"`
+	Embed           *MessageEmbed           `json:"embed,omitempty"`
+	TTS             bool                    `json:"tts"`
+	Files           []*File                 `json:"-"`
+	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 
 	// TODO: Remove this when compatibility is not required.
 	File *File `json:"-"`
@@ -141,8 +142,9 @@ type MessageSend struct {
 // MessageEdit is used to chain parameters via ChannelMessageEditComplex, which
 // is also where you should get the instance from.
 type MessageEdit struct {
-	Content *string       `json:"content,omitempty"`
-	Embed   *MessageEmbed `json:"embed,omitempty"`
+	Content         *string                 `json:"content,omitempty"`
+	Embed           *MessageEmbed           `json:"embed,omitempty"`
+	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 
 	ID      string
 	Channel string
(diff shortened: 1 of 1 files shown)


message.go
```diff
@@ -171,6 +173,42 @@ func (m *MessageEdit) SetEmbed(embed *MessageEmbed) *MessageEdit {
...
+// MessageAllowedMentions allows the user to specify which mentions
+// Discord is allowed to parse in this message. This is useful when
+// sending user input as a message, as it prevents unwanted mentions.
+// If this type is used, all mentions must be explicitly whitelisted,
+// either by putting an AllowedMentionType in the Parse slice
+// (allowing all mentions of that type) or, in the case of roles and
+// users, explicitly allowing those mentions on an ID-by-ID basis.
+// For more information on this functionality, see:
+// https://discordapp.com/developers/docs/resources/channel#allowed-mentions-object-allowed-mentions-reference
+type MessageAllowedMentions struct {
```
ewohltman, line 196: If one were to try to allow mentions in a message for a particular role (likewise for particular users), would this struct look like the following?

```go
messageAllowedMentionsCase1 := &MessageAllowedMentions{
    Roles: []string{"someRoleID"},
}
```

If so, then this comment is a little confusing to me because it seems to imply that no mentions would be allowed:

```go
	// The mention types that are allowed to be parsed in this message.
	// Please note that this is purposely **not** marked as omitempty,
	// so if a zero-value MessageAllowedMentions object is provided no
	// mentions will be allowed.
	Parse []AllowedMentionType `json:"parse"`
```

If instead the struct should look like:

```go
messageAllowedMentionsCase2 := &MessageAllowedMentions{
    Parse: []AllowedMentionType{AllowedMentionTypeRoles},
    Roles: []string{"someRoleID"},
}
```

Then this comment becomes a bit confusing because it states we should not use `AllowedMentionTypeRoles` in the `Parse` slice:

```go
	// A list of role IDs to allow. This cannot be used when specifying
	// AllowedMentionTypeRoles in the Parse slice.
	Roles []string `json:"roles,omitempty"`
```

Based on the Discord docs, this is a bit convoluted but the former case (`messageAllowedMentionsCase1`) appears to be the intention. They also have an example when the `parse` field is not present in the resultant JSON:

```json
{
    "content": "<@123> Time for some memes.",
    "allowed_mentions": {
        "users": ["123", "125"]
    }
}
```

Perhaps we do want to allow `omitempty` for `Parse` because if it is not present, it seems to satisfy the intention of disallowing all mention types except for the roles and users that were specifically whitelisted.

CarsonHoffman: The intention behind the original comment was to warn the user about the case of a zero-value **`MessageAllowedMentions`** object, not necessarily a zero-value slice; other fields can still be specified.

I do not believe `Parse` can be left as `omit empty`, as then there is no way to disallow everyone and role mentions without specifying a whitelisted user or role; `"allowed_mentions": {}` doesn't disallow them, while `"allowed_mentions": {"parse": []}` does.

tiehm: Yes, leaving `Parse` to `omitempty` can't really be done. There needs to be a good way of disabling all mentions as per the Discord API. If we want to get this as close as possible to the Discord API we can't do much but just warn users that an empty `allowed_mentions` does something different than an empty `allowed_mentions.parse`.

PR #766, opened 2020-04-26
Fix typo in comment of Ping Pong example

ZusorCode:

PR #767, opened 2020-04-30
Fix typo in comment of PingPong example

ZusorCode:
This is #766 again but due to changing the target branch things got messed up


iopred:
Thanks!

PR #774, opened 2020-05-09
Introduce GitHub action

lukasz-horonziak:
Cause travis is not working for unknown reason I moved CI to Github action.


.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Install golint
+        run: |
+          go get -u golang.org/x/lint/golint
+      
+      - name: Checkout 
+        uses: actions/checkout@v2
+
+      - name: Init
+        shell: bash
+        run: go get -v ./...
```
ewohltman, line 29: Would this be better served with `go mod download` to ensure we pull dependencies that are defined in `go.mod`, instead of pulling the latest version of dependencies and building with that without reflecting it back into `go.mod`?

```suggestion
        run: go mod download
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Lint
+        shell: bash
+        run: |
+          diff <(gofmt -d .) <(echo -n)
+          go vet -x ./...
+          golint -set_exit_status ./...
+
+      - name: Test
+        shell: bash
+        run: go test -v -race ./...
```
ewohltman, line 40: [Nit] Adding a newline to end of document:

```suggestion
        run: go test -v -race ./... 

```

ewohltman:
I've become a big fan of GitHub Actions myself :+1:

lukasz-horonziak:
@ewohltman Fixed. I don't know why test are failing both on travis and my local machine. Do you know if it's a problem on Travis?

ewohltman:
Hmm, I don't see CI builds in Travis for this PR for some reason: https://travis-ci.org/github/bwmarrin/discordgo/pull_requests

That said, the tests will pretty much fail for everyone developing locally because they are written to require secrets (e.g. account credentials). For example when I run `go test -v -race ./...` locally with my fork:

```
=== RUN   TestVoiceICE
    TestVoiceICE: restapi_test.go:188: VoiceICE() returned error: HTTP 401 Unauthorized, {"message": "401: Unauthorized", "code": 0}
--- FAIL: TestVoiceICE (0.05s)
```

Could you point to the failed CI build(s) for this PR? Maybe I'm just missing them.

I can also say, I used to use Travis for all my CI. Sometimes I would see weird issues like this, which is why I was motivated to move all my CI to GitHub Actions :smile:

ewohltman:
@bwmarrin @CarsonHoffman @iopred 

As a quick FYI, an option will have to be set for the repo to allow this to run if it isn't already set:

https://github.community/t/workflow-error-action-not-allowed/16788/3

> I had to change the “Actions permissions” configuration to " Enable local and third-party Actions for this repository" to fix it.

I ran into this recently with hooking up linting via GitHub actions for https://github.com/OpenDiablo2/OpenDiablo2.

That option will allow the usage of remote actions such as:

```yaml
      - name: Setup go
        uses: actions/setup-go@v2
        with:
          go-version: ${{ matrix.go }}
```
-- https://github.com/actions/setup-go

```yaml
      - name: Checkout 
        uses: actions/checkout@v2
```
-- https://github.com/actions/checkout

ewohltman:
Additionally, the required secrets would need to be moved into the repo's config for the tests.

An example of passing GitHub secrets as environment variables in actions can be found here:
* https://github.com/ewohltman/ephemeral-roles/blob/master/.github/workflows/pullRequest.yml
* https://github.com/ewohltman/ephemeral-roles/blob/master/.github/workflows/pushToMaster.yml

PR #787, opened 2020-06-28
Add team data to applications

Nekiono:
This adds data about the Team owning an application (if there is a team owning the application), as well as its members


zkp12:
This should be added

CarsonHoffman:
LGTM! Thanks for the PR!

PR #789, opened 2020-06-28
Update GuildMemberMove to support removing from all voice channel

DylanMeador:


CarsonHoffman:
Thanks for the PR! As full disclosure, this change might end up getting overwritten by changes to how we handle nullable values throughout the library before the next release, but this should be a good enough stop-gap solution to put onto `develop` in the mean time.

PR #790, opened 2020-07-06
chore: add missing fields in User struct

moul:
based on https://discord.com/developers/docs/resources/user

### DIFF
--- a/user.go
+++ b/user.go
@@ -62,6 +62,17 @@ type User struct {
 	// This is a combination of bit masks; the presence of a certain flag can
 	// be checked by performing a bitwise AND between this int and the flag.
 	PublicFlags UserFlags `json:"public_flags"`
+
+	// The type of Nitro subscription on a user's account.
+	// Only available when the request is authorized via a Bearer token.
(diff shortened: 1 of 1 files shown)


moul:
:wave: sorry I didn't see that my PR was red on CI, I just fixed everything and force-pushed :)

CarsonHoffman:
Sorry for the wait—with a change like this the biggest concern is the usability of the interface. It would be great to have an indication in the comments of `PremiumType` and `Flags` that these fields are only available when the request is authorized via a Bearer token, otherwise LGTM!

adityaxdiwakar:
If this is still being worked on or is being updated (cc @moul), #801 has been merged adding some changes to `users.go`, some of which overlapped with the proposed changes from this PR as well, so this would need to be updated, just as an FYI.

moul:
@CarsonHoffman thanks, comments added
@adityaxdiwakar thanks, rebased

CarsonHoffman:
Sorry for missing the update on this. Thank you for the additions!

PR #796, opened 2020-07-27
Update Emoji endpoints to point to the CDN.

iopred:


CarsonHoffman:
It's a bit interesting that this was in this state so long; it looks like it isn't actually used in the library anywhere so it slipped by. 😅

iopred:
It used to work, but I think discord switched things up!

Issue #750, opened 2020-03-15
How Do You Add A Custom Emoji As A Reaction?


ryanhughesdev:
I have a custom emoji on my server called "tree2." I want to add it as a reaction to a message, but when I try I get the following error: 
`HTTP 400 Bad Request, {"emoji_id": ["Value \"688276502470590490>\" is not snowflake."]}`
I have tried all of the following:
`MessageReactionAdd(channelID, messageID, "<:tree2:688276502470590490>")`
`MessageReactionAdd(channelID, messageID, ":tree2:")`
`MessageReactionAdd(channelID, messageID, "688276502470590490")`
`MessageReactionAdd(channelID, messageID, "tree2")`
And probably others that I'm forgetting. None of them work. MessageReactionAdd simply takes an "emojiID" string but there's no indication of what this string should look like and getting the emoji then using emoji.ID, which would be the most obvious thing, doesn't work as it just gives me the suffix number.

What is the correct format then?

ryanhughesdev:
Well, I figured it out minutes after posting this. So it goes.
`MessageReactionAdd(channelID, messageID, "tree2:688276502470590490")`
I'll leave this here for anyone else who is as confused as I was. :)

PR #798, opened 2020-08-05
Add example of emojiID to MessageReactionAdd

ceriath:
Since i ran into the same issue today, the documentation for MessageReactionAdd should be improved, thanks to #750

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2100,7 +2100,7 @@ func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *Webho
 // MessageReactionAdd creates an emoji reaction to a message.
 // channelID : The channel ID.
 // messageID : The message ID.
-// emojiID   : Either the unicode emoji for the reaction, or a guild emoji identifier.
+// emojiID   : Either the unicode emoji for the reaction, or a guild emoji identifier in name:id format (e.g. "hello:1234567654321")
 func (s *Session) MessageReactionAdd(channelID, messageID, emojiID string) error {
 
 	// emoji such as  #⃣ need to have # escaped


CarsonHoffman:
I definitely think that the documentation could be improved here, and though an example is surely helpful, I think actually defining the format would be helpful as well—it might not be immediately clear that `hello:1234567654321` refers to `name:id`.

adityaxdiwakar:
I might be wrong here but doesn't the format also require a colon, such as `:name:id`?
image

CarsonHoffman:
The leading colon is necessary when specifying an emoji in the content of a message; it is not necessary when specifying an emoji in an API context. (I believe it won't fail as it will be considered part of the name, as the API seems to parse `name:id` from right-to-left starting from the ID and considering everything before the right-most colon as part of the name)

restapi.go
```diff
@@ -2100,7 +2100,7 @@ func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *Webho
 // MessageReactionAdd creates an emoji reaction to a message.
 // channelID : The channel ID.
 // messageID : The message ID.
-// emojiID   : Either the unicode emoji for the reaction, or a guild emoji identifier.
+// emojiID   : Either the unicode emoji for the reaction, or a guild emoji identifier (e.g. "hello:1234567654321").
```
FedorLap2006, line 2103: As Carson already pointed out, I think adding something like
```suggestion
// emojiID   : Either the unicode emoji for the reaction, or a guild emoji identifier in name:id format (e.g. "hello:1234567654321")
```
would be really nice

FedorLap2006:
Sorry response took a long time.

ceriath:
I added the suggestion

FedorLap2006:
Thanks for your contribution!

PR #800, opened 2020-08-05
Add support for news channels

### LINKED ISSUE #799: Add Support for /followers and /crosspost endpoints
As per https://github.com/discord/discord-api-docs/issues/1510, the `/followers` and `/crosspost` endpoints for news channels (now available to all servers, rather than just Verified/Partnered/Dev) is allowed to be accessed by bots. These endpoints are yet to be documented but the issue is labeled "documentation-needed" and there is an open PR for it as well, https://github.com/discord/discord-api-docs/pull/1692.

With that, it would be cool to see support for these endpoints within Discordgo, I would love to add these features myself, but want to open it for conversation beforehand.

Thoughts? Thanks in advance!

adityaxdiwakar:
Resolves #799 

Adds support for the following endpoints:
- `@/channels/:id/messages/:id/crosspost`: publishes a message in a news channel to followers
- `@/channels/:id/followers`: follows a news channel

### DIFF
--- a/endpoints.go
+++ b/endpoints.go
@@ -112,6 +112,8 @@ var (
 	EndpointChannelMessagesBulkDelete = func(cID string) string { return EndpointChannel(cID) + "/messages/bulk-delete" }
 	EndpointChannelMessagesPins       = func(cID string) string { return EndpointChannel(cID) + "/pins" }
 	EndpointChannelMessagePin         = func(cID, mID string) string { return EndpointChannel(cID) + "/pins/" + mID }
+	EndpointChannelMessageCrosspost   = func(cID, mID string) string { return EndpointChannel(cID) + "/messages/" + mID + "/crosspost" }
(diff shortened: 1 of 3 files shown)


CarsonHoffman:
It doesn't appear that there's any way to extract the webhook ID from the `ChannelNewsFollow` method if I'm understanding the docs PR correctly; we'd definitely want to support getting back any available information on any new requests we add.

adityaxdiwakar:
My bad, I've implemented that now in eb70c17733f3795f1cc40cd9b8db6c774deb20c0.

CarsonHoffman:
It also seems that the error on `s.RequestWithBucketID` in `ChannelNewsFollow` isn't checked; this leads to `nil` errors being incorrectly being returned by the function as long as the body is empty or valid JSON, so errors like 401, 403 or 404 result in `nil` errors.

adityaxdiwakar:
Not sure how I missed that, thanks, resolved!

CarsonHoffman:
Awesome, the methods look great now. Thank you for the PR!

PR #801, opened 2020-08-08
add user public flags field and public flag consts

zekroTJA:
In this pull request, I've added the `PublicFlags` field to the `User` struct as documented here:  
https://discord.com/developers/docs/resources/user#user-object-user-structure

This field holds publicly avaliable flags like some profile badges, for example the "Hype Squad" badges, the "Partner" badge or the "Verified Bot Developer" badge.

Also, I've added the flag values as constants to perform bitwise operations on user flags in an convenient way. The flag values are defined as documented here:  
https://discord.com/developers/docs/resources/user#user-object-user-flags

I've tried to adapt the code documentation from other flag and type descriptions as consistent as possible. If there is any issue with my implementation, please contact me and I will correct the issue. :)

### DIFF
--- a/user.go
+++ b/user.go
@@ -2,6 +2,27 @@ package discordgo
 
 import "strings"
 
+// UserFlags is the flags of "user" (see UserFlags* consts)
+// https://discord.com/developers/docs/resources/user#user-object-user-flags
+type UserFlags int
+
+// Valid UserFlags values
+const (
+	UserFlagDiscordEmployee      UserFlags = 1 << iota
+	UserFlagDiscordPartner                 = 1 << iota
+	UserFlagHypeSquadEvents                = 1 << iota
+	UserFlagBugHunterLevel1                = 1 << iota
+	UserFlagHouseBravery                   = 1 << iota
+	UserFlagHouseBrilliance                = 1 << iota
+	UserFlagHouseBalance                   = 1 << iota
+	UserFlagEarlySupporter                 = 1 << iota
+	UserFlagTeamUser                       = 1 << iota
+	UserFlagSystem                         = 1 << iota
+	UserFlagBugHunterLevel2                = 1 << iota
+	UserFlagVerifiedBot                    = 1 << iota
+	UserFlagVerifiedBotDeveloper           = 1 << iota
+)
+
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -2,6 +2,27 @@ package discordgo
...
 import "strings"
 
+// UserFlags is the flags of "user" (see UserFlags* consts)
+// https://discord.com/developers/docs/resources/user#user-object-user-flags
+type UserFlags int
+
+// Valid UserFlags values
+const (
+	UserFlagDiscordEmployee      UserFlags = 1 << iota
+	UserFlagDiscordPartner                 = 1 << iota
```
CarsonHoffman, line 12: The `= 1 << iota` bit is only necessary on the first line; because the flags are each explicitly specified in the docs I’d be fine if each line has the value, but if you’d rather just use `iota` then only the first is necessary.

user.go
```diff
@@ -36,6 +57,11 @@ type User struct {
 
 	// Whether the user is a bot.
 	Bot bool `json:"bot"`
+
+	// The public flags on a users account.
```
CarsonHoffman, line 61: Super small nit, but if we’re already fixing something else: should be “user’s” :)

user.go
zekroTJA: Oh well, actually did not know that the bitshitf operation with `itoa` is also adopted to the other values.

CarsonHoffman:
It appears that the flags don't match up with the actual values; notice that there are gaps between "bug hunter level 1" (`1 << 3`) and "house Bravery" (`1 << 6`), and more gaps later (e.g. shifts of 4, 5, 11, 13, 15 are skipped). It would probably better to individually list each value as I mentioned as an option in the original review; I didn't realize there were gaps in the values originally, which is why I suggested fixing `iota`, but now that I know that it'd much more explicit to just list each one, and would allow for easier additions in the future if those values get filled in without fighting with the positional logic of `iota`.

zekroTJA:
Oh lord, I really did not notice that there are gaps between the bit flags. :D Now, it is explicitly defined should be fine. ^^

CarsonHoffman:
Awesome, looks good now. Thanks for the PR!

PR #804, opened 2020-08-12
Add GuildMemberEditComplex for Guild Member Mute/Deaf Edit Support

adityaxdiwakar:
Currently, as far as I know, there is no support for muting or deafening a user (when they are in the voice channel). I've added support for that in this PR.

As a note, `GuildMemberNickname` is still using the old code because of the `/guilds/{guild.id}/members/@me/nick` endpoint. 

On top of that, the `GuildMemberMove` command is also not using `GuildMemberEditComplex` because Discord requires the JSON request value for the `channel_id` key when removing a user to be `null` (if you want to disconnect the user), in Go, this is represented as `nil` and would be omitted by the default JSON marshaller. As a result, if `GuildMemberEditComplex` also attempted to support `ChannelID` modifications, then every change would disconnect the user unintentionally, so I omitted that functionality.

This is a non-breaking change.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -784,6 +784,20 @@ func (m *Member) Mention() string {
 	return "<@!" + m.User.ID + ">"
 }
 
+// A MemberEdit holds Member Field data for a member edit.
+type MemberEdit struct {
+	Nick  string   `json:"nick,omitempty"`
+	Roles []string `json:"roles,omitempty"`
+
+	// Cannot change below fields if user is not in voice channel
+	Mute bool `json:"mute,omitempty"`
+	Deaf bool `json:"deaf,omitempty"`
+
+	// ChannelID is ommitted from this because Discord API requires a nil value
+	// which cannot work with omitempty, since nil value would be omitted from
+	// the request, use GuildMemberMove instead
+}
+
 // A Settings stores data for a specific users Discord client settings.
 type Settings struct {
 	RenderEmbeds           bool               `json:"render_embeds"`
--- a/structs.go
+++ b/structs.go
@@ -784,20 +784,6 @@ func (m *Member) Mention() string {
 	return "<@!" + m.User.ID + ">"
 }
 
-// A MemberEdit holds Member Field data for a member edit.
-type MemberEdit struct {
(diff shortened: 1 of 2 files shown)


structs.go
```diff
@@ -784,6 +784,20 @@ func (m *Member) Mention() string {
...
 }
 
+// A MemberEdit holds Member Field data for a member edit.
+type MemberEdit struct {
+	Nick  string   `json:"nick,omitempty"`
+	Roles []string `json:"roles,omitempty"`
+
+	// Cannot change below fields if user is not in voice channel
+	Mute bool `json:"mute,omitempty"`
+	Deaf bool `json:"deaf,omitempty"`
```
diamondburned, line 794: There's no way to unmute people with `omitempty` present, as a `false` will be treated as an empty.

structs.go
```diff
@@ -784,6 +784,20 @@ func (m *Member) Mention() string {
 	return "<@!" + m.User.ID + ">"
 }
 
+// A MemberEdit holds Member Field data for a member edit.
+type MemberEdit struct {
+	Nick  string   `json:"nick,omitempty"`
+	Roles []string `json:"roles,omitempty"`
```
diamondburned, line 790: You may need `*[]string` here. Refer to https://play.golang.org/p/m-coc5VIfPv.

structs.go
adityaxdiwakar: Oy vey! You're right. :disappointed: Thanks.

structs.go
adityaxdiwakar: :disappointed: Right, of course - no way to remove all roles from someone w/ `omitempty`, argh!

adityaxdiwakar:
@diamondburned Mind reviewing again? Thank you.

diamondburned:
LGTM.

CarsonHoffman:
I think this is a good solution—there had been some concern in the past about supporting all the parameters on this endpoint as they've added quite a few over the years, but I think with the relatively niche nature of the feature having separate methods works well. Thanks for the PR!

PR #805, opened 2020-08-12
Add query escape to GuildEmojiCreate and update documentation

zekroTJA:
As described in the Discord API documentation of the `Create Reaction` endpoint, the `emoji` parameter must be URL encoded:  
https://discord.com/developers/docs/resources/channel#create-reaction

So, I've added URL query escaping for the `emoji` parameter of the `MessageReactionAdd` function. This should also cover the current implementation to escape `#` characters in emoji strings.  
https://github.com/bwmarrin/discordgo/blob/master/restapi.go#L2144

Sadly, the Discord documentation does not specify how the emoji identifier is composed. This took me some time to figure out by asking some other developers and taking a look into the docs of discord.js, where the emoji identifier is shown in an example. To clarify the usage of the emoji identifier, I've added the structure of it to the methods documentation.

Also, I found another potential ambiguity in the documentation of the `GuildEmojiCreate` function. To clarify that the `image` string must be in the base64 data URI scheme, following to the Discord documentation, I've added this to the parameters documentation of the function.  
https://discord.com/developers/docs/reference#image-data

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2137,11 +2137,10 @@ func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *Webho
 // MessageReactionAdd creates an emoji reaction to a message.
 // channelID : The channel ID.
 // messageID : The message ID.
-// emojiID   : Either the unicode emoji for the reaction, or a guild emoji identifier.
+// emojiID   : Either the unicode emoji for the reaction, or a guild emoji identifier (":emojiName:emojiID").
 func (s *Session) MessageReactionAdd(channelID, messageID, emojiID string) error {
 
-	// emoji such as  #⃣ need to have # escaped
-	emojiID = strings.Replace(emojiID, "#", "%23", -1)
+	emojiID = url.QueryEscape(emojiID)
 	_, err := s.RequestWithBucketID("PUT", EndpointMessageReaction(channelID, messageID, emojiID, "@me"), nil, EndpointMessageReaction(channelID, "", "", ""))
 
 	return err
--- a/restapi.go
+++ b/restapi.go
@@ -1325,7 +1325,7 @@ func (s *Session) GuildEmojis(guildID string) (emoji []*Emoji, err error) {
 // GuildEmojiCreate creates a new emoji
 // guildID : The ID of a Guild.
 // name    : The Name of the Emoji.
-// image   : The base64 encoded emoji image, has to be smaller than 256KB.
+// image   : The base64 encoded emoji image, has to be smaller than 256KB and provided in the data URI scheme.
 // roles   : The roles for which this emoji will be whitelisted, can be nil.
 func (s *Session) GuildEmojiCreate(guildID, name, image string, roles []string) (emoji *Emoji, err error) {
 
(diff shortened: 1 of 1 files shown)


CarsonHoffman:
Could you provide an example of an emoji that doesn't currently work? I believe that `net/http` should be automatically escaping the URL for us already.

iopred:
Additionally, updating the docs to use Emoji.ApiName() may be helpful

zekroTJA:
Actually, every emote needs to be URL encoded fowllowing to the documentation, when I interpret this right. And also the `:` characters in the path using the emoji identifier (`:emojiName:emojiID`) needs to be escaped. 

Well, I did not found something in the `net/http` code which URL encodes the parameter before request. To verify this, I've run the following little program (`go.1.14 windows amd64`):
```go
package main

import (
	"net/http"
	"net/url"
)

func main() {
	http.Get("https://zekro.de/:test:")
	http.Get("https://zekro.de/" + url.QueryEscape(":test:"))
}
```

The result in my nginx logs show the following:
```
84.62.139.203 - - [12/Aug/2020:16:54:42 +0200] "GET /%3Atest%3A HTTP/1.1" 404 153 "-" "Go-http-client/1.1",
84.62.139.203 - - [12/Aug/2020:16:54:42 +0200] "GET /:test: HTTP/1.1" 404 153 "-" "Go-http-client/1.1"
```

Apperently, URLs are not automatically escaped.

But if you want to, I can revert this change to the state before, of course. :)

zekroTJA:
Actually, every emote needs to be URL encoded fowllowing to the documentation, when I interpret this right. And also the `:` characters in the path using the emoji identifier (`:emojiName:emojiID`) needs to be escaped. 

Well, I did not found something in the `net/http` code which URL encodes the parameter before request. To verify this, I've run the following little program (`go.1.14 windows amd64`):
```go
package main

import (
	"net/http"
	"net/url"
)

func main() {
	http.Get("https://zekro.de/:test:")
	http.Get("https://zekro.de/" + url.QueryEscape(":test:"))
}
```

The result in my nginx logs show the following:
```
*** - - [12/Aug/2020:16:54:42 +0200] "GET /%3Atest%3A HTTP/1.1" 404 153 "-" "Go-http-client/1.1",
*** - - [12/Aug/2020:16:54:42 +0200] "GET /:test: HTTP/1.1" 404 153 "-" "Go-http-client/1.1"
```

Apperently, URLs are not automatically escaped.

But if you want to, I can revert this change to the state before, of course. :)

CarsonHoffman:
`:` isn't a character that needs to be escaped. If we take the following example:

```go
req, _ := http.NewRequest("GET", `http://test.com/test space😳emoji`, nil)
b, _ := httputil.DumpRequest(req, false)
fmt.Println(string(b))
```

We see: 
```
GET /test%20space%F0%9F%98%B3emoji HTTP/1.1
Host: test.com
```

Which we would expect. I'd politely ask again: could you provide a *specific* example of an emoji that doesn't work? I think there is something that needs to be changed here, but finding specific examples (for example perhaps: emoji that contain path separators or something like that) would help us narrow down the fix.

zekroTJA:
Ah okay, that's interesting. Well, I just read the Discord documentation and escaped the emote ID before it goes into the MessageReactionAdd function to fulfil that requirement and was wondered why this is not done by the wrapper already, but I did not know that the client does this already.

I had no specific emotes which were not working with the current implementation, but I also did not test every single one existent. ^^
Sorry, if you want to keep the documentation changes, I can revert the change in the code. ;)

zekroTJA:
I've reverted the query escape change and added the reference to `Emoji.APIName` to get the emote identifier like mentioned by @iopred in f102817.

PR #824, opened 2020-10-24
Bump to v8, add Sticker struct

vysiondev:
This PR aims to bump discordgo from v6 to v8, as well as add the new Sticker struct that is currently present in the Discord API.

In accordance with the [v8 change logs](https://discord.com/developers/docs/change-log), the following changes have been made:

- `Game` no longer a member of any struct in favor of using `Activities[0]`
- Channel permission overwrite types are now 0 and 1 instead of "role" and "member".
- `Retry-After` header value is now 1 instead of 1000. Similarily, it is parsed using time.Second instead of time.Millisecond
- `delete-message-days` changed to `delete_message_days`.
- Removed `Roles` and `Nick` from the presence update struct
- All permission overwrites now deserialize to `string` instead of `int` but parsed when calculating overwrites

Additionally, the Sticker struct has been added. While it is not in the Discord API documentation, it has been exhibited in stable Discord clients and thus this struct has been built around what is currently known about stickers.

### DIFF
--- a/message.go
+++ b/message.go
@@ -122,20 +122,21 @@ type Message struct {
 	// be checked by performing a bitwise AND between this int and the flag.
 	Flags MessageFlags `json:"flags"`
 
-	// A sticker, which is a special type of message.
-	Stickers []*Sticker `json:"stickers"`
+	// A sticker which is a special type of message that displays a GIF.
+	// TODO: Currently not documented in Discord API but appears in stable clients.
+	Stickers *[]Sticker `json:"stickers"`
 }
 
 // Sticker contains the data that determines how a sticker is displayed in a channel.
 type Sticker struct {
-	ID           string `json:"string"`
-	Name         string `json:"name"`
-	Description  string `json:"description,omitempty"`
-	PackID       string `json:"pack_id"`
-	Asset        string `json:"asset"`
-	PreviewAsset string `json:"preview_asset,omitempty"`
-	FormatType   int    `json:"format_type"`
-	Tags         string `json:"tags,omitempty"`
+	ID           string  `json:"string"`
+	Name         string  `json:"name"`
+	Description  *string `json:"description,omitempty"`
+	PackID       string  `json:"pack_id"`
+	Asset        string  `json:"asset"`
+	PreviewAsset *string `json:"preview_asset,omitempty"`
+	FormatType   int     `json:"format_type"`
+	Tags         *string `json:"tags,omitempty"`
 }
 
 // MessageFlags is the flags of "message" (see MessageFlags* consts)
--- a/message.go
+++ b/message.go
@@ -121,6 +121,21 @@ type Message struct {
 	// This is a combination of bit masks; the presence of a certain permission can
 	// be checked by performing a bitwise AND between this int and the flag.
 	Flags MessageFlags `json:"flags"`
+
+	// A sticker, which is a special type of message.
+	Stickers []*Sticker `json:"stickers"`
+}
+
+// Sticker contains the data that determines how a sticker is displayed in a channel.
+type Sticker struct {
+	ID           string `json:"string"`
+	Name         string `json:"name"`
+	Description  string `json:"description,omitempty"`
+	PackID       string `json:"pack_id"`
+	Asset        string `json:"asset"`
+	PreviewAsset string `json:"preview_asset,omitempty"`
+	FormatType   int    `json:"format_type"`
+	Tags         string `json:"tags,omitempty"`
 }
 
 // MessageFlags is the flags of "message" (see MessageFlags* consts)
--- a/restapi.go
+++ b/restapi.go
@@ -33,7 +33,7 @@ import (
 // All error constants
 var (
 	ErrJSONUnmarshal           = errors.New("json unmarshal")
-	ErrStatusOffline           = errors.New("You can't set your Status to offline")
+	ErrStatusOffline           = errors.New("status cannot be set to offline")
 	ErrVerificationLevelBounds = errors.New("VerificationLevel out of bounds, should be between 0 and 3")
 	ErrPruneDaysBounds         = errors.New("the number of days should be more than or equal to 1")
 	ErrGuildNoIcon             = errors.New("guild does not have an icon set")
@@ -704,7 +704,7 @@ func (s *Session) GuildBanCreateWithReason(guildID, userID, reason string, days 
 
 	queryParams := url.Values{}
 	if days > 0 {
-		queryParams.Set("delete-message-days", strconv.Itoa(days))
+		queryParams.Set("delete_message_days", strconv.Itoa(days))
 	}
 	if reason != "" {
 		queryParams.Set("reason", reason)
--- a/restapi.go
+++ b/restapi.go
@@ -540,7 +540,7 @@ func memberPermissions(guild *Guild, channel *Channel, userID string, roles []st
 	for _, overwrite := range channel.PermissionOverwrites {
 		if guild.ID == overwrite.ID {
 			denyInt, _ := strconv.Atoi(overwrite.Deny)
-			allowInt, _ := strconv.Atoi(overwrite.Deny)
+			allowInt, _ := strconv.Atoi(overwrite.Allow)
 
 			apermissions &= ^denyInt
 			apermissions |= allowInt
@@ -554,9 +554,9 @@ func memberPermissions(guild *Guild, channel *Channel, userID string, roles []st
 	// Member overwrites can override role overrides, so do two passes
 	for _, overwrite := range channel.PermissionOverwrites {
 		for _, roleID := range roles {
-			if overwrite.Type == 1 && roleID == overwrite.ID {
+			if overwrite.Type == 0 && roleID == overwrite.ID {
 				denyInt, _ := strconv.Atoi(overwrite.Deny)
-				allowInt, _ := strconv.Atoi(overwrite.Deny)
+				allowInt, _ := strconv.Atoi(overwrite.Allow)
 
 				denies |= denyInt
 				allows |= allowInt
@@ -569,9 +569,9 @@ func memberPermissions(guild *Guild, channel *Channel, userID string, roles []st
 	apermissions |= allows
 
 	for _, overwrite := range channel.PermissionOverwrites {
-		if overwrite.Type == 0 && overwrite.ID == userID {
+		if overwrite.Type == 1 && overwrite.ID == userID {
 			denyInt, _ := strconv.Atoi(overwrite.Deny)
-			allowInt, _ := strconv.Atoi(overwrite.Deny)
+			allowInt, _ := strconv.Atoi(overwrite.Allow)
 
 			apermissions &= ^denyInt
 			apermissions |= allowInt
--- a/restapi.go
+++ b/restapi.go
@@ -508,6 +508,7 @@ func (s *Session) UserChannelPermissions(userID, channelID string) (apermissions
 // Calculates the permissions for a member.
 // https://support.discord.com/hc/en-us/articles/206141927-How-is-the-permission-hierarchy-structured-
 func memberPermissions(guild *Guild, channel *Channel, userID string, roles []string) (apermissions int) {
+
 	if userID == guild.OwnerID {
 		apermissions = PermissionAll
 		return
@@ -515,7 +516,8 @@ func memberPermissions(guild *Guild, channel *Channel, userID string, roles []st
 
 	for _, role := range guild.Roles {
 		if role.ID == guild.ID {
-			apermissions |= role.Permissions
+			p, _ := strconv.Atoi(role.Permissions)
+			apermissions |= p
 			break
 		}
 	}
@@ -523,7 +525,8 @@ func memberPermissions(guild *Guild, channel *Channel, userID string, roles []st
 	for _, role := range guild.Roles {
 		for _, roleID := range roles {
 			if role.ID == roleID {
-				apermissions |= role.Permissions
+				p, _ := strconv.Atoi(role.Permissions)
+				apermissions |= p
 				break
 			}
 		}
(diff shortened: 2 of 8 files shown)


structs.go
```diff
@@ -650,7 +644,7 @@ type Role struct {
 	// The permissions of the role on the guild (doesn't include channel overrides).
 	// This is a combination of bit masks; the presence of a certain permission can
 	// be checked by performing a bitwise AND between this int and the permission.
-	Permissions int `json:"permissions"`
+	Permissions string `json:"permissions"`
```
CarsonHoffman, line 647: Perhaps instead of changing this type, we can use the `,string` property on the struct tag to have `encoding/json` automatically convert it for us?

wsapi.go
```diff
@@ -334,22 +334,25 @@ type updateStatusOp struct {
...
+	if game != "" {
+		newActivities = append(newActivities, Game{
+			Name: game,
+			Type: gameType,
+			URL:  url,
+		})
+	}
+
 	usd := &UpdateStatusData{
 		Status: "online",
```
CarsonHoffman, line 348: Please make sure to run `gofmt` on all files before committing.

structs.go
```diff
@@ -325,9 +325,9 @@ type ChannelFollow struct {
 // A PermissionOverwrite holds permission overwrite data for a Channel
 type PermissionOverwrite struct {
 	ID    string `json:"id"`
-	Type  string `json:"type"`
-	Deny  int    `json:"deny"`
-	Allow int    `json:"allow"`
+	Type  int    `json:"type"`
```
CarsonHoffman, line 328: I know that we didn't have one before, but this seems like a great opportunity to define a constant in the library for easy comparison (rather than looking up the value in the Discord docs).

structs.go
```diff
@@ -325,9 +325,9 @@ type ChannelFollow struct {
...
+	Deny  string `json:"deny"`
+	Allow string `json:"allow"`
```
CarsonHoffman, line 330: See comment below on `Role.Permissions`.

restapi.go
```diff
@@ -508,22 +508,25 @@ func (s *Session) UserChannelPermissions(userID, channelID string) (apermissions
...
+
 	if userID == guild.OwnerID {
 		apermissions = PermissionAll
 		return
 	}
 
 	for _, role := range guild.Roles {
 		if role.ID == guild.ID {
-			apermissions |= role.Permissions
+			p, _ := strconv.Atoi(role.Permissions)
```
CarsonHoffman, line 519: See comment on `Role.Permissions` below. This would help clean up a lot of these unchecked errors, and means we would only pay the string to integer conversion cost once.

restapi.go
```diff
@@ -548,9 +554,12 @@ func memberPermissions(guild *Guild, channel *Channel, userID string, roles []st
 	// Member overwrites can override role overrides, so do two passes
 	for _, overwrite := range channel.PermissionOverwrites {
 		for _, roleID := range roles {
-			if overwrite.Type == "role" && roleID == overwrite.ID {
-				denies |= overwrite.Deny
-				allows |= overwrite.Allow
+			if overwrite.Type == 0 && roleID == overwrite.ID {
```
CarsonHoffman, line 557: See comment on `PermissionOverwrite.Type` below.

CarsonHoffman (changes requested):
Thank you for your changes! This is a great start; there are a few places where I think that the code could be cleaned up a bit, but otherwise the bulk of these changes look great, and I'm super excited to have the latest and greatest API version in the library!

vysiondev:
The most recent commit I've made should fix the requests.

structs.go
```diff
@@ -644,7 +653,7 @@ type Role struct {
 	// The permissions of the role on the guild (doesn't include channel overrides).
 	// This is a combination of bit masks; the presence of a certain permission can
 	// be checked by performing a bitwise AND between this int and the permission.
-	Permissions string `json:"permissions"`
+	Permissions int `json:",string,permissions"`
```
CarsonHoffman, line 656: Sorry, I should've specified more clearly—the `,string` option is appended to the value in the quotes, i.e. `json:"permissions,string"`.

structs.go
```diff
@@ -322,12 +322,21 @@ type ChannelFollow struct {
...
 // A PermissionOverwrite holds permission overwrite data for a Channel
 type PermissionOverwrite struct {
-	ID    string `json:"id"`
-	Type  int    `json:"type"`
-	Deny  string `json:"deny"`
-	Allow string `json:"allow"`
+	ID    string                  `json:"id"`
+	Type  PermissionOverwriteType `json:"type"`
+	Deny  int                     `json:",string,deny"`
+	Allow int                     `json:",string,allow"`
```
CarsonHoffman, line 339: Same as with `Role.Permissions`.

CarsonHoffman (changes requested):
Sorry; I didn't make clear how the `,string` option is used.

restapi.go
```diff
@@ -1785,13 +1786,13 @@ func (s *Session) ChannelInviteCreate(channelID string, i Invite) (st *Invite, e
...
 
 	data := struct {
-		ID    string `json:"id"`
-		Type  string `json:"type"`
-		Allow int    `json:"allow"`
-		Deny  int    `json:"deny"`
+		ID    string         `json:"id"`
+		Type  PermissionType `json:"type"`
+		Allow int            `json:"allow"`
+		Deny  int            `json:"deny"`
```
CarsonHoffman, line 1795: These should probably also be marked as `,string`. They seem to work currently, but I'm guessing that we're hitting some non-guaranteed coercion behavior on the API's side, which could break if they add more permissions.

CarsonHoffman (changes requested):
When testing, I also ran into a failure of `strconv.ParseInt` hitting a float, which seems to be related to the following line from the change log:
> `X-RateLimit-Reset` and `X-RateLimit-Reset-After` are always returned at millisecond precision (e.g. `123.456` instead of `124`)

See [`ratelimit.go:171`](https://github.com/bwmarrin/discordgo/blob/master/ratelimit.go#L171). It might also be a good idea to add a test specifically for that case! 🙂

vysiondev:
I believe that we can just drop the numbers after the decimal point, since the library handled it fine as an integer before.

CarsonHoffman:
It's true that the library handled it fine as an integer, but that's because the API only provided it as an integer (at second precision). Is there any particular reason we shouldn't use the millisecond precision now that we're provided with it?

vysiondev:
Sorry for getting back to this late--I made that commit earlier while sleep deprived (college applications and just generally not getting enough sleep). This commit should correctly use millisecond precision in calculating how long before the bucket should be released.

structs.go
```diff
@@ -1105,10 +1106,10 @@ type GatewayBotResponse struct {
 // GatewayStatusUpdate is sent by the client to indicate a presence or status update
 // https://discord.com/developers/docs/topics/gateway#update-status-gateway-status-update-structure
 type GatewayStatusUpdate struct {
-	Since  int      `json:"since"`
-	Game   Activity `json:"game"`
-	Status string   `json:"status"`
-	AFK    bool     `json:"afk"`
+	Since      int         `json:"since"`
+	Activities *[]Activity `json:"activities"`
```
CarsonHoffman, line 1110: Probably should be `[]*Activity`

wsapi.go
```diff
@@ -322,10 +322,10 @@ func (s *Session) heartbeat(wsConn *websocket.Conn, listening <-chan interface{}
 
 // UpdateStatusData ia provided to UpdateStatusComplex()
 type UpdateStatusData struct {
-	IdleSince *int   `json:"since"`
-	Game      *Game  `json:"game"`
-	AFK       bool   `json:"afk"`
-	Status    string `json:"status"`
+	IdleSince  *int    `json:"since"`
+	Activities *[]Game `json:"activities"`
```
CarsonHoffman, line 326: Probably should also be `[]*Game` to match `Presence.Activities`

CarsonHoffman (commented):
The rate-limit change looks great to me! Sorry for not noticing these earlier; it seems odd that we'd have a pointer to a slice, and given that the types were previously pointers to the types (with the exception of `GatewayStatusUpdate.Game`), I think it'd make more sense to have slices of pointers instead. Is that what you meant to do, or are you explicitly meaning to use pointers to slices here?

vysiondev:
My intention for making pointers to slices instead of an array of pointers was mainly based on the comment mentioning the relevant [section of the docs surrounding gateway status updates](https://discord.com/developers/docs/topics/gateway#update-status-gateway-status-update-structure), which indicated that it could be a nil value.

CarsonHoffman:
Ah, okay—that tracks, I didn't look closely enough at the changes. Is it a similar story with `Message.Stickers`? I only ask because the typical pattern in the library is to have slices of pointers to other objects, so the pointer to a slice of non-pointers is a bit of a departure from the usual pattern. (I know this might be getting into the nit territory, but I just want to make sure that a PR like this fully explores those types of things!)

vysiondev:
Yeah, the key only appears when the message contains a sticker/array of stickers. `Message.Stickers` could also be a nil value if no stickers were sent.

CarsonHoffman:
That's perfectly okay; slices can be `nil`. Unless we explicitly want to sent a `null` value for `stickers`, I'm not sure why the pointer to the slice is necessary (see `Message.Mentions`, for example, as another slice with the same characteristics).

structs.go
```diff
@@ -322,12 +322,22 @@ type ChannelFollow struct {
 	WebhookID string `json:"webhook_id"`
 }
 
+// PermissionType is the integer representing the permission type.
+// (PermissionTypeRole, PermissionTypeMember)
```
qaisjp, line 326:
```suggestion
```
I think this should be removed, as it duplicates the constant list below.

structs.go
qaisjp:
```suggestion
	Activities []*Activity `json:"activities"`
```

wsapi.go
qaisjp:
```suggestion
	Activities []*Game `json:"activities"`
```

wsapi.go
```diff
@@ -334,20 +334,23 @@ type updateStatusOp struct {
...
 
 func newUpdateStatusData(idle int, gameType GameType, game, url string) *UpdateStatusData {
-	usd := &UpdateStatusData{
-		Status: "online",
-	}
-
-	if idle > 0 {
-		usd.IdleSince = &idle
-	}
+	var newActivities []Game
```
qaisjp, line 337:
```suggestion
	var newActivities []*Game
```

wsapi.go
```diff
@@ -334,20 +334,23 @@ type updateStatusOp struct {
...
 
 	if game != "" {
-		usd.Game = &Game{
+		newActivities = append(newActivities, Game{
```
qaisjp, line 340:
```suggestion
		newActivities = append(newActivities, &Game{
```

wsapi.go
```diff
@@ -334,20 +334,23 @@ type updateStatusOp struct {
...
 			Name: game,
 			Type: gameType,
 			URL:  url,
-		}
+		})
+	}
+
+	usd := &UpdateStatusData{
+		Status:     "online",
+		Activities: &newActivities,
```
qaisjp, line 349:
```suggestion
		Activities: newActivities,
```

message.go
```diff
@@ -121,6 +121,22 @@ type Message struct {
 	// This is a combination of bit masks; the presence of a certain permission can
 	// be checked by performing a bitwise AND between this int and the flag.
 	Flags MessageFlags `json:"flags"`
+
+	// A sticker which is a special type of message that displays a GIF.
+	// TODO: Currently not documented in Discord API but appears in stable clients.
+	Stickers *[]Sticker `json:"stickers"`
```
qaisjp, line 127:
```suggestion
	Stickers []*Sticker `json:"stickers"`
```

vysiondev:
Sorry, issues with commit history. Everything should be up to date now

wsapi.go
```diff
@@ -322,10 +322,10 @@ func (s *Session) heartbeat(wsConn *websocket.Conn, listening <-chan interface{}
 
 // UpdateStatusData ia provided to UpdateStatusComplex()
 type UpdateStatusData struct {
-	IdleSince *int   `json:"since"`
-	Game      *Game  `json:"game"`
-	AFK       bool   `json:"afk"`
-	Status    string `json:"status"`
+	IdleSince  *int    `json:"since"`
```
qaisjp, line 325: GatewayStatusUpdate uses `int` but `UpdateStatusData` uses `*int`.

I think there should be some consistency here.

restapi.go
```diff
@@ -157,7 +157,7 @@ func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b
 		s.log(LogInformational, "Rate Limiting %s, retry in %d", urlStr, rl.RetryAfter)
```
ewohltman, line 157: Since `rl.RetryAfter` is now a `float64`, this format verb should be `%f`.

```suggestion
		s.log(LogInformational, "Rate Limiting %s, retry in %f", urlStr, rl.RetryAfter)
```

structs.go
```diff
@@ -835,7 +835,7 @@ type Relationship struct {
 type TooManyRequests struct {
 	Bucket     string        `json:"bucket"`
 	Message    string        `json:"message"`
-	RetryAfter time.Duration `json:"retry_after"`
+	RetryAfter float64       `json:"retry_after"`
 }
```
ewohltman, line 839: This needs to be `gofmt`ed:

```suggestion
type TooManyRequests struct {
	Bucket     string  `json:"bucket"`
	Message    string  `json:"message"`
	RetryAfter float64 `json:"retry_after"`
}
```

restapi.go
```diff
@@ -157,7 +157,7 @@ func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b
...
 		s.handleEvent(rateLimitEventType, RateLimit{TooManyRequests: &rl, URL: urlStr})
 
-		time.Sleep(rl.RetryAfter * time.Millisecond)
+		time.Sleep(time.Duration(rl.RetryAfter * 1000) * time.Millisecond)
```
ewohltman, line 160: This should be `gofmt`ed:

```suggestion
		time.Sleep(time.Duration(rl.RetryAfter*1000) * time.Millisecond)
```

ewohltman (changes requested):

CarsonHoffman:
Closing in favor of #866. Sorry that this wasn't able to be merged, but with the account deletion and v6 deprecation underway we didn't have much of a choice.

Issue #823, opened 2020-10-17
VoiceStatusUpdate - Unable to determine what channel a user left from


nickmcski:
I'm trying to run an action when a user disconnects from a specific voice channel. The discord API shows the disconnect by firing a Voice State Update with a channel ID of null, as such state tracking is required to determine which server the user left from.

I've tried using the **guild.voiceStates** to determine what channel the user was in previous, but this information is already cleared out before the **VoiceStateUpdate** handler is called

```cs
func voiceStateUpdate(s *discordgo.Session, m *discordgo.VoiceStateUpdate) {

	if m.ChannelID == "" { //User disconnected from a voice channel
		guild, _ := s.State.Guild(m.GuildID)

		for _, key := range guild.VoiceStates {
			if key.UserID == m.UserID {
				//This code is never reached as the user was already removed from the VoiceStates array
				println(key.UserID, " left channel ", key.ChannelID)
			}
		}
	}
}
```

I would like feedback on this potential solution before I submit a pull request:

Trigger the new event **VoiceStateDisconnect** from **voiceStateUpdate** whenever an entry is removed.

I'm still fairly new to Go so any feedback or alternative solutions are greatly appreciated.

Mexican-Man:
I think this would be _incredibly useful._ Here's a thought: Discord.js's "voiceStateUpdate" event sends two states, and "old" and a "new" one. I think that gets the best of both worlds: being able to track complex states, while also keeping it simple.

Here's a code example:
Old:
```go
type VoiceStateUpdate struct {
    *VoiceState
}
```
New:
```go
type VoiceStateUpdate struct {
    OldState *VoiceState
    NewState *VoiceState
}
```

Technically this could be handled by the user, but functionality is looking poor. My first thought would be to just go through the guild object and read the voice states, but:
```go
// This field is only present in GUILD_CREATE events and websocket    <---- here
// update events, and thus is only present in state-cached guilds.
VoiceStates []*VoiceState `json:"voice_states"`
``` 
From Discord's docs, it sounds like GUILD_CREATE fires on startup anyway, so you might be able to get away with tracking each individual user's voiceState on startup, then managing it internally from there. If not, you would have to deal with losing the current voiceState on restart.

[...]

PR #828, opened 2020-10-26
Add an event for VoiceStateDisconnect 

nickmcski:
#823 These changes allows for using the built in State table to track when a users leaves a particular channel. 

As an example this can be useful when you want to assign temporary permissions to a user when they are in a particular voice channel


Mexican-Man:
I think this would be _incredibly useful._ Here's a thought: Discord.js's "voiceStateUpdate" event sends two states, and "old" and a "new" one. I think that gets the best of both worlds: being able to track complex states, while also keeping it simple.

Here's a code example:
Old:
```go
type VoiceStateUpdate struct {
    *VoiceState
}
```
New:
```go
type VoiceStateUpdate struct {
    OldState *VoiceState
    NewState *VoiceState
}
```

Technically this could be handled by the user, but functionality is looking poor. My first thought would be to just go through the guild object and read the voice states, but:
```go
// This field is only present in GUILD_CREATE events and websocket    <---- here
// update events, and thus is only present in state-cached guilds.
VoiceStates []*VoiceState `json:"voice_states"`
``` 
From Discord's docs, it sounds like GUILD_CREATE fires on startup anyway, so you might be able to get away with tracking each individual user's voiceState on startup, then managing it internally from there. If not, you would have to deal with losing the current voiceState on restart.

If you're trying to distinguish between joining from nowhere, vs from another channel, you could keep an internal list of connected members (as mentioned above). When you receive a voiceStateUpdate from that user add them to the list. If you receive a voiceStateUpdate "disconnect", set a timer to check again, then delete them from the list. If they disconnect for more than X time, they won't be on the list. If they are on the list, then you know they probably came from another channel, or only briefly disconnected.
```go
import "time"

...
time.AfterFunc(3*time.Second, somefunction)
```

nickmcski:
After discussion on issue #823, a better solution has been presented. These changes have been opened as part of pull request #833

PR #833, opened 2020-11-05
Add previous state to VoiceStateUpdate event (#823)

nickmcski:
Similar to MessageUpdate, this change allows tracking which voice channel a user left from.

This uses the existing VoiceState tracker to include the previous VoiceState as part of the VoiceStateUpdate event.


CarsonHoffman:
Thank you for adding this into the library!

PR #842, opened 2020-11-20
Add support for sending message reply

Zabqer:
Added the ability to send messages with a response (replying to another message). According to the `message_reference` field in API

https://discord.com/developers/docs/resources/channel#create-message-params

Code example:
```go
session.ChannelMessageSendComplex(channelID, &discordgo.MessageSend{
	Content:   "Test",
	Reference: message.Reference(),
})
```
Or:
```go
session.ChannelMessageSendReply(channelID, "Test", message.Reference())
```

### DIFF
--- a/message.go
+++ b/message.go
@@ -150,6 +150,7 @@ type MessageSend struct {
 	TTS             bool                    `json:"tts"`
 	Files           []*File                 `json:"-"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
+	Reference       *MessageReference       `json:"message_reference,omitempty"`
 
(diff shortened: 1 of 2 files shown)


elliotwms (approved):

FedorLap2006 (approved):

FedorLap2006:
Great idea. It will be really useful!

CarsonHoffman:
It seems to me like adding this into messages universally rather than only on sends would be essentially a one-line change since the `MessageReference` type is already being added, and would allow the consumption of message references as well; I think that encapsulating both pieces into one PR would be beneficial for the feature!

CarsonHoffman:
Thank you for the addition; the reply feature is very cool!

PR #845, opened 2020-12-03
Add omitempty to MessageReference's GuildID field

LagSwitchedVirginity:
Direct Messages do not have the **guild_id** field for Message Referencing / the new Reply feature.
Instead it is only **message_id** and **guild_id**


CarsonHoffman:
Good catch; thank you for the contribution!

PR #848, opened 2020-12-11
GuildBan method and UnknownBan error code addition

FedorLap2006:
I added UnknownBan code to error codes in structs.go and also I made GuildBan method, so we can now get user's ban in a guild.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -693,6 +693,18 @@ func (s *Session) GuildBanCreate(guildID, userID string, days int) (err error) {
 	return s.GuildBanCreateWithReason(guildID, userID, "", days)
 }
 
+// GuildBan finds ban by given guild and user id and returns GuildBan structure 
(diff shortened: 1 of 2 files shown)


CarsonHoffman:
I only see the addition of `ErrCodeUnknownBan`; it looks like you may have forgotten to add the other changes. 🙂

FedorLap2006:
> I only see the addition of `ErrCodeUnknownBan`; it looks like you may have forgotten to add the other changes.

Oh... right. Wait a sec. I will upload them

FedorLap2006:
Everything is ok... but note: I've done commit amend, so if something broke, please mention me :slightly_smiling_face:

FedorLap2006:
Btw, why is the check still tries to run itself?

CarsonHoffman:
Thank you for the contribution (and following up with the CI)!

PR #855, opened 2020-12-22
implement interaction request signing verification

bsdlp:
i had previously written a small lib to do this: [bsdlp/discord-interactions-go](https://github.com/bsdlp/discord-interactions-go), and this is basically copypasta from there.

https://discord.com/developers/docs/interactions/slash-commands#security-and-authorization

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -1,0 +1,54 @@
+package discordgo
+
+import (
+	"bytes"
+	"crypto/ed25519"
+	"encoding/hex"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
+
+// VerifyInteraction implements message verification of the discord interactions api
+// signing algorithm, as documented here:
+// https://discord.com/developers/docs/interactions/slash-commands#security-and-authorization
+func VerifyInteraction(r *http.Request, key ed25519.PublicKey) bool {
+	var msg bytes.Buffer
+
+	signature := r.Header.Get("X-Signature-Ed25519")
+	if signature == "" {
+		return false
+	}
+
+	sig, err := hex.DecodeString(signature)
+	if err != nil {
+		return false
+	}
+
+	if len(sig) != ed25519.SignatureSize {
+		return false
+	}
+
+	timestamp := r.Header.Get("X-Signature-Timestamp")
+	if timestamp == "" {
+		return false
+	}
+
+	msg.WriteString(timestamp)
+
+	defer r.Body.Close()
+	var body bytes.Buffer
+
+	// at the end of the function, copy the original body back into the request
+	defer func() {
+		r.Body = ioutil.NopCloser(&body)
+	}()
+
+	// copy body into buffers
+	_, err = io.Copy(&msg, io.TeeReader(r.Body, &body))
+	if err != nil {
+		return false
+	}
+
+	return ed25519.Verify(key, msg.Bytes(), sig)
+}
--- a/interactions_test.go
+++ b/interactions_test.go
@@ -1,0 +1,68 @@
+package discordgo
+
+import (
+	"bytes"
+	"crypto/ed25519"
+	"encoding/hex"
+	"net/http/httptest"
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -1,7 +1,9 @@
 module github.com/bwmarrin/discordgo
 
 require (
+	filippo.io/edwards25519 v1.0.0-beta.2 // indirect
```
bsdlp, line 4: this is a test dependency from the nacl lib

go.mod
```diff
@@ -1,7 +1,9 @@
...
 	github.com/gorilla/websocket v1.4.0
+	github.com/kevinburke/nacl v0.0.0-20201212185859-3ef16edae976
```
bsdlp, line 6: this library implements nacl signing, and is a thin wrapper around `crypto/ed25519`.

CarsonHoffman:
Hi! This library hasn't traditionally featured any type of HTTP server functionality, but given the new interactions feature I think that this might be a good addition to consider! A few things I'm unsure about:

1. Do we really need a third-party dependency on `github.com/kevinburke/nacl/sign`? The check can be completed by `crypto/ed25519` outright, and I'd prefer not adding additional third-party dependencies (especially crypto ones!) if it's not necessary.
2. Is there a particular reason that you've vendored all dependencies in this change? We haven't previously, and while I'm not inherently against it, I think that it's a bit out of the scope of this change.

bsdlp:
you're 100% right. i've removed the external lib dependency. thanks!

bsdlp:
build is failing due to `crypto/ed25519` not existing in go stdlib before 1.13. imo those versions can be dropped, go release policy says the currently supported releases are `1.14` and `1.15`.

Jleagle:
Why do you touch the request body? Seems unnecessary

ewohltman:
> Why do you touch the request body? Seems unnecessary

Looks like we need to read the request body to validate the signature, so a copy of the original `io.Reader` body is created to replace the request body 
after it's read from for the signature check before we return. Other functions later that use the same request can then read the body in normally.

interactions.go
```diff
@@ -0,0 +1,54 @@
...
+	if signature == "" {
+		return false
+	}
+
+	sig, err := hex.DecodeString(signature)
+	if err != nil {
+		return false
+	}
+
+	if len(sig) != ed25519.SignatureSize || sig[63]&224 != 0 {
```
ewohltman, line 28: `63` and `224` seem to be "magic" numbers. Does it make sense to define them as `const`s so we know what they represent?

interactions.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	timestamp := r.Header.Get("X-Signature-Timestamp")
+	if timestamp == "" {
+		return false
+	}
+
+	msg.WriteString(timestamp)
+
+	defer r.Body.Close()
```
ewohltman, line 39: If I am not mistaken, this will close the copied request body by firing after the `defer` defined below.

Perhaps we should but the two statements together in a single defer?

```go
var body bytes.Buffer

defer func() {
    _ = r.Body.Close()
    r.Body = ioutil.NopCloser(&body)
}
```

ewohltman (changes requested):

interactions.go
bsdlp: i just removed this check as it appears this already happens in the stdlib implementation and i'd rather just check signaturesize since it is a defined const. https://golang.org/src/crypto/ed25519/ed25519.go#L208

interactions.go
CarsonHoffman: The `r.Body` receiver is evaluated when the `defer` executes (rather than when the call to `Close`) executes. This seems fine to me.

CarsonHoffman:
Thank you for the addition! This should definitely be useful for the adoption of webhook interactions 😃

bsdlp:
thanks for your work on this lib!

PR #866, opened 2021-01-21
Switch library to API version v8

CarsonHoffman:
Hi all!

I'm currently in the process of implementing API version 8 support; there was a previous PR, but it seems to have stalled out (with the user contributing having their account vanish). I've started an implementation on the `v8` branch, and would appreciate some review on it before we go ahead and merge!

There are some breaking changes: the `Game` struct is completely gone (the API doesn't have any such concept anymore), and `UpdateStatus` was changed to `UpdateGameStatus` to better reflect the true meaning of it now that games don't have elevated status; presences no longer contain information about nicknames and roles, which are now relegated to members; permissions are now represented throughout the library by `int64` rather than `int` now that they can go above 31 bits; permission overwrite types switched from a string to an integer type. There might be some I missed. Although breaking changes are never desired, they correspond to the breaking nature of switching API versions, and the current tags will still exist for those looking to remain on v6.

The ultimate goal here is to get the changes into the hands of users, as there are inevitably changes that won't be picked up by the test suite (this does speak to the quality of the test suite to an extent, but fully modeling API behavior is also hard). v6 is going by the wayside soon, and this should help propel DiscordGo into a better position for the future!

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -22,7 +22,7 @@ import (
 )
 
 // VERSION of DiscordGo, follows Semantic Versioning. (http://semver.org/)
-const VERSION = "0.22.0"
+const VERSION = "0.23.0-v8alpha"
 
 // ErrMFA will be risen by New when the user has 2FA.
 var ErrMFA = errors.New("account has 2FA enabled")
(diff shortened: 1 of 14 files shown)


CarsonHoffman:
With no particular input, the motion passes! I expect that there might be issues, so please file an issue if anything comes up; I'm perfectly okay with releasing fairly quickly if it helps push out v8 support as fast as possible.

PR #867, opened 2021-01-27
Set channel and role edit perms to int64

IntrntSrfr:
Should not the set role/channel permissions methods also take in int64s?

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1048,7 +1048,7 @@ func (s *Session) GuildRoleCreate(guildID string) (st *Role, err error) {
 // hoist     : Whether to display the role's users separately.
 // perm      : The permissions for the role.
 // mention   : Whether this role is mentionable
-func (s *Session) GuildRoleEdit(guildID, roleID, name string, color int, hoist bool, perm int, mention bool) (st *Role, err error) {
+func (s *Session) GuildRoleEdit(guildID, roleID, name string, color int, hoist bool, perm int64, mention bool) (st *Role, err error) {
 
 	// Prevent sending a color int that is too big.
 	if color > 0xFFFFFF {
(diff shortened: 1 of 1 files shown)


restapi.go
```diff
@@ -1060,7 +1060,7 @@ func (s *Session) GuildRoleEdit(guildID, roleID, name string, color int, hoist b
 		Name        string `json:"name"`        // The role's name (overwrites existing)
 		Color       int    `json:"color"`       // The color the role should have (as a decimal, not hex)
 		Hoist       bool   `json:"hoist"`       // Whether to display the role's users separately
-		Permissions int    `json:"permissions"` // The overall permissions number of the role (overwrites existing)
+		Permissions int64  `json:"permissions"` // The overall permissions number of the role (overwrites existing)
```
CarsonHoffman, line 1063: To serialize as a string, the JSON tag should be `` `json:"permissions,string"` ``.

restapi.go
```diff
@@ -1807,13 +1807,13 @@ func (s *Session) ChannelInviteCreate(channelID string, i Invite) (st *Invite, e
...
-func (s *Session) ChannelPermissionSet(channelID, targetID string, targetType PermissionOverwriteType, allow, deny int) (err error) {
+func (s *Session) ChannelPermissionSet(channelID, targetID string, targetType PermissionOverwriteType, allow, deny int64) (err error) {
 
 	data := struct {
 		ID    string                  `json:"id"`
 		Type  PermissionOverwriteType `json:"type"`
-		Allow int                     `json:"allow"`
-		Deny  int                     `json:"deny"`
+		Allow int64                   `json:"allow"`
+		Deny  int64                   `json:"deny"`
```
CarsonHoffman, line 1816: Same as above.

CarsonHoffman (changes requested):
Great catch! In addition to the issue you mentioned, we'll also need to switch to serializing these permissions as strings; this is an issue we missed originally and isn't your fault, but would be great to fix in this change!

IntrntSrfr:
I think I've added what you mean

IntrntSrfr (commented):
Done

CarsonHoffman (approved):
Awesome, LGTM! It appears that API v8 won't reject integers in the JSON payload so this change doesn't need to be immediately deployed, but it'll most likely be rolled out with other smaller changes. Thank you for the contribution! 😃

PR #868, opened 2021-01-28
Update badge

maful:
Since godoc.org will be redirected to pkg.go.dev, I think this project can start to use the new badge.

Badge generation tool: https://pkg.go.dev/badge/

Source: https://github.com/golang/go/issues/43178


CarsonHoffman:
Thank you for this; having the pkg.go.dev badge is a small but surprisingly meaningful thing 😄

PR #877, opened 2021-02-14
Update dependencies for voice receive example

RichardoC:
This avoids "./main.go:78:21: cannot use discordgo.IntentsGuildVoiceStates (type discordgo.Intent) as type *discordgo.Intent in assignment"


examples/voice_receive/go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/bwmarrin/discordgo/examples/voice_receive
 
 require (
-	github.com/bwmarrin/discordgo v0.21.1
-	github.com/pion/rtp v1.6.0
-	github.com/pion/webrtc/v3 v3.0.0-20200721060053-ca3cc9d940bc
+    github.com/bwmarrin/discordgo v0.23.1
```
FedorLap2006, line 4: We have recently released v0.24.0, so I think it would be fair to update it here as well
```suggestion
    github.com/bwmarrin/discordgo v0.24.0
```

FedorLap2006 (commented):
Sorry it took a while

FedorLap2006:
Fixed by #1182, but going to merge this nonetheless, as having old dependencies in the examples isn't a good idea.

FedorLap2006:
After a little bit more thinking, we decided to fully drop the voice receive example, as it is undocumented, and therefore I'll close this PR.

PR #880, opened 2021-02-22
Fix RTP extended header parsing

galberti:
As ext header length isn't always 1, the shift to find the beginning of
the Opus packet must be adjusted dynamically.


CarsonHoffman:
Thank you for delving into the world of no documentation to figure the issue out here, and the contribution! 🤠

PR #882, opened 2021-02-27
Add ping pong example for direct messages

chanbakjsd:
This PR adds an example on sending direct messages as it's a frequently asked question on the Discord Gophers server.
The example sends a "Pong" response through DM when a user sends "ping" in any channel the bot has the permission to see.

The example is simply adopted from the ping pong example with minor changes to respond through DM.


CarsonHoffman:
Thank you for the comprehensive example! 🙂

PR #883, opened 2021-02-27
Updated Activity struct and added an emoji function.

post04:
I've added everything to the Activity struct that is listed [here](https://discord.com/developers/docs/topics/gateway#activity-object) and added a function to the Message struct that gets all the emojis from the content of the message, it then returns a Slice of Emoji structs.

### DIFF
--- a/message.go
+++ b/message.go
@@ -125,6 +125,24 @@ type Message struct {
 	Flags MessageFlags `json:"flags"`
 }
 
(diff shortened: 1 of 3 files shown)


postrequest69:
I'm not sure why the build failed; I tested the PR on a bot in around 50 servers getting a lot of presence updates and it seemed to not have any problems after that fix.

PR #886, opened 2021-03-03
Fix slash commands example bug

Nv7-GitHub:
This fixes a bug in the slash commands example that causes an out-of-range error.

### DIFF
--- a/examples/slash_commands/main.go
+++ b/examples/slash_commands/main.go
@@ -150,6 +150,7 @@ var (
 							Value: 5,
 						},
 					},
+					Required: true,
 				},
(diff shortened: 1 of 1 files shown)


FedorLap2006 (approved):
That's not a bug tho, but yet needs to be fixed. Thanks.

FedorLap2006:
Also I think, for upcoming changes I need to change the example, especially "responses" part.

Nv7-GitHub:
Yeah, it seems like you just forgot to put a line of code, which can cause an out-of-range error. Is this PR going to be merged?

PR #895, opened 2021-03-17
update Activity struct and add emoji functions

post04:
I added a function to easily get the emojis out of a message and updated the Activity struct to include everything from the documentation.

### DIFF
--- a/message.go
+++ b/message.go
@@ -125,6 +125,24 @@ type Message struct {
 	Flags MessageFlags `json:"flags"`
 }
 
+// GetEmojis pulls out all the emojis from a message and returns a Slice of the Emoji struct.
+func (m *Message) GetEmojis() []*Emoji {
+	var toReturn []*Emoji
+	emojis := EmojiRegex.FindAllString(m.Content, -1)
+	if len(emojis) < 1 {
+		return toReturn
+	}
+	for _, em := range emojis {
+		parts := strings.Split(em, ":")
+		toReturn = append(toReturn, &Emoji{
+			ID:       parts[2][:len(parts[2])-1],
+			Name:     parts[1],
+			Animated: strings.HasPrefix(em, "<a:"),
+		})
+	}
+	return toReturn
+}
+
 // MessageFlags is the flags of "message" (see MessageFlags* consts)
 // https://discord.com/developers/docs/resources/channel#message-object-message-flags
 type MessageFlags int
--- a/message.go
+++ b/message.go
@@ -125,8 +125,8 @@ type Message struct {
 	Flags MessageFlags `json:"flags"`
 }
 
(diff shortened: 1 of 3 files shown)


CarsonHoffman:
I'm a bit confused about the use of inline struct definitions here—we have struct definitions in the library for the [`Timestamps`](https://pkg.go.dev/github.com/bwmarrin/discordgo#TimeStamps), [`Emoji`](https://pkg.go.dev/github.com/bwmarrin/discordgo#Emoji), and [`Assets`](https://pkg.go.dev/github.com/bwmarrin/discordgo#Assets) fields, and the remaining two fields should be fairly easy to extract into named types. Using inline struct definitions makes the types extremely difficult to work with (specifying a literal of the type requires repeating the entire struct definition, rather than the name of a type).

In addition, given that `CreatedAt` represents a time, it seems to me that it should really be of type `time.Time`. Not all types in the library currently exhibit this behavior, but implementing a custom `UnmarshalJSON` is pretty trivial (especially once it is a named type!), and saves the user from having to perform this conversion themself.

post04:
> I'm a bit confused about the use of inline struct definitions here—we have struct definitions in the library for the [`Timestamps`](https://pkg.go.dev/github.com/bwmarrin/discordgo#TimeStamps), [`Emoji`](https://pkg.go.dev/github.com/bwmarrin/discordgo#Emoji), and [`Assets`](https://pkg.go.dev/github.com/bwmarrin/discordgo#Assets) fields, and the remaining two fields should be fairly easy to extract into named types. Using inline struct definitions makes the types extremely difficult to work with (specifying a literal of the type requires repeating the entire struct definition, rather than the name of a type).
> 
> In addition, given that `CreatedAt` represents a time, it seems to me that it should really be of type `time.Time`. Not all types in the library currently exhibit this behavior, but implementing a custom `UnmarshalJSON` is pretty trivial (especially once it is a named type!), and saves the user from having to perform this conversion themself.

Thanks for the response! I completely understand what you're saying about the inline type definitions. The only thing I'm not understanding is CreatedAt being a time.Time, I've tried to do this and it seems to return an error unmarshaling. With that being said I've made all the other edits!

hhhapz:
Please correct me if I'm incorrect, but it doesn't seem like your current implementation supports extracting unicode emojis either. Is that intentional?

post04:
> Please correct me if I'm incorrect, but it doesn't seem like your current implementation supports extracting unicode emojis either. Is that intentional?

You'd be correct, that's intentional as I'm not sure how I'd be able to do that. I could pull the list from discord but besides adding that big chunk of data I'm unsure how I'd go about doing that. Perhaps I should rename the func to `GetCustomEmojis`?

CarsonHoffman:
Thank you for the changes! Having that timestamp parsing done in the unmarshaling step makes the `CreatedAt` field much simpler to interact with 🤠

PR #896, opened 2021-03-21
Update gorilla to 1.4.2

mutemule:
This addresses a [DoS in gorilla/websocket](https://github.com/gorilla/websocket/security/advisories/GHSA-jf24-p9p9-4rjh). Exploitation appears to require that Discord be malicious,which is very low risk, but it's probably worth closing off regardless.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/bwmarrin/discordgo
 
 require (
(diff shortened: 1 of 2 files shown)


CarsonHoffman:
Thank you for spotting this, this is definitely a no-brainer update!

Issue #904, opened 2021-04-04
Missing permissions in the `Permissions` constants


evieDelta:
i noticed there are a couple of permissions which are missing from the permission constants

Specifically
- `1 << _9` Voice stream video
- `1 << 19` View guild insights
- `1 << 31` Use slash commands
- `1 << 32` Request to speak

(the latter 2 are not currently in the official documentation, and the last one was only added within the past week, but being they are launched features i don't think they'll see any sudden changes at this point and are probably safe to add, and the former 2 have been around for awhile and really shouldn't be missing)

and i also noticed that multiple permissions are missing from the combination constants, notably
- `PermissionUseExternalEmojis`
- `PermissionChangeNickname`
- `PermissionManageNicknames`

are completely absent from even PermissionsAll, in addition to the permissions which are missing from the constants

to which, I'd propose

to `PermissionsAlltext`
- `PermissionUseExternalEmojis`

to `PermissionsAllVoice`
- `PermissionVoiceStreamVideo`
- `PermissionVoiceRequestToSpeak`

to `PermissionsAll`
- `PermissionChangeNickname`
- `PermissionManageNicknames`
- `PermissionViewGuildInsights`

to either `PermissionsAll` or to `PermissionsAllText`
- `PermissionUseSlashCommands`

I've got a commit pretty much ready to go for the constants that are just missing so I'll probably submit a PR for that part later

But i wanted to open an issue about it before submitting anything including any kind of change to the `PermissionsAll_` combination constants, because while changing them will make them more accurate to their name, any such change is potential for unintended changes in behavior in anything that might use them (it shouldn't break anything since its only adding permissions, but there is still risk a risk it may cause a bot to act unexpectedly from how it would before) so i figured it was necessary for that to have a period of discussion before any changes are PR'ed

PR #906, opened 2021-04-05
Add in some missing permission constants

evieDelta:
Half of what i mentioned in #904, adds in the 4 missing permissions constants, specifically:
- `1 << _9` PermissionVoiceStreamVideo
- `1 << 19` PermissionViewGuildInsights
- `1 << 31` PermissionUseSlashCommands
- `1 << 32` PermissionRequestToSpeak

to quote from the issue
> (the latter 2 are not currently in the official documentation, and the last one was only added within the past week, but being they are launched features i don't think they'll see any sudden changes at this point and are probably safe to add, and the former 2 have been around for awhile and really shouldn't be missing)

Nothing else, just adds those 4 constants

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -1167,6 +1167,7 @@ const (
 	PermissionReadMessageHistory
 	PermissionMentionEveryone
 	PermissionUseExternalEmojis
+	PermissionUseSlashCommands = 1 << (iota + 22)
 )
 
 // Constants for the different bit offsets of voice permissions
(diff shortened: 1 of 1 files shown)


CarsonHoffman:
I just added change 843c765ae3eeecaf8e9b808ea8e3da10ba600f91 because of this PR—quite frankly, it was impossible to review this PR without breaking out an editor to confirm the values, which isn't an ideal workflow for anybody (though I recognize that this is fully not your fault—I understand you were working with the style used in the library!). If you could update this PR to use the absolute values of those permissions like this change, that would be awesome, and would make it much easier to confirm that the values are correct! 😃

evieDelta:
alright, the changes made to the upstream should be merged into the PR and the new constants are now also using the absolute values

(and yeah, it certainly would have been easier for me too if it used absolutes, since my initial testing was done with absolutes i could have just copy and pasted and not have had to spend time figuring out offsets for iota or testing to make sure i didn't mess anything up)

CarsonHoffman:
Awesome, thank you for making those changes so quickly!

PR #912, opened 2021-04-24
Add GuildMemberDisconnect (Disconnects a user from a voice channel)

Not-Cyrus:
I noticed there was no "proper" way to disconnect a user from a voice chat. (Of which I needed to do), so I decided to implement it myself.

This will disconnect the user from any voice channel they are in, (Just like how the discord client does it) without the need of hacky-code (Albeit, this code itself is hacky but less hacky for the end-user)

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -892,10 +892,23 @@ func (s *Session) GuildMemberNickname(guildID, userID, nickname string) (err err
 	return
 }
 
+// GuildMemberDisconnect disconnects a guild member from a voice chat.
+//  guildID   : The ID of a Guild.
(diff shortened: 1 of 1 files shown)


CarsonHoffman:
In what way does this differ from the existing [`GuildMemberMove`](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L855)?

Not-Cyrus:
Sorry, I did not realize that was a thing that you could do with that function, (Although it is kindof confusing to look for in the docs. I'll close this now)

(I feel stupid now lmao)

PR #916, opened 2021-04-28
Best way to detect that the author is not a bot

marcos-dionisio:


jolheiser:
The examples are just ignoring messages from the example bot itself, not _all_ bots.

Generally I'd agree with the PR in terms of ignoring bots, but the comment would need to be changed to reflect that it's now ignoring _all_ bots, not just itself.

PR #922, opened 2021-05-03
Threads

Pedro-Pessoa:
I've started the work on [threads](https://discord.com/developers/docs/topics/threads).
This PR also adds missing message types/flags.
Any feedback/help is greatly appreciated. Mainly what is missing are the endpoints and the RESTAPI funcs.

-- EDIT
Everything should be done now

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2650,7 +2650,7 @@ func (s *Session) LeaveThread(channelID string) (err error) {
 	return
 }
 
-// RemoveUserFromThread adds another user to a thread
+// RemoveUserFromThread removes another user from a thread
 // PUT /channels/{channel.id}/thread-members/{user.id}
 func (s *Session) RemoveUserFromThread(channelID, uID string) (err error) {
 	_, err = s.RequestWithBucketID("DELETE", EndpointThreadMembers(channelID)+uID, nil, EndpointThreadMembers(""))
--- a/restapi.go
+++ b/restapi.go
@@ -2128,8 +2128,18 @@ func (s *Session) WebhookDeleteWithToken(webhookID, token string) (st *Webhook, 
 func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *WebhookParams) (st *Message, err error) {
 	uri := EndpointWebhookToken(webhookID, token)
 
+	v := url.Values{}
+
 	if wait {
-		uri += "?wait=true"
+		v.Set("wait", "true")
+	}
+
+	if data.ThreadID != "" {
+		v.Set("thread_id", data.ThreadID)
+	}
+
+	if len(v) > 0 {
+		uri += "?" + v.Encode()
 	}
 
 	var response []byte
--- a/restapi.go
+++ b/restapi.go
@@ -2639,7 +2639,7 @@ func (s *Session) RemoveUserFromThread(channelID, uID string) (err error) {
 	return
 }
 
-// TheadMembers lists the members of a thread
+// ThreadMembers lists the members of a thread
 // GET /channels/{channel.id}/threads-members
 func (s *Session) ThreadMembers(channelID string) (st []*ThreadMember, err error) {
 	body, err := s.RequestWithBucketID("GET", EndpointThreadMembers(channelID), nil, EndpointThreadMembers(""))
@@ -2651,6 +2651,8 @@ func (s *Session) ThreadMembers(channelID string) (st []*ThreadMember, err error
 	return
 }
 
+// ThreadsResponseBody is the returned data from the discord gateway
+// of some specific threads requests
 type ThreadsResponseBody struct {
 	Threads []*Channel      `json:"threads"`
 	Members []*ThreadMember `json:"members"`
@@ -2767,7 +2769,7 @@ func (s *Session) JoinedPrivateArchivedThreads(channelID string, before *time.Ti
 // threadID  : The ID of a Thread
 // name       : The new name to assign the thread.
 func (s *Session) ThreadEdit(channelID string, name string) (*Channel, error) {
-	return s.ThreadEditComplex(channelID, &ThreadEdit{
+	return s.ThreadEditComplex(channelID, &ThreadEditData{
 		Name: name,
 	})
 }
@@ -2775,7 +2777,7 @@ func (s *Session) ThreadEdit(channelID string, name string) (*Channel, error) {
 // ThreadEditComplex edits an existing thread, replacing the parameters entirely with ThreadEdit struct
 // threadID  : The ID of a Thread
 // data          : The thread struct to send
-func (s *Session) ThreadEditComplex(threadID string, data *ThreadEdit) (st *Channel, err error) {
+func (s *Session) ThreadEditComplex(threadID string, data *ThreadEditData) (st *Channel, err error) {
 	body, err := s.RequestWithBucketID("PATCH", EndpointChannel(threadID), data, EndpointChannel(threadID))
 	if err != nil {
 		return
--- a/restapi.go
+++ b/restapi.go
@@ -2650,18 +2650,6 @@ type ThreadsResponseBody struct {
 	HasMore bool `json:"has_more"`
 }
 
-// ActiveThreads returns all active threads in the channel
-// GET /channels/{channel.id}/threads/active
-func (s *Session) ActiveThreads(channelID string) (st *ThreadsResponseBody, err error) {
-	body, err := s.RequestWithBucketID("GET", EndpointChannelActiveThreads(channelID), nil, EndpointThreadMembers(""))
-	if err != nil {
-		return
-	}
-
-	err = unmarshal(body, &st)
-	return
-}
-
 // PublicArchivedThreads returns archived threads in the channel that are public
 // GET /channels/{channel.id}/threads/archived/public
 //
@@ -2753,15 +2741,6 @@ func (s *Session) JoinedPrivateArchivedThreads(channelID string, before *time.Ti
 	return
 }
 
-// ThreadEdit edits the given thread
-// threadID  : The ID of a Thread
-// name       : The new name to assign the thread.
-func (s *Session) ThreadEdit(channelID string, name string) (*Channel, error) {
-	return s.ThreadEditComplex(channelID, &ThreadEditData{
-		Name: name,
-	})
-}
-
 // ThreadEditComplex edits an existing thread, replacing the parameters entirely with ThreadEdit struct
 // threadID  : The ID of a Thread
 // data          : The thread struct to send
@@ -2775,8 +2754,8 @@ func (s *Session) ThreadEditComplex(threadID string, data *ThreadEditData) (st *
 	return
 }
 
-// ListActiveThreadsResponseBody is the returned data from ListActiveThreads
-type ListActiveThreadsResponseBody struct {
+// listActiveThreadsResponseBody is the returned data from ListActiveThreads
+type listActiveThreadsResponseBody struct {
 	Threads []*Channel      `json:"threads"`
 	Members []*ThreadMember `json:"members"`
 }
@@ -2784,12 +2763,21 @@ type ListActiveThreadsResponseBody struct {
 // ListActiveThreads returns all active threads in the guild,
 // including public and private threads.
 // Threads are ordered by their `id`, in descending order.
-func (s *Session) ListActiveThreads(guildID string) (st *ListActiveThreadsResponseBody, err error) {
+func (s *Session) ListActiveThreads(guildID string) (threads []*Channel, members []*ThreadMember, err error) {
 	body, err := s.RequestWithBucketID("GET", EndpointGuildActiveThreads(guildID), nil, EndpointGuildActiveThreads(guildID))
 	if err != nil {
 		return
 	}
 
-	err = unmarshal(body, &st)
+	response := &listActiveThreadsResponseBody{}
+
+	err = unmarshal(body, &response)
+	if err != nil {
+		return
+	}
+
+	threads = response.Threads
+	members = response.Members
+
 	return
 }
--- a/restapi.go
+++ b/restapi.go
@@ -2125,21 +2125,11 @@ func (s *Session) WebhookDeleteWithToken(webhookID, token string) (st *Webhook, 
 // webhookID: The ID of a webhook.
 // token    : The auth token for the webhook
 // wait     : Waits for server confirmation of message send and ensures that the return struct is populated (it is nil otherwise)
-func (s *Session) WebhookExecute(webhookID, token string, wait bool, threadID string, data *WebhookParams) (st *Message, err error) {
+func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *WebhookParams) (st *Message, err error) {
 	uri := EndpointWebhookToken(webhookID, token)
 
-	v := url.Values{}
-
 	if wait {
-		v.Set("wait", "true")
-	}
-
-	if threadID != "" {
-		v.Set("thread_id", threadID)
-	}
-
-	if len(v) > 0 {
-		uri += "?" + v.Encode()
+		uri += "?wait=true"
 	}
 
 	var response []byte
@@ -2557,7 +2547,7 @@ func (s *Session) InteractionResponseDelete(appID string, interaction *Interacti
 // wait        : Waits for server confirmation of message send and ensures that the return struct is populated (it is nil otherwise)
 // data        : Data of the message to send.
 func (s *Session) FollowupMessageCreate(appID string, interaction *Interaction, wait bool, data *WebhookParams) (*Message, error) {
-	return s.WebhookExecute(appID, interaction.Token, wait, "", data)
+	return s.WebhookExecute(appID, interaction.Token, wait, data)
 }
 
 // FollowupMessageEdit edits a followup message of an interaction.
--- a/restapi.go
+++ b/restapi.go
@@ -2587,9 +2587,14 @@ func (s *Session) FollowupMessageDelete(appID string, interaction *Interaction, 
 //////////  THREADS  //////////
 //////////////////////////////
 
-// StartPublicThread creates a new public thread from an existing message.
+// StartThreadWithMessage creates a new public thread from an existing message.
+// When called on a ChannelTypeGuildText channel, creates a ChannelTypeGuildPublicThread.
+// When called on a ChannelTypeGuildNews channel, creates a ChannelTypeGuildNewsThread.
+// The id of the created thread will be the same as the id of the message,
+// and as such a message can only have a single thread created from it.
+//
 // POST /channels/{channel.id}/messages/{message.id}/threads
-func (s *Session) StartPublicThread(channelID, messageID string, data *ThreadCreateData) (st *Channel, err error) {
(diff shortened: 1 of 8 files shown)


Pedro-Pessoa:
I think that's everything.
Waiting on feedbacks

Pedro-Pessoa:
Updated the code to be in line with these changes on the threads docs: https://github.com/discord/discord-api-docs/commit/663dc595e06fa809ea96ac85cebeb1ec01e7765e

Pedro-Pessoa:
# Breaking change!

Webhooks now have an optional `ThreadID` field. Docs: https://github.com/discord/discord-api-docs/commit/cd0cae871e593876098ac8baf85d5f6403cd3d2c

Because of the way `WebhookExecute` is currently written, this is a breaking change. 
Updated the code to be in line with discord.

Lemme know if there should be another/better approach to this.

belak:
It looks like this is being pushed out publicly now - what would it take to get this merged?

Pedro-Pessoa:
There were a few changes to the discord API since I last updated this PR (nothing major, though). This is why this is still a draft. 
I also need to merge the new changes from the upstream. 

I've been busy lately, that's why I haven't finished this. But I will finish and undraft this PR later this week.

Pedro-Pessoa:
I think this is ready for review now. @CarsonHoffman

state.go
```diff
@@ -557,6 +560,98 @@ func (s *State) Channel(channelID string) (*Channel, error) {
...
+	thread, ok := s.channelMap[t.ID]
+	if ok {
+		for _, memberID := range t.RemovedMembersIDs {
+			if memberID == s.User.ID {
+				thread.Member = nil
+				break
+			}
+		}
+	}
+
```
kovetskiy, line 602: _I'm just lurking around since I'm waiting for this feature in the upstream and decided to give the PR a look._

Should not the method take into account `AddedMembers` field too? The documentation of the method says it `updates the member object of the thread` but it only takes care of the case when `the user was removed`.

Pedro-Pessoa: I think you are right! Will push an update. Let me know what you think.

restapi.go
```diff
@@ -2552,3 +2562,234 @@ func (s *Session) FollowupMessageEdit(appID string, interaction *Interaction, me
...
+}
+
+// ThreadEdit edits the given thread
+// threadID  : The ID of a Thread
+// name       : The new name to assign the thread.
+func (s *Session) ThreadEdit(channelID string, name string) (*Channel, error) {
+	return s.ThreadEditComplex(channelID, &ThreadEditData{
+		Name: name,
+	})
+}
```
CarsonHoffman, line 2763: Given that `ThreadEditComplex` clears out all properties not specified, I'm not sure that having this function as is here makes a ton of sense. It seems to me like it would quite quickly lead to unexpected behavior and not really have much use.

structs.go
```diff
@@ -298,13 +306,106 @@ type Channel struct {
...
+	ArchiveTimestamp    Timestamp       `json:"archive_timestamp"`
+	Locked              bool            `json:"locked,omitempty"`
+}
+
+// ThreadMember is used to indicate whether a user has joined
+// a thread or not.
+type ThreadMember struct {
+	ID            string    `json:"id,omitempty"`
+	UserID        string    `json:"user_id,omitempty"`
+	JoinTimeStamp Timestamp `json:"join_timestamp"`
```
CarsonHoffman, line 358: Should probably stick to `JoinTimestamp` to match `ArchiveTimestamp`.

structs.go
```diff
@@ -298,13 +306,106 @@ type Channel struct {
...
+type ThreadCreateData struct {
+	// 2-100 character channel name
+	Name string `json:"name"`
+
+	// Duration in minutes to automatically archive the thread
+	// after recent activity.
+	AutoArchiveDuration ArchiveDuration `json:"auto_archive_duration"`
+
+	// Defaults to `PRIVATE_THREAD` in order to match the behavior
+	// when thread documentation was first published.
```
CarsonHoffman, line 372: I'm not sure that this comment is really relevant here given that we don't specify `omitempty`; there's no way for the "default" behavior on the API side to take effect. Perhaps having `omitempty` might make sense?

restapi.go
```diff
@@ -2552,3 +2562,234 @@ func (s *Session) FollowupMessageEdit(appID string, interaction *Interaction, me
...
+// GET /channels/{channel.id}/threads/active
+func (s *Session) ActiveThreads(channelID string) (st *ThreadsResponseBody, err error) {
+	body, err := s.RequestWithBucketID("GET", EndpointChannelActiveThreads(channelID), nil, EndpointThreadMembers(""))
+	if err != nil {
+		return
+	}
+
+	err = unmarshal(body, &st)
+	return
+}
```
CarsonHoffman, line 2663: It looks like this endpoint [is already deprecated](https://discord.com/developers/docs/resources/channel#list-active-threads). We might just want to move to the guild active threads endpoint to begin with; that one also has a more reasonable response structure, such that we could return a slice of threads and a slice of members rather than a custom struct specifically for one function.

structs.go
```diff
@@ -298,13 +306,106 @@ type Channel struct {
 
 	// ApplicationID of the DM creator Zeroed if guild channel or not a bot user
 	ApplicationID string `json:"application_id"`
+
+	// Voice region ID for the voice channel, automatic when set to null
+	RTCRegion *string `json:"rtc_region"`
```
CarsonHoffman, line 311: It looks like this field is replacing `Region`, so we should probably remove that one while we're at it.

CarsonHoffman (changes requested):

restapi.go
```diff
@@ -2552,3 +2562,234 @@ func (s *Session) FollowupMessageEdit(appID string, interaction *Interaction, me
...
+// ListActiveThreadsResponseBody is the returned data from ListActiveThreads
+type ListActiveThreadsResponseBody struct {
+	Threads []*Channel      `json:"threads"`
+	Members []*ThreadMember `json:"members"`
+}
+
+// ListActiveThreads returns all active threads in the guild,
+// including public and private threads.
+// Threads are ordered by their `id`, in descending order.
+func (s *Session) ListActiveThreads(guildID string) (st *ListActiveThreadsResponseBody, err error) {
```
CarsonHoffman, line 2787: I would just return the threads and members as two return values, rather than having `ListActiveThreadsResponseBody` exported just for this function.

Pedro-Pessoa:
@CarsonHoffman I think that is everything you requested.
Let me know if I missed anything or if I did something wrongly please.

restapi.go
```diff
@@ -2125,21 +2125,11 @@ func (s *Session) WebhookDeleteWithToken(webhookID, token string) (st *Webhook,
...
 
-	v := url.Values{}
-
 	if wait {
-		v.Set("wait", "true")
-	}
-
-	if threadID != "" {
-		v.Set("thread_id", threadID)
-	}
```
CarsonHoffman: We still need to add the thread ID in, as it's not part of the body; it's a query parameter.

webhook.go
```diff
@@ -35,6 +36,7 @@ type WebhookParams struct {
 	Components      []MessageComponent      `json:"components"`
 	Embeds          []*MessageEmbed         `json:"embeds,omitempty"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
+	ThreadID        string                  `json:"thread_id,omitempty"`
```
CarsonHoffman, line 39: Given that the thread ID isn't part of the body, we should just have `` `json:"-"` `` here.

CarsonHoffman (changes requested):

state.go
```diff
@@ -557,6 +560,105 @@ func (s *State) Channel(channelID string) (*Channel, error) {
...
+	if s == nil {
+		return ErrNilState
+	}
+
+	g, err := s.Guild(t.GuildID)
+	if err != nil {
+		return err
+	}
+
+	if len(t.ChannelIDs) > 0 {
```
CarsonHoffman, line 626: The docs note that `channel_ids` could be omitted, in which case all channels for the guild are being synced. Should we really be doing nothing and returning `ErrStateNotFound` if this is the case?

state.go
```diff
@@ -557,6 +560,105 @@ func (s *State) Channel(channelID string) (*Channel, error) {
...
+		s.Lock()
+		defer s.Unlock()
+
+		for _, parentChannelID := range t.ChannelIDs {
+			for _, stateChannel := range s.channelMap {
```
CarsonHoffman, line 631: This seems *really* nasty—it's O(nm), where n is the list of **all channels the bot has access to** and m is the number of channels being synced. e.g. if the bot can see 2000 guilds, each has 20 channels, and 10 channels were  synced (these numbers seem somewhat reasonable for a larger bot), we just looped **four hundred thousand times**. Perhaps we could limit the search to only the indicated guild's channels?

Pedro-Pessoa:
Updated it @CarsonHoffman 
Let me know what you think!

state.go
```diff
@@ -557,6 +560,114 @@ func (s *State) Channel(channelID string) (*Channel, error) {
...
+		for _, parentChannelID := range t.ChannelIDs {
+			if c.ParentID == parentChannelID {
+				delete(s.channelMap, c.ID)
+				continue OUTER
+			}
+		}
+
+		g.Channels[index] = c
+		index++
+	}
```
CarsonHoffman, line 653: A labeled loop with three different (and similar) `continue` points seems really difficult to reason about to me. I would suggest inverting the loop—first loop over the thread sync event's channels, and then do the guild channels in the inner loop, since you'll then be starting the guild channels loop anew each time and won't need to worry about mutating in place.

CarsonHoffman (changes requested):

Pedro-Pessoa:
Updated @CarsonHoffman

state.go
```diff
@@ -557,6 +560,117 @@ func (s *State) Channel(channelID string) (*Channel, error) {
...
+
+		for _, member := range t.AddedMembers {
+			if member.UserID == s.User.ID {
+				thread.Member = member
+				break
+			}
+		}
+	}
+
+	return ErrStateNotFound
```
CarsonHoffman, line 610: It seems to me like this function always returns `ErrStateNotFound` if it makes it past the initial `s == nil` check. I'd probably invert the `if` statement to `if !ok` to return early and keep the happy path left (in addition to above in `ThreadMemberUpdate`).

CarsonHoffman (changes requested):

state.go
Pedro-Pessoa: Updated

restapi.go
```diff
@@ -2566,3 +2576,222 @@ func (s *Session) FollowupMessageEdit(appID string, interaction *Interaction, me
...
+// GET /channels/{channel.id}/threads-members
+func (s *Session) ThreadMembers(channelID string) (st []*ThreadMember, err error) {
+	body, err := s.RequestWithBucketID("GET", EndpointThreadMembers(channelID), nil, EndpointThreadMembers(""))
+	if err != nil {
+		return
+	}
+
+	err = unmarshal(body, &st)
+	return
+}
```
CarsonHoffman, line 2654: It doesn't look like this function works at all:
```go
panic: HTTP 404 Not Found, {"message": "404: Not Found", "code": 0}
```
It seems like it's calling to `/thread-members/` rather than `/thread-members`. The fact that `EndpointThreadMembers` doesn't take in the user ID as a parameter seems like a bit of a smell. Having two separate endpoint functions and taking in the user ID as a parameter in one (and consequently updating `JoinThread`, `AddUserToThread`, etc. might be a good idea.

CarsonHoffman (changes requested):

restapi.go
Pedro-Pessoa: Updated

restapi.go
```diff
@@ -2566,3 +2576,222 @@ func (s *Session) FollowupMessageEdit(appID string, interaction *Interaction, me
...
+
+// StartThreadWithMessage creates a new public thread from an existing message.
+// When called on a ChannelTypeGuildText channel, creates a ChannelTypeGuildPublicThread.
+// When called on a ChannelTypeGuildNews channel, creates a ChannelTypeGuildNewsThread.
+// The id of the created thread will be the same as the id of the message,
+// and as such a message can only have a single thread created from it.
+//
+// POST /channels/{channel.id}/messages/{message.id}/threads
+func (s *Session) StartThreadWithMessage(channelID, messageID string, data *ThreadCreateData) (st *Channel, err error) {
+	body, err := s.RequestWithBucketID("POST", EndpointChannelMessageThreads(channelID, messageID), data, EndpointChannelMessageThreads("", ""))
```
CarsonHoffman, line 2592: Given the following passage from the rate limits documentation:
> Additionally, rate limits take into account major parameters in the URL. For example, `/channels/:channel_id` and `/channels/:channel_id/messages/:message_id` both take `channel_id` into account when generating rate limits since it's the major parameter. Currently, the only major parameters are `channel_id`, `guild_id`, and `webhook_id + webhook_token`.

It seems to me like the bucket ID should include the channel ID here, since each different channel will be in a different bucket. I think that this also applies to `StartThreadWithoutMessage`, `JoinThread`, `AddUserToThread`, `LeaveThread`, `RemoveUserFromThread`, `ThreadMembers`, `PublicArchivedThreads`, `PrivateArchivedThreads`, and `JoinedPrivateArchivedThreads` (`ThreadEditComplex` and `ListActiveThreads` already do this).

CarsonHoffman (changes requested):
I think that things are nearing ready to go here! I noticed that we seem to be limiting many of the thread API calls too strictly by not taking into account the channel ID as a major parameter, but otherwise things are looking great.

state.go
```diff
@@ -557,6 +560,98 @@ func (s *State) Channel(channelID string) (*Channel, error) {
...
+		for _, memberID := range t.RemovedMembersIDs {
+			if memberID == s.User.ID {
+				thread.Member = nil
+				break
+			}
+		}
+	}
+
+	return ErrStateNotFound
+}
```
CarsonHoffman: This appears to be copied from https://github.com/bwmarrin/discordgo/pull/982/files/ecd5d0d76e298de23c533d7e4f1f8c026930fb07#diff-320959764bfa2277e0bdd4eaa741f02891d123c42150d4ccb76de0bdbf6e591aR582-R603, and retains the same issue of always returning `ErrStateNotFound`.

state.go
```diff
@@ -557,6 +560,98 @@ func (s *State) Channel(channelID string) (*Channel, error) {
...
+	if ok {
+		s.Lock()
+		defer s.Unlock()
+
+		channel.Member = t
+		return nil
+	}
+
+	return ErrStateNotFound
+}
```
CarsonHoffman: This appears to be copied from https://github.com/bwmarrin/discordgo/pull/982/files/ecd5d0d76e298de23c533d7e4f1f8c026930fb07#diff-320959764bfa2277e0bdd4eaa741f02891d123c42150d4ccb76de0bdbf6e591aR564-R580, and retains the same bug of accessing `s.channelMap` without any synchronization.

state.go
```diff
@@ -557,6 +560,98 @@ func (s *State) Channel(channelID string) (*Channel, error) {
...
+			if ok {
+				channel.Member = member
+			}
+		}
+
+		return nil
+	}
+
+	return ErrStateNotFound
+}
```
CarsonHoffman: This appears to be copied from https://github.com/bwmarrin/discordgo/pull/982/files/ecd5d0d76e298de23c533d7e4f1f8c026930fb07#diff-320959764bfa2277e0bdd4eaa741f02891d123c42150d4ccb76de0bdbf6e591aR605-R659, and retains the same very unusual structure of the nested loops (it's so non-straightforward that I cannot imagine two people coming to this independently), the same misunderstanding of what the case of `len(t.ChannelIDs) == 0` means, and the same return of `ErrStateNotFound` when that error _really_ doesn't make a ton of sense in this context.

CarsonHoffman (commented):
I had another chance to look more closely at the code here, and I am really concerned about what I see. It appears that much of the gateway handling code that was introduced in this commit is copied from #982's implementation (it was committed on July 29th there and August 2nd here). I listed three examples in this review; it seems like you attempted to obfuscate this by removing comments and changing `!=` to `>`, but the code retains almost the exact same structure, the same logic, _and the same bugs_ as the implementation in #982. Are you claiming that the work in this PR is your own? If you were looking to credit the author of #982, **the time was when you added that code, not after** (and I'm still not sure that this would be an acceptable route given another active PR, but at least it's not direct stealing). I hope that I'm somehow misinterpreting the situation here, but taking code from others' active PRs and trying to represent it as your own is in no way acceptable behavior.

Pedro-Pessoa:
> I had another chance to look more closely at the code here, and I am really concerned about what I see. It appears that much of the gateway handling code that was introduced in this commit is copied from #982's implementation (it was committed on July 29th there and August 2nd here). I listed three examples in this review; it seems like you attempted to obfuscate this by removing comments and changing `!=` to `>`, but the code retains almost the exact same structure, the same logic, _and the same bugs_ as the implementation in #982. Are you claiming that the work in this PR is your own? If you were looking to credit the author of #982, **the time was when you added that code, not after** (and I'm still not sure that this would be an acceptable route given another active PR, but at least it's not direct stealing). I hope that I'm somehow misinterpreting the situation here, but taking code from others' active PRs and trying to represent it as your own is in no way acceptable behavior.

It is definitely not a copied work. This PR was opened way before that other one. I did take a look at that, to see if I missed something. And I missed the state work and the new webhook type. 
Still it doesn't mean I copied anything from that. We had similar approaches.
Yes, I misinterpreted the threadlistsync handler because of how that PR handled it. Should have referred to the docs instead of that PR to work on it. 

I also should have mentioned that I added these changes because I saw that PR. I apologize for that, it was my bad. 
But code is not a copy. And it is definitely not stealing. 

There are bugs there that are not on this PR. And vice versa. 

A lot of code there is extremely similar to code here too.
Even some comments there are the same as mine.

CarsonHoffman:
I'm not sure that "we had similar approaches" really does a good job of explaining the situation here. Other than changing one-line map-access-in-`if`-statements to two-line ones (which consistently changed in all three of them), removing comments and using the `IsThread` method, the three functions are essentially exactly the same character-for-character (`ThreadMembersUpdate` and `ThreadMemberUpdate` are). These are not methods where there's one defined pattern of doing things in the library, like a REST API method might have, or property changes directly corresponding to Discord's objects; there are choices to be made in how things are done. They retain the same bugs in all cases (including a function always returning the wrong thing). They have the same nested loop structure with exactly the same variable names; `stateChannel.ParentID == parentChannelID` in line 625 means that this whole nested loop is essentially just the same as `s.channelMap[parentChannelID]`, but with more unnecessary looping. Roundabout ways of doing things like this don't coincidentally match, especially considering the other two methods.

> This PR was opened way before that other one.

Sure. I'm talking specifically about the state management code, which was added after that one's, as you acknowledge.

> There are bugs there that are not on this PR.

Not in the state management code.

> And vice versa.

Sure, but I'm only referring to the state management code.

> Even some comments there are the same as mine.

I fail to see anything significant that isn't either a natural way of describing things or from Discord's documentation.

I unfortunately see no path forward here. I cannot in good conscience accept code that began its journey as copied code, even if it is no longer directly in the final PR. It would be crediting you with someone else's work. I'd recommend really trying to reflect on this and the impact it might have on others if you're planning on contributing to other open-source projects in the future.

PR #924, opened 2021-05-03
Update unmarshal function to wrap ErrJSONUnmarshal. Requires at least Go 1.13.

### LINKED ISSUE #923: Function unmarshal swallows specific error message
While recently troubleshooting an issue, I found that it was very useful to include the specific error message in the error returned by the function `unmarshal(data []byte, v interface{}) error` in `restapi.go`.

Currently, we swallow the specific error message:

```go
func unmarshal(data []byte, v interface{}) error {
	err := json.Unmarshal(data, v)
	if err != nil {
                // err is effectively lost here
		return ErrJSONUnmarshal
	}

	return nil
}
```

Instead, we can wrap `ErrJSONUnmarshal` and include the specific error message:

```go
func unmarshal(data []byte, v interface{}) error {
	err := json.Unmarshal(data, v)
	if err != nil {
		return fmt.Errorf("%w: %s", ErrJSONUnmarshal, err)
	}

	return nil
}
```

This error wrapping functionality was introduced to the standard library in Go 1.13, so I believe we should update the `go.mod` file as well. Background: [https://blog.golang.org/go1.13-errors](https://blog.golang.org/go1.13-errors)

```go

go 1.13

require (
	github.com/gorilla/websocket v1.4.2
	golang.org/x/crypto v0.0.0-20210421170649-83a5a9bb288b
)
```

Fortunately, the CI process only tests 1.13 and later so in a sense we already no longer support Go versions less than 1.13. While we're working on this, and since Go 1.16 has been out for awhile, might as well update the CI to test with 1.16 too:

```yml
language: go
go:
    - 1.13.x
    - 1.14.x
    - 1.15.x
    - 1.16.x
env:
    - GO111MODULE=on
install:
    - go get github.com/bwmarrin/discordgo
    - go get -v .
    - go get -v golang.org/x/lint/golint
script:
    - diff <(gofmt -d .) <(echo -n)
    - go vet -x ./...
    - golint -set_exit_status ./...
    - go test -v -race ./...
```

ewohltman:
Update unmarshal function to wrap ErrJSONUnmarshal and return specific error message.

Requires at least Go 1.13.

Resolves #923

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ go:
     - 1.13.x
     - 1.14.x
     - 1.15.x
+    - 1.16.x
 env:
     - GO111MODULE=on
 install:
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/bwmarrin/discordgo
 
+go 1.13
+
 require (
 	github.com/gorilla/websocket v1.4.2
-	golang.org/x/crypto v0.0.0-20181030102418-4d3f4d9ffa16
+	golang.org/x/crypto v0.0.0-20210421170649-83a5a9bb288b
 )
-
-go 1.10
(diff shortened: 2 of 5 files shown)


jolheiser (approved):
LGTM, this seems nice for better troubleshooting and more accurate error responses. 👍

CarsonHoffman:
Just for future reference: I considered whether it might make more sense to go all the way and expose the underlying JSON error via wrapping rather than just including its error text, but the following passage from [Working with Errors in Go 1.13](https://blog.golang.org/go1.13-errors) is enlightening here:
> It’s important to remember that whether you wrap or not, the error text will be the same. A person trying to understand the error will have the same information either way; the choice to wrap is about whether to give programs additional information so they can make more informed decisions, or to withhold that information to preserve an abstraction layer.

It doesn't seem to me like a program using an API wrapper would have any particular course of action based on a transit error, and thus this solution seems reasonable to me. Perhaps someone looking to extract that information in the future will see this—my point of putting my reasoning here—and it might be reasonable to revisit this in the future, but this is a good change now, so thank you for the change!

PR #925, opened 2021-05-05
Fix all the typos that goreportcard detected.

courtier:
Fixed all the misspells Go Report Card found except "themselvs" which is not in the code anymore.


bwmarrin:
Thanks @courtier

PR #927, opened 2021-05-10
Add a toggle for ephemeral messages to slash commands response

Lucaskyy:
Just an ease-of-use feature, does not break the API because it defaults to `false`.

I can also change `1 << 6` to `64`, if that's possible.

This maybe could've been done better, but this is how I implemented it for myself locally to test.

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -252,6 +252,8 @@ type InteractionApplicationCommandResponseData struct {
 	Content         string                  `json:"content,omitempty"`
 	Embeds          []*MessageEmbed         `json:"embeds,omitempty"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
+	
+	Ephemeral       bool
 
 	// NOTE: Undocumented feature, be careful with it.
 	Flags uint64 `json:"flags,omitempty"`
--- a/restapi.go
+++ b/restapi.go
@@ -2533,6 +2533,10 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 func (s *Session) InteractionRespond(interaction *Interaction, resp *InteractionResponse) error {
 	endpoint := EndpointInteractionResponse(interaction.ID, interaction.Token)
 
+	if resp.Data.Ephemeral {
+		resp.Data.Flags = 1 << 6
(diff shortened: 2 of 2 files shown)


jolheiser:
I would personally prefer some variables for the flags (of which I only know of ephemeral at the moment) rather than a bool that ends up overriding the flags.

restapi.go
```diff
@@ -2533,6 +2533,12 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 func (s *Session) InteractionRespond(interaction *Interaction, resp *InteractionResponse) error {
 	endpoint := EndpointInteractionResponse(interaction.ID, interaction.Token)
 
+	if resp.Data != nil {
+		if resp.Data.Ephemeral {
+			resp.Data.Flags = 1 << 6
```
plally, line 2538: i think you can have ephemeral with other types of flags in theory. (although i don't think any other flags are supported currently) im thinking the proper way to do this might be `resp.Data.Flags = resp.Data.Flags | (1 << 6)`

plally (commented):
> I would personally prefer some variables for the flags (of which I only know of ephemeral at the moment) rather than a bool that ends up overriding the flags.
Yeah I agree. if you want an easy way to toggle ephemeral would having a method to toggle the ephemeral flag might be better than a field on the struct.

also all the flag values are documented here if you want to add them as constants (if they arent constants already)
https://discord.com/developers/docs/resources/channel#message-object-message-flags

interactions.go
```diff
@@ -252,6 +252,8 @@ type InteractionApplicationCommandResponseData struct {
 	Content         string                  `json:"content,omitempty"`
 	Embeds          []*MessageEmbed         `json:"embeds,omitempty"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
+	
+	Ephemeral       bool
```
FedorLap2006, line 256: I think would be better to add `json:"-"` so it doesn't get marshaled

FedorLap2006:
Hi. Sorry review took a long time.
We try to keep DiscordGo 1:1 mapping of Discord API, therefore we try to avoid helper fields & functions.

PR #930, opened 2021-05-12
Remove deprecated response types

plally:
~~These types were not registering as deprecated in GoLand. I updated them so that they would be recognized as deprecated.~~
Remove deprecated types as stated by the previous comments.

Relevant discord documentation: https://discord.com/developers/docs/interactions/slash-commands#interaction-response-interactioncallbacktype

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -226,12 +226,11 @@ type InteractionResponseType uint8
 // Interaction response types.
 const (
 	// InteractionResponsePong is for ACK ping event.
-	InteractionResponsePong = InteractionResponseType(iota + 1)
+	InteractionResponsePong InteractionResponseType = 1
 	// InteractionResponseChannelMessageWithSource is for responding with a message, showing the user's input.
-	InteractionResponseChannelMessageWithSource
+	InteractionResponseChannelMessageWithSource InteractionResponseType = 4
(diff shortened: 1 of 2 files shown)


interactions.go
```diff
@@ -227,12 +227,6 @@ type InteractionResponseType uint8
 const (
 	// InteractionResponsePong is for ACK ping event.
 	InteractionResponsePong = InteractionResponseType(iota + 1)
-	// InteractionResponseAcknowledge is for ACK a command without sending a message, eating the user's input.
-	// NOTE: this type is being imminently deprecated, and **will be removed when this occurs.**
-	InteractionResponseAcknowledge
-	// InteractionResponseChannelMessage is for responding with a message, eating the user's input.
-	// NOTE: this type is being imminently deprecated, and **will be removed when this occurs.**
-	InteractionResponseChannelMessage
```
jolheiser, line 235: Removing these outright will mess up the `iota`
Consider replacing with `_` and comments if needed

jolheiser (changes requested):

interactions.go
plally: is using the iota even worth it for this anymore. could just actually write `1, 4, 5`?

jolheiser: It sounds like the consensus in Discord is to hard-code them. 😊

jolheiser (approved):

PR #913, opened 2021-04-24
Webhook files

plally:
Webhook edits can now add files to messages.

https://discord.com/developers/docs/resources/webhook#edit-webhook-message

The code for sending files is nearly identical to the code in `WebhookExecute` and `ChannelMessageSendComplex`. Should I move the duplicate code in all 3 functions to a helper function?

PR #931, opened 2021-05-14
support allowing webhook edits with files, and responding to interactions with files

plally:
Webhook edits and as a result interaction followups support uploading files
https://discord.com/developers/docs/resources/webhook#edit-webhook-message

While not documented it is also possible to include files in slash command responses

This implements the same functionality as https://github.com/bwmarrin/discordgo/pull/913. and that pr may be closed if this one is approved.

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -327,7 +327,7 @@ const (
 // InteractionResponse represents a response for an interaction event.
 type InteractionResponse struct {
 	Type  InteractionResponseType  `json:"type,omitempty"`
-  Files []*File                 `json:"-"`
+	Files []*File                  `json:"-"`
 	Data  *InteractionResponseData `json:"data,omitempty"`
 }
 
--- a/interactions.go
+++ b/interactions.go
@@ -242,8 +242,9 @@ const (
 
 // InteractionResponse represents a response for an interaction event.
 type InteractionResponse struct {
-	Type InteractionResponseType                    `json:"type,omitempty"`
-	Data *InteractionApplicationCommandResponseData `json:"data,omitempty"`
+	Type  InteractionResponseType                    `json:"type,omitempty"`
+	Files []*File `json:"-"`
+	Data  *InteractionApplicationCommandResponseData `json:"data,omitempty"`
 }
 
 // InteractionApplicationCommandResponseData is response data for a slash command interaction.
@@ -252,7 +253,6 @@ type InteractionApplicationCommandResponseData struct {
 	Content         string                  `json:"content,omitempty"`
 	Embeds          []*MessageEmbed         `json:"embeds,omitempty"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
-	Files           []*File `json:"-"`
 
 	// NOTE: Undocumented feature, be careful with it.
 	Flags uint64 `json:"flags,omitempty"`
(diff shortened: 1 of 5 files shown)


CarsonHoffman:
Is there any particular reason that `EncodeWithFiles` is exported here? It seems to me like this is pretty firmly internal details, and exposing it in the library's interface might cause additional confusion around file handling (e.g. I can see someone calling this function to then use as the value of a file).

plally:
The function would be useful to anyone implementing an http interactions endpoint. Anyone who wants to respond with files to an http interactions request would need to do this.

restapi.go
```diff
@@ -2530,11 +2449,19 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 // appID       : The application ID.
 // interaction : Interaction instance.
 // resp        : Response message data.
-func (s *Session) InteractionRespond(interaction *Interaction, resp *InteractionResponse) error {
+func (s *Session) InteractionRespond(interaction *Interaction, resp *InteractionResponse) (err error) {
 	endpoint := EndpointInteractionResponse(interaction.ID, interaction.Token)
 
-	_, err := s.RequestWithBucketID("POST", endpoint, *resp, endpoint)
+	if len(resp.Files) > 0 {
+		contentType, body, err := EncodeWithFiles(interaction, resp.Files)
```
CarsonHoffman, line 2456: Are you sure that the first argument is meant to be `interaction`, not `resp`? I haven't been able to get responding to an interaction with a file working.

plally: You may be right, its been a bit since I wrote this. I'll do further testing later today and make sure responding with the callback works (and fix it if it does not) as most of my testing has been with an outgoing webhook instead of the gateway.

plally: changed interaction -> resp. and added a working example to examples/slash_commands.

CarsonHoffman:
Having given it another thought, since having the function exported seems like the way to go, I'm not sure that `EncodeWithFiles` is the best name for the function—"encode" is quite ambiguous, and the name of the function doesn't really describe what it does without the additional context of it being an HTTP body that contains JSON in a part. The first thought that comes to my mind would be something like `JSONBodyWithFiles`; do you have any opinions on this?

plally:
Im happy to rename the function to `JSONBodyWithFiles` or anything better anyone else thinks of. I agree EncodeWithFiles isn't great.

SirFist:
Not sure if you'd want it to fall under this PR or another, but this could be expanded to also cover uploading files when editing messages as they [also support it now](https://discord.com/developers/docs/resources/channel#edit-message). Everything also seems to support editing attachments too, which might be sensible to have covered alongside this functionality?

interactions.go
```diff
@@ -326,8 +326,9 @@ const (
 
 // InteractionResponse represents a response for an interaction event.
 type InteractionResponse struct {
-	Type InteractionResponseType  `json:"type,omitempty"`
-	Data *InteractionResponseData `json:"data,omitempty"`
+	Type  InteractionResponseType  `json:"type,omitempty"`
+	Files []*File                  `json:"-"`
```
CarsonHoffman, line 330: It seems to me like this field would be better situated inside `InteractionResponseData` next to `Content`, `Embeds`, etc. just like `WebhookEdit`.

CarsonHoffman (changes requested):

interactions.go
plally: My original reasoning was that Files aren't really part of the response structure, more just  things that get passed alongside the response. You could even return  just `{"type":4}` as the response if it contains files as well.

PR #856, opened 2020-12-24
Slash commands

FedorLap2006:
Here we go! Slash commands are _ready_ to use (almost).
So... I saw that everyone wants it, so I'm opening it now, not when I will complete it. 
Now it includes some example in `examples/` folder of using this new feature and basic API of slash commands (CRUD).

Well, enjoy!

PR #933, opened 2021-05-16
Interactions: the Buttons

FedorLap2006:
Hello everyone. Just finished implementing buttons functionality for DiscordGo. Looking forward a feedback.


interactions.go
```diff
@@ -63,17 +63,22 @@ type InteractionType uint8
 
 // Interaction types
 const (
-	InteractionPing = InteractionType(iota + 1)
+	InteractionPing InteractionType = iota + 1
 	InteractionApplicationCommand
```
FedorLap2006, line 67: Also wanted to mention. I came across these constants and thought, maybe we should change the their names like so ```suggestion
	PingInteraction InteractionType = iota + 1
	ApplicationCommandInteraction
```
What do you think?

File: components.go [hunk 2]
```diff
@@ -0,0 +1,97 @@
...
+// Component is a base interface for all components
+type Component interface {
+	json.Marshaler
+	Type() ComponentType
+}
+
+// ActionsRow is a container for components within one row.
+type ActionsRow struct {
+	Components []Component `json:"components"`
+}
```
FedorLap2006, line 25: I was reading the docs and it was clearly saying "Actions row", but maybe we should name it ActionsBar, or stick to the docs? :thinking: 
```suggestion
// ActionsRow is a container for components within one row.
type ActionsRow struct {
	Components []Component `json:"components"`
}
```

examples/components/main.go
```diff
@@ -0,0 +1,151 @@
...
+	s.AddHandler(func(s *discordgo.Session, i *discordgo.InteractionCreate) {
+		if i.Type == discordgo.InteractionApplicationCommand {
+			if i.Data.Name == "feedback" {
+				err := s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
+					Type: discordgo.InteractionResponseChannelMessageWithSource,
+					Data: &discordgo.InteractionResponseData{
+						Content: "Are you satisfied with Buttons?",
+						// Buttons and other components are specified in Components field.
+						Components: []discordgo.Component{
+							// ActionRow is a container of all buttons in the same raw.
```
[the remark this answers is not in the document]
FedorLap2006: Oh, you're right, my bad.

interactions.go
```diff
@@ -90,11 +95,16 @@ type Interaction struct {
 	Version int    `json:"version"`
 }
 
-// ApplicationCommandInteractionData contains data received in an interaction event.
-type ApplicationCommandInteractionData struct {
+// InteractionData contains data received from InteractionCreate event.
+type InteractionData struct {
```
Strum355, line 99: Dont change the name please. It is a breaking change to rename this

FedorLap2006: Alright, let's keep `ApplicationCommandInteractionData ` if the data isn't for not only for application commands. Besides, there was no release of Slash Commands... so this is not a breaking change. Those who wanted master branch were aware, that it may contain breaking changes.

interactions.go
CarsonHoffman: What's the reasoning behind this change (especially with `InteractionButton` conforming to the old pattern)?

FedorLap2006: Well, on my opinion, the new names looks and sounds better, plus this change will blend in perfectly with such enums as:

```go
const (
	// PrimaryButton is a button with blurple color.
	PrimaryButton ButtonStyle = iota + 1
	// SecondaryButton is a button with grey color.
	SecondaryButton
	// SuccessButton is a button with green color.
	SuccessButton
	// DangerButton is a button with red color.
	DangerButton
	// LinkButton is a special type of button which navigates to a URL. Has grey color.
	LinkButton
)
```

interactions.go
Strum355: > Besides, there was no release of Slash Commands... so this is not a breaking change. Those who wanted master branch were aware, that it may contain breaking changes.

Yup thats fair enough :+1:

FedorLap2006:
After some discussion about `InteractionData`. I'd say that we decided to separate data into different structures. And I personally decided to make it an interface. So, waiting for feedback!

interactions.go
```diff
@@ -63,17 +64,22 @@ type InteractionType uint8
 
 // Interaction types
 const (
-	InteractionPing = InteractionType(iota + 1)
+	InteractionPing InteractionType = iota + 1
 	InteractionApplicationCommand
+	InteractionButton
```
typpo, line 69: This should be `InteractionMessageComponent`, and you will have to check elsewhere that the Message Component type is `Button`.

FedorLap2006: > This should be `InteractionMessageComponent`, and you will have to check elsewhere that the Message Component type is `Button`.

Oh, alright.

FedorLap2006:
Hello everyone. It seems that documentation has been published and the buttons are ungated. I think It's time to mark this PR as ready for review. Well, enjoy the buttons!

interactions.go
```diff
@@ -232,18 +311,23 @@ const (
...
 
-// InteractionApplicationCommandResponseData is response data for a slash command interaction.
-type InteractionApplicationCommandResponseData struct {
-	TTS             bool                    `json:"tts,omitempty"`
-	Content         string                  `json:"content,omitempty"`
+// InteractionResponseData is response data for an interaction.
+type InteractionResponseData struct {
+	TTS             bool                    `json:"tts"`
+	Content         string                  `json:"content"`
+	Components      []MessageComponent      `json:"components,omitempty"`
```
zacharyburkett, line 330: Sorry, previous comment didn't link to the lines.

This is a breaking change, no? The name of a type has changed.

FedorLap2006: >This is a breaking change, no?

See [this](https://github.com/bwmarrin/discordgo/pull/933#discussion_r633618717) comment.

yyewolf:
You can add the "Component" feature to these endpoints :

https://discord.com/developers/docs/resources/webhook#edit-webhook-message
https://discord.com/developers/docs/resources/webhook#execute-webhook
https://discord.com/developers/docs/interactions/slash-commands#edit-original-interaction-response

They all support perfectly fine components objects and are needed to put buttons on a follow up message for example.

FedorLap2006:
> You can add the "Component" feature to these endpoints :
> 
> https://discord.com/developers/docs/resources/webhook#edit-webhook-message
> https://discord.com/developers/docs/resources/webhook#execute-webhook
> https://discord.com/developers/docs/interactions/slash-commands#edit-original-interaction-response
> 
> They all support perfectly fine components objects and are needed to put buttons on a follow up message for example.

Interesting, it's not documented. But if they support it, then okay, I will add components to these endpoints.

interactions.go
CarsonHoffman: I'm not sure I follow; it seems to me like the types of interactions should follow a consistent naming scheme, rather than all adhering to the buttons. I don't really see a reason to change this.

components.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"encoding/json"
+)
+
+// ComponentType is type of component.
+type ComponentType uint
+
+// MessageComponent types.
+const (
+	ActionsRowComponent ComponentType = iota + 1
```
CarsonHoffman, line 12: Please avoid introducing additional uses of `iota` for constants that depend on external values; it's better to list them out explicitly (even if we don't do this universally in the library at the moment). See https://github.com/bwmarrin/discordgo/commit/843c765ae3eeecaf8e9b808ea8e3da10ba600f91 for more reasoning.

components.go
```diff
@@ -0,0 +1,101 @@
...
+	return ActionsRowComponent
+}
+
+// ButtonStyle is style of button.
+type ButtonStyle uint
+
+// Button styles.
+const (
+	// PrimaryButton is a button with blurple color.
+	PrimaryButton ButtonStyle = iota + 1
```
CarsonHoffman, line 51: Same as above.

interactions.go
```diff
@@ -90,13 +96,86 @@ type Interaction struct {
...
+		return
+	}
+	return i.Data.(MessageComponentInteractionData)
+}
+
+// ApplicationCommandData is helper function to convert InteractionData to ApplicationCommandInteractionData.
+func (i Interaction) ApplicationCommandData() (data ApplicationCommandInteractionData) {
+	if i.Type != InteractionApplicationCommand {
+		return
+	}
```
CarsonHoffman, line 147: Same as above.

interactions.go
```diff
@@ -90,13 +96,86 @@ type Interaction struct {
...
+		i.Data = v
+	}
+	return nil
+}
+
+// MessageComponentData is helper function to convert InteractionData to MessageComponentInteractionData.
+func (i Interaction) MessageComponentData() (data MessageComponentInteractionData) {
+	if i.Type != InteractionMessageComponent {
+		return
+	}
```
CarsonHoffman, line 139: I'm quite honestly not sure if this check is necessary. IMO, this pattern expects the developer to check that the interaction contains a component before calling MessageComponentData, so if they fail to do so, it's indicative of a developer error; this is what panics are for, and luckily the type assertion already panics in this situation. It could be worth a note in the comment about this behavior, but I think that this is the right way to go; returning the zero value is ripe for bugs.

CarsonHoffman (changes requested):

CarsonHoffman:
Do you think supporting components in the [webhook endpoints](https://github.com/bwmarrin/discordgo/pull/933#issuecomment-850614011) is reasonable for this PR, or would it be better included in a separate one?

vertex-admin:
> Do you think supporting components in the [webhook endpoints](https://github.com/bwmarrin/discordgo/pull/933#issuecomment-850614011) is reasonable for this PR, or would it be better included in a separate one?

Consider making a new PR for webhooks; we should try to update the library asap when major API changes are made. Webhooks can be phased out and waited on for an extra couple of days, but we need to catch up with other libraries.

SirFist:
Unless I am missing something, sending something like `Components: []discordgo.MessageComponent{}` back is needed to remove buttons when updating a message with the `InteractionResponseUpdateMessage` type, so it presumably shouldn't have `omitempty`?

FedorLap2006:
> Unless I am missing something, sending something like `Components: []discordgo.MessageComponent{}` back is needed to remove buttons when updating a message with the `InteractionResponseUpdateMessage` type, so it presumably shouldn't have `omitempty`?

I'm sorry, this issue will be fixed in next commits. I just noticed it.

FedorLap2006:
> > Do you think supporting components in the [webhook endpoints](https://github.com/bwmarrin/discordgo/pull/933#issuecomment-850614011) is reasonable for this PR, or would it be better included in a separate one?
> 
> Consider making a new PR for webhooks; we should try to update the library asap when major API changes are made. Webhooks can be phased out and waited on for an extra couple of days, but we need to catch up with other libraries.

Well, we can, but we can't: this PR depends on this functionality.

FedorLap2006:
> Do you think supporting components in the [webhook endpoints](https://github.com/bwmarrin/discordgo/pull/933#issuecomment-850614011) is reasonable for this PR, or would it be better included in a separate one?

Yes, this is reasonable. This PR covers the full buttons functionality, so it covers the webhooks as well. Because webhooks are part of interactions.

vertex-admin (changes requested):
Add a Components field to the MessageEdit struct in message.go, as we can edit the buttons on a Message without needing an interaction: https://discord.com/developers/docs/resources/channel#edit-message
```
type MessageEdit struct {
	Content         *string                 `json:"content,omitempty"`
	Embed           *MessageEmbed           `json:"embed,omitempty"`

	ID      string
	Channel string
}
```

FedorLap2006:
Well, probably everything is finished (except webhooks API), so I think this should be merged soon.

message.go
```diff
@@ -80,6 +80,9 @@ type Message struct {
 	// A list of attachments present in the message.
 	Attachments []*MessageAttachment `json:"attachments"`
 
+	// A list of components attached to the message.
+	Components []UnmarshableMessageComponent `json:"components"`
```
CarsonHoffman, line 84: Instead of having this field be of type `UnmarshableMessageComponent`, it might be better to keep this type as an unexported internal implementation detail, then use an `UnmarshalJSON` function on `Message`; this doesn't really seem like something that we'd want to expose to the library's interface.

CarsonHoffman (changes requested):

components.go
```diff
@@ -0,0 +1,135 @@
...
+
+// MessageComponent is a base interface for all message components.
+type MessageComponent interface {
+	json.Marshaler
+	Type() ComponentType
+}
+
+// UnmarshableMessageComponent a helper for components which need to be unmarshaled (like the ones in message object).
+// Since interfaces can't be unmarshaled this exists.
+type UnmarshableMessageComponent struct {
```
CarsonHoffman, line 24: Regardless of whether this type stays exported or not, "unmarshalable" represents the term more accurately than "unmarshable", even if it is a bit clunky.

message.go
FedorLap2006: hmm, thought about that, but yeah, thanks

components.go
FedorLap2006: Thanks for advise.

components.go
```diff
@@ -0,0 +1,133 @@
...
+	err = json.Unmarshal(src, &v)
+	if err != nil {
+		return
+	}
+
+	var data MessageComponent
+	switch v.Type {
+	case ActionsRowComponent:
+		v := ActionsRow{}
+		err = json.Unmarshal(src, &v)
```
CarsonHoffman, line 40: I'm getting the following panic in this example:
```go
s, _ := discordgo.New("Bot <token>")

ms := &discordgo.MessageSend{
	Content: "test",
	Components: []discordgo.MessageComponent{
		&discordgo.ActionsRow{
                    Components: []discordgo.MessageComponent{
			&discordgo.Button{
				Label:    "test",
				CustomID: "test",
			},
		    },
		},
	},
}
m, err := s.ChannelMessageSendComplex("<id>", ms)
if err != nil {
	panic(err)
}
```
```go
panic: json unmarshal: json: cannot unmarshal object into Go struct field .components of type discordgo.MessageComponent
```
Are you sure that we don't need to have a custom `UnmarshalJSON` function for `ActionsRow`, since it contains a slice of `MessageComponent`, which is an interface? Did you see this type of issue in your testing?

FedorLap2006: Nope, thanks for pointing it out. However, on my opinion would be more cleaner if we just make for the slice another type, like I was doing previously.

CarsonHoffman: I'd much rather have the tedious work done once internally than have the embedding as a confusing API surface to newcomers (in addition to, say, having to manually convert each element of the slice if they wanted to reuse the components in another message). Having the separate type just isn't something that should be exposed by the library if we can help it.

FedorLap2006: Yeah, sure.

ChrisMcDearman:
My command router which just operates off of a MessageCreate handler breaks with this PR with the following trace:

`panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x71daff]

goroutine 56 [running]:
github.com/bwmarrin/discordgo.(*Message).UnmarshalJSON(0x0, 0xc000438600, 0x2c3, 0x300, 0x0, 0x0)
        /home/chris/go/pkg/mod/github.com/!fedor!lap2006/discordgo@v0.22.1-0.20210619182440-23a197a98005/message.go:144 +0x9f
encoding/json.(*decodeState).object(0xc000284420, 0xa0d840, 0xc000010218, 0x16, 0xc000284448, 0xc0000f9d7b)
        /home/chris/go/src/encoding/json/decode.go:609 +0x20a3
encoding/json.(*decodeState).value(0xc000284420, 0xa0d840, 0xc000010218, 0x16, 0xc0000f9d68, 0x524285)
        /home/chris/go/src/encoding/json/decode.go:370 +0x6d
encoding/json.(*decodeState).unmarshal(0xc000284420, 0xa0d840, 0xc000010218, 0xc000284448, 0x0)
        /home/chris/go/src/encoding/json/decode.go:180 +0x1f4
encoding/json.Unmarshal(0xc000438600, 0x2c3, 0x300, 0xa0d840, 0xc000010218, 0x1, 0x0)
        /home/chris/go/src/encoding/json/decode.go:107 +0x11a
github.com/bwmarrin/discordgo.(*Session).onEvent(0xc00024ae00, 0x1, 0xc00001a400, 0x2eb, 0x400, 0x0, 0x0, 0x0)
        /home/chris/go/pkg/mod/github.com/!fedor!lap2006/discordgo@v0.22.1-0.20210619182440-23a197a98005/wsapi.go:584 +0xf65
github.com/bwmarrin/discordgo.(*Session).listen(0xc00024ae00, 0xc0003609a0, 0xc0004f2480)
        /home/chris/go/pkg/mod/github.com/!fedor!lap2006/discordgo@v0.22.1-0.20210619182440-23a197a98005/wsapi.go:250 +0x7e
created by github.com/bwmarrin/discordgo.(*Session).Open
        /home/chris/go/pkg/mod/github.com/!fedor!lap2006/discordgo@v0.22.1-0.20210619182440-23a197a98005/wsapi.go:202 +0xbc5
exit status 2`

FedorLap2006:
> My command router which just operates off of a MessageCreate handler breaks with this PR with the following trace:

@ChrisMcDearman  Should be fixed now.

components.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Button represents button component.
+type Button struct {
+	Label    string      `json:"label"`
+	Style    ButtonStyle `json:"style"`
+	Disabled bool        `json:"disabled"`
+	Emoji    ButtonEmoji `json:"emoji"`
+
+	// NOTE: Only button with LinkButton style can have link. Also, Link is mutually exclusive with CustomID.
+	Link     string `json:"url,omitempty"`
```
CarsonHoffman, line 125: A bit of a nit, but is there any particular reason this field name deviates from `URL`? I don't see a compelling reason to go against what will be in the Discord documentation (as I discovered in testing this!).

components.go
```diff
@@ -0,0 +1,149 @@
...
+	json.Marshaler
+	Type() ComponentType
+}
+
+type unmarshalableMessageComponent struct {
+	MessageComponent
+}
+
+// UnmarshalJSON is a helper function to unmarshal MessageComponent object.
+func (umc *unmarshalableMessageComponent) UnmarshalJSON(src []byte) (err error) {
```
CarsonHoffman, line 27: This discussion was hashed out [in the slash commands PR](https://github.com/bwmarrin/discordgo/pull/856#discussion_r577693861); especially when we have only one return value, using named returns isn't super great practice. There are quite a few functions like this, and going with explicit returns simply makes them easier to follow, as well as making it harder for subtle bugs to crop up in the future.

CarsonHoffman (changes requested):

CarsonHoffman:
Outside of a few outstanding small things, this is looking quite solid to me; the API surface is definitely cleaner than at the start. Not to jinx it, but as long as nothing else comes up in the near future, I'm excited to get this merged in! 🤠

FedorLap2006:
I agree with @CarsonHoffman - this is pretty much it. But if anyone wanna give additional review/feedback, feel free while Carson hasn't merged it yet :cowboy_hat_face:

FedorLap2006:
Oh, yeah, forgot about them!

FedorLap2006:
Thanks everyone for contribution, feedbacks, and special thanks @CarsonHoffman for his patience and reviews!

kenvogg:
Thanks for this! 

Could we get some comments in https://github.com/bwmarrin/discordgo/blob/master/examples/components/main.go about how to use this / what it does?

It's not immediately clear what this new _interactions_ functionality is, or what we are supposed to see when running the example. After attempting it, I was met with:

`Cannot create slash command: HTTP 403 Forbidden, {"message": "Missing Access", "code": 50001}`. 

Are there some prerequisite permissions that the bot needs? If so, a comment on adding them would be great.

PR #937, opened 2021-05-21
fix UserFlags constant types

zephyrtronium:
Only UserFlagDiscordEmployee had type UserFlags. All other constants
defined in the same block were untyped integer constants.

This change makes all UserFlags constants have type UserFlags. The
primary impact is that godoc and pkg.go.dev should render these
constants under the documentation for UserFlags, rather than under
package-level constants.

### DIFF
--- a/user.go
+++ b/user.go
@@ -9,19 +9,19 @@ type UserFlags int
 // Valid UserFlags values
 const (
 	UserFlagDiscordEmployee           UserFlags = 1 << 0
-	UserFlagDiscordPartner                      = 1 << 1
-	UserFlagHypeSquadEvents                     = 1 << 2
-	UserFlagBugHunterLevel1                     = 1 << 3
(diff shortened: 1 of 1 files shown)


jolheiser (approved):

zephyrtronium:
It occurs to me that, since these constants were untyped, they could be used in expressions with variables of types other than `UserFlags`. So, this might technically be a breaking change. It might be a better idea to drop this PR, then possibly create a new set of constants which will show up in the right place in documentation.

CarsonHoffman:
You're technically correct, but this was clearly a bug. I think that fixing it supersedes concerns about a breaking change here (and part of typing constants is to help prevent type misuse by intermingling with other types, so I think that forcing current users to be explicit in their usage is worth preventing the cases where users are accidentally _not_ explicit).

PR #940, opened 2021-05-21
Implement `RESTClient` interface and update `Session.Client` to be of type `RESTClient`

### LINKED ISSUE #939: Changing `Session.Client` to be interface type instead of concrete `*http.Client` can help with writing unit tests
Changing `Session.Client` to be an interface type instead of a concrete `*http.Client` can help with writing unit tests for projects that depend upon `discordgo`~~, while not breaking existing users by still allowing `*http.Client`.~~ @zacharyburkett brought up a good point that people may be using the existing `Session.Client` fields directly, so this could be a breaking change. It might be hard to find an real-world example where someone is using those fields directly where this change would break them, but the possibility exists.

An example repository I've been working on to help ease unit testing with `discordgo` can be found here: https://github.com/ewohltman/discordgo-mock

(FYI, the repo is still pretty new and no doubt missing API paths for the time being).

In particular, this function below allows us to create a ` discordgo.RESTClient`-satisfying object that will use `discordgo.State` to process API requests (found in https://github.com/ewohltman/discordgo-mock/blob/main/pkg/mockrest/mockrest.go):

```go
type roundTripperFunc func(req *http.Request) (*http.Response, error)

func (rt roundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
	return rt(req)
}

type Client struct {
	*http.Client
	state  *discordgo.State
	router *mux.Router
}

func NewClient(state *discordgo.State) *Client {
	router := mux.NewRouter()

	client := &Client{
		Client: &http.Client{
			Transport: roundTripperFunc(func(r *http.Request) (*http.Response, error) {
				responseRecorder := httptest.NewRecorder()
				router.ServeHTTP(responseRecorder, r)
				return responseRecorder.Result(), nil
			}),
		},
		state:  state,
		router: router,
	}

	apiVersion := "/api/v" + discordgo.APIVersion

	client.addHandlersGuilds(apiVersion)
	client.addHandlersChannels(apiVersion)
	client.addHandlersUsers(apiVersion)

	return client
}
```

Example usage of the `RESTClient` with `Session.Client` change (found in https://github.com/ewohltman/discordgo-mock/blob/main/pkg/mocksession/mocksession_test.go):

```go
func newSession(expectedResources *resources) *discordgo.Session {
	state, err := mockstate.New(
		mockstate.WithUser(mockuser.New(
			mockuser.WithID("Bot"),
			mockuser.WithUsername("Bot"),
			mockuser.IsBot(true),
		)),
		mockstate.WithGuilds(expectedResources.guild),
	)
	if err != nil {
		panic(err)
	}

	session, err := mocksession.New(
		mocksession.WithState(state),
		mocksession.WithRESTClient(mockrest.NewClient(state)),
	)
	if err != nil {
		panic(err)
	}

	return session
}
```

ewohltman:
This PR is a small change to `Session.Client` to be of type `RESTClient` interface instead of a concrete `*http.Client`.

Resolves #939

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -24,6 +24,14 @@ import (
 	"github.com/gorilla/websocket"
 )
 
+// RESTClient is an interface definition for performing REST API calls to
+// Discord. It allows for the usage of a standard library *http.Client as well
+// as other HTTP clients with custom functionality and enables the ability to
+// write unit tests.
+type RESTClient interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
 // A Session represents a connection to the Discord API.
 type Session struct {
 	sync.RWMutex
@@ -90,7 +98,7 @@ type Session struct {
 	State *State
 
 	// The http client used for REST requests
-	Client *http.Client
+	Client RESTClient
 
 	// The user agent used for REST APIs
 	UserAgent string


FedorLap2006:
This is a very good idea actually. But please, note the checks.

FedorLap2006 (approved):
That's a great idea! But note the following checks please.

ewohltman:
@FedorLap2006 Glad you think so! Guess there was a weird hiccup during that last CI build. I just triggered it again and it passed this time :+1:

zacharyburkett:
As noted in #939, this is a breaking API change. PR should be closed.

AverageComet250:
Or using function overrides to allow both can be done, and deprecation warnings can be added if the function is called with the `*http.client`

ewohltman:
> Or using function overrides to allow both can be done, and deprecation warnings can be added if the function is called with the `*http.client`

@AverageComet250: I started looking into an alternative approach by replacing the `*http.Client.Transport` with a custom `http.RoundTripper` that seems to do the job:

https://github.com/ewohltman/discordgo-mock

Example usage of this new "helper" library can be found here until I can more fully flesh out the `discordgo-mock` repo with better examples and unit tests:

https://github.com/ewohltman/ephemeral-roles/blob/develop/internal/pkg/mock/session.go

Comments, questions, and improvements are welcome to `discordgo-mock`!

Issue #942, opened 2021-05-23
User/role permissions for using slash commands


NotUnlikeTheWaves:
Hello, I've implemented slash commands successfully in a bot but haven't found permissions implemented in this library yet, as described in https://discord.com/developers/docs/interactions/slash-commands#permissions and the four subsections describing the API calls for those permissions (get guild permissions for all commands in a guild, get permissions of a single command in a guild, edit a single command in a guild, bulk edit permissions in a guild)

I've already taken steps to add this functionality myself in the following pull request #943. Feedback is welcomed in either location

NotUnlikeTheWaves:
One thing I've noticed in the docs under the permissions part is the default_permissions, which can optionally be sent on creating and editing a command. If set to false, it means the command is disabled until a permissions overwrite exists on that command.

It's a little vague on how best to implement this as the default boolean value is 'false' which could be fatal for many commands made that assume that creating a command without special intervention enables it for everyone. This is also what's implied via the discord API docs since the field is optional.

FedorLap2006:
Merged in d5bacb5401a6652f48edfbf133fd55a159b95b18

PR #943, opened 2021-05-23
Add permissions API endpoints for slash commands

NotUnlikeTheWaves:
See main issue #942 

This adds the permissions API endpoints for application commands. I have tested these locally with success, and pushed my test version here: https://github.com/Southclaws/cj/commit/a07d1fa39ecca36f4de508800f1bb24c37a6575d
This indeed lacks the single command edit call, which is the first I implemented and later replaced with the bulk edit due to the amount of edits I'm making (and to test that one too). The commit should serve as an idea on how to work with it in the current form, to which feedback is welcome.

I have tried to keep naming as close to Discord's API naming for objects, but I'm happy to receive any feedback regarding them as well as for anything else in this PR.

Kind regards

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -67,15 +67,15 @@ type ApplicationCommandPermissions struct {
 
 // ApplicationCommandPermissionsList represents a list of ApplicationCommandPermissions, needed for serializing to JSON.
 type ApplicationCommandPermissionsList struct {
-	Permissions []ApplicationCommandPermissions `json:"permissions"`
+	Permissions []*ApplicationCommandPermissions `json:"permissions"`
 }
 
 // GuildApplicationCommandPermissions represents all permissions for a single guild command.
 type GuildApplicationCommandPermissions struct {
-	ID            string                          `json:"id"`
-	ApplicationID string                          `json:"application_id"`
-	GuildID       string                          `json:"guild_id"`
-	Permissions   []ApplicationCommandPermissions `json:"permissions"`
+	ID            string                           `json:"id"`
+	ApplicationID string                           `json:"application_id"`
+	GuildID       string                           `json:"guild_id"`
+	Permissions   []*ApplicationCommandPermissions `json:"permissions"`
 }
 
 // ApplicationCommandPermissionType indicates whether a permission is user or role based.
@@ -83,8 +83,8 @@ type ApplicationCommandPermissionType uint8
 
 // Application command permission types.
 const (
-	ApplicationCommandPermissionTypeRole = ApplicationCommandPermissionType(iota + 1)
-	ApplicationCommandPermissionTypeUser
+	ApplicationCommandPermissionTypeRole ApplicationCommandPermissionType = 1
+	ApplicationCommandPermissionTypeUser ApplicationCommandPermissionType = 2
 )
 
 // InteractionType indicates the type of an interaction event.
--- a/interactions.go
+++ b/interactions.go
@@ -58,6 +58,30 @@ type ApplicationCommandOptionChoice struct {
 	Value interface{} `json:"value"`
 }
 
+type ApplicationCommandPermissions struct {
+	ID         string                           `json:"id"`
+	Type       ApplicationCommandPermissionType `json:"type"`
+	Permission bool                             `json:"permission"`
+}
+
+type ApplicationCommandPermissionsList struct {
+	Permissions []ApplicationCommandPermissions `json:"permissions"`
+}
+
+type GuildApplicationCommandPermissions struct {
+	ID            string                          `json:"id"`
+	ApplicationId string                          `json:"application_id"`
+	GuildID       string                          `json:"guild_id"`
+	Permissions   []ApplicationCommandPermissions `json:"permissions"`
+}
+
+type ApplicationCommandPermissionType uint8
+
+const (
+	ApplicationCommandPermissionTypeRole = ApplicationCommandPermissionType(iota + 1)
+	ApplicationCommandPermissionTypeUser
+)
+
 // InteractionType indicates the type of an interaction event.
 type InteractionType uint8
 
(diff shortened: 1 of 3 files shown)


restapi.go
```diff
@@ -2545,6 +2545,62 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
...
+	_, err = s.RequestWithBucketID("PUT", endpoint, permissions, endpoint)
+
+	return
+}
+
+// ApplicationCommandBatchEditPermissions edits the permissions of a batch of commands
+// appID       : The Application ID
+// guildID     : The guild ID to batch edit commands of
+// permissions : A list of permissions paired with a command ID, guild ID, and application ID per application command
+func (s *Session) ApplicationCommandBatchEditPermissions(appID, guildID string, permissions []GuildApplicationCommandPermissions) (err error) {
```
bigblind, line 2596: It seems like this library mostly puts the verb at the end of the method name, as in `ApplicationCommandBulkOverwrite`. This same comment applies to `ApplicationCommandEditPermissions` above.

bigblind (commented):
I'm not a maintainer of this repo, but after a reply from Carson on Discord, I took a look at the code for a review. It looks good to me, but in one place, the naming seems inconsistent with the rest of the repo in one place.

restapi.go
NotUnlikeTheWaves: Thanks for your feedback, I've reflected it in the new change. I've kept the naming as BatchEdit as the API docs specifically mention it as a batch, not a bulk.

interactions.go
```diff
@@ -58,6 +58,35 @@ type ApplicationCommandOptionChoice struct {
...
+	Permissions   []ApplicationCommandPermissions `json:"permissions"`
+}
+
+// ApplicationCommandPermissionType indicates whether a permission is user or role based.
+type ApplicationCommandPermissionType uint8
+
+// Application command permission types.
+const (
+	ApplicationCommandPermissionTypeRole = ApplicationCommandPermissionType(iota + 1)
+	ApplicationCommandPermissionTypeUser
```
CarsonHoffman, line 87: If you could use absolute values rather than `iota` here, that'd be great. In a simple case like this it seems excessive, but `iota` is constantly a source of issues in the library; if new constants are added in the future, and especially if gaps between constants are added (Discord likes to do this), things go wrong quickly. If we have a specific external value that we're depending on, we should use it.

restapi.go
```diff
@@ -2545,6 +2545,62 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 	return
 }
 
+// GuildApplicationCommandsPermissions receives all permissions for application commands that have any permissions
+// appID       : The application ID
+// guildID     : Guild ID to retrieve all guild-specific permissions for application commands in the guild, both global and guild specific
+func (s *Session) GuildApplicationCommandsPermissions(appID, guildID string) (permissions []*GuildApplicationCommandPermissions) {
```
CarsonHoffman, line 2551: Is there a specific reason why this function returns `[]*GuildApplicationCommandPermissions`, while `ApplicationCommandPermissions` returns a non-pointer `GuildApplicationCommandPermissions`, and `ApplicationCommandPermissionsBatchEdit` takes a slice of non-pointer `[]GuildApplicationCommandPermissions`? It seems like we'd most likely want to keep some consistency here.

CarsonHoffman (changes requested):

restapi.go
NotUnlikeTheWaves: Good point, wasn't very focused on it. I've changed it to use pointer types where possible for structs

NotUnlikeTheWaves:
One thing I've noticed in the docs under the [permissions](https://discord.com/developers/docs/interactions/slash-commands#permissions) part is the `default_permissions`, which can optionally be sent on creating and editing a command. If set to false, it means the command is disabled until a permissions overwrite exists on that command.

I am currently not sure how to best handle this. Perhaps it's good to move it to the issue and tackle it in a new pull request

kislball:
any updates?

FedorLap2006:
> One thing I've noticed in the docs under the [permissions](https://discord.com/developers/docs/interactions/slash-commands#permissions) part is the `default_permissions`, which can optionally be sent on creating and editing a command. If set to false, it means the command is disabled until a permissions overwrite exists on that command.
> 
> I am currently not sure how to best handle this. Perhaps it's good to move it to the issue and tackle it in a new pull request

True, but I have some solution to this.

```go
type DefaultApplicationCommandPermission uint8

func (dp DefaultApplicationCommandPermission) MarshalJSON() ([]byte, error) {
  switch dp {
  case DefaultApplicationCommandPermissionDeny:
    return []byte(`true`), nil
  case DefaultApplicationCommandPermissionDeny:
    return []byte(`false`), nil
  }
  // The field is set as omitempty, so if we will output null it will omit this field: https://play.golang.org/p/HIIhLa0mGVN
  return []byte(`null`), nil
}

const (
  DefaultApplicationCommandPermissionNone DefaultApplicationCommandPermission = iota // 0, so will be the default value of the field
  DefaultApplicationCommandPermissionDeny
  DefaultApplicationCommandPermissionAllow
)

type ApplicationCommandPermissions struct {
  DefaultPermission DefaultApplicationCommandPermission `json:"default_permission,omitempty"`
}
```

@CarsonHoffman what do you think?

Omegaice:
@FedorLap2006 I borrowed this code as I needed to add some permission stuff quickly to a bot I was working on. It works fine but does also need an UnmarshalJSON if you end up having it as part of the ApplicationCommand structure such as [this](https://github.com/Omegaice/discordgo/blob/master/interactions.go#L37).

Luberry:
I'm using a [fork](https://github.com/Luberry/discordgo) of this pr merged with the upstream branch in my bot, however using a bool pointer for default permissions instead. which just works without unmarshaljson

Helfull:
any updates here?

FedorLap2006:
Merged by d5bacb5401a6652f48edfbf133fd55a159b95b18

PR #945, opened 2021-05-28
added basic embed message example

XiovV:


AverageComet250 (approved):
I'm surprised this doesn't exist already!

Someone else will need to review the code though cause I don't know go

jolheiser:
This exists, just in the wiki rather than an example.

https://github.com/bwmarrin/discordgo/wiki/FAQ#sending-embeds

XiovV:
> This exists, just in the wiki rather than an example.
> 
> https://github.com/bwmarrin/discordgo/wiki/FAQ#sending-embeds

I didn't know that, thank you for pointing it out. However, I think it would make more sense to put the example in the examples directory rather than the FAQ.

FedorLap2006:
I don't think so. Having it in the examples makes sense, but... message embeds are the core functionality of the Discord API. And all core functionality may have a separated example, but in this case - creating a new example just for embeds - doesn't make any real sense. Since only major sets of API functionality had a separate example: slash commands, components (coming soon btw), audio, etc.

XiovV:
The reason why I made this PR is because I was struggling to figure out how to use embeds. I tried looking through the examples directory but there was no example. It never occured to me to look through the FAQ because, quite frankly, putting examples in the FAQ makes no sense because you've already got the examples directory. I think the future users of this library will find this example useful. If I struggled with this I'm willing to bet that more people did too, and merging my example could save people a ton of time and confusion.

AverageComet250:
I think having this in the examples directory is a good idea, and creating examples for each major feature rather than each core feature would be better in my opinion

omznc:
I believe adding this to the examples directory would be a good course of action, for reasons stated by @AverageComet250

tobigiwa:
Thanks God for this... believe me, I was this giving up on embed with discordgo when I saw this... it needs to be in the example dir

tobigiwa:
> This exists, just in the wiki rather than an example.
> 
> https://github.com/bwmarrin/discordgo/wiki/FAQ#sending-embeds
 
@jolheiser the helper func with newEmbed(). please where is this comming from? it not on discordgo struct itself

FedorLap2006:
This example provides a link to the https://gist.github.com/Necroforger/8b0b70b1a69fa7828b8ad6387ebb3835. Which wraps `discordgo.MessageEmbed` and provides this functionality

PR #948, opened 2021-06-02
Get webhook message endpoint

Xpl0itR:
This is useful for getting the ID of the interaction response message which can be used for adding reactions among other things

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2236,6 +2236,23 @@ func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *Webho
 	return
 }
 
+// WebhookMessageGet gets a webhook message.
+// webhookID : The ID of a webhook
+// token     : The auth token for the webhook
+// messageID : The ID of message to get
+func (s *Session) WebhookMessageGet(webhookID, token, messageID string) (message *Message, err error) {
+	uri := EndpointWebhookMessage(webhookID, token, messageID)
+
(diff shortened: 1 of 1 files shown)


restapi.go
```diff
@@ -2236,6 +2236,23 @@ func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *Webho
 	return
 }
 
+// WebhookMessageGet gets a webhook message.
+// webhookID : The ID of a webhook
+// token     : The auth token for the webhook
+// messageID : The ID of message to get
+func (s *Session) WebhookMessageGet(webhookID, token, messageID string) (message *Message, err error) {
```
FedorLap2006, line 2243: Probably would be better to use here `WebhookMessage` instead.

restapi.go
```diff
@@ -2557,6 +2574,13 @@ func (s *Session) InteractionRespond(interaction *Interaction, resp *Interaction
 	return err
 }
 
+// InteractionResponseGet gets the response to an interaction.
+// appID       : The application ID.
+// interaction : Interaction instance.
+func (s *Session) InteractionResponseGet(appID string, interaction *Interaction) (*Message, error) {
```
FedorLap2006, line 2580: Look at another comment

FedorLap2006 (commented):
Probably you should rename this PR to "Get webhook message endpoint". But anyways, I think that's a part of the #951. Also, mind the comments.

CarsonHoffman:
Thank you for the additions!

PR #952, opened 2021-06-17
add ReferencedMessage field to Message struct

sauerbraten:
The field is called 'referenced_message' on https://discord.com/developers/docs/resources/channel#message-object, and holds information about the message being replied to.

### DIFF
--- a/message.go
+++ b/message.go
@@ -123,6 +123,11 @@ type Message struct {
 	// This is a combination of bit masks; the presence of a certain permission can
 	// be checked by performing a bitwise AND between this int and the flag.
 	Flags MessageFlags `json:"flags"`
+
+	// The message being replied to.
+	// Only messages of type 19 (MessageTypeReply) or 21 can have this field (but it might still be missing).
+	// If it exists but is null, the referenced message was deleted.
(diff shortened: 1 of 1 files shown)


CarsonHoffman:
I'm not sure that the "If it exists but is null, the referenced message was deleted." portion is really relevant here, as we don't have a way of differentiating between an absent field and a `null` one in this current implementation—they both end up in `nil`. Given that Discord makes this distinction, I think we should try to allow users to effectively differentiate between the two, as well.

macmv:
> I'm not sure that the "If it exists but is null, the referenced message was deleted." portion is really relevant here, as we don't have a way of differentiating between an absent field and a `null` one in this current implementation—they both end up in `nil`. Given that Discord makes this distinction, I think we should try to allow users to effectively differentiate between the two, as well.

You can check the message type. It will be `MessageTypeReply` if this was a reply, and another type if it was not a reply. Isn't this enough to differentiate between these situations?

ayntee:
Bump

FedorLap2006:
Merged in by 992358e106f1fd755830290ddb6a7d36af5ec558

PR #956, opened 2021-06-22
Components Support

fletcherist:
Spec: https://discord.com/developers/docs/interactions/message-components

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -86,6 +86,8 @@ type Interaction struct {
 	// if it was invoked in a guild, the `Member` field will be filled instead.
 	// Make sure to check for `nil` before using this field.
 	User *User `json:"user"`
+	// For components, the message they were attached to
+	Message *Message `json:"message"`
 
 	Token   string `json:"token"`
 	Version int    `json:"version"`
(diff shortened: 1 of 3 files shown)


Pedro-Pessoa:
Isn't this already being worked on #933?

FedorLap2006:
Same question.

fletcherist:
@Pedro-Pessoa not exactly. The implementation by @FedorLap2006 breaks backwards-compatibility and contains breaking changes. This, on the other hand, can be released as minor patch.
Screen Shot 2021-06-23 at 1 19 26 PM

So i think the best way is to treat this as yet another approach

CarsonHoffman:
The changes introduced in the existing PR serve to more closely align the library with the state of the world on Discord's end after _they_ made some breaking changes; it's not worth having incorrectly-named things years down the line in the name of backwards compatibility on a feature that **has yet to appear in any tagged release** (i.e. it's not really a breaking change at all).

I don't mean to shoot you down too hard, but at this point, I see no reason that the other PR cannot be moved ahead with, so if you'd like to see buttons in the library as soon as possible, I'd suggest contributing a bit of your time as a review on that PR rather than duplicating effort.

fletcherist:
@CarsonHoffman okay I clearly got your idea. Closing this tho, thanks for explanation

PR #959, opened 2021-06-24
Add bindings for custom un/marshal functions

zekroTJA:
Because large bots need to handle a lot of API requests and responses as well as web socket event messages which all need to go through JSON serialization and deserialization, I thought it might be advantageous if you would be able to swap out the internal JSON implementation with custom ones like [`bytedance/sonic`](https://github.com/bytedance/sonic) as an example.

Therefore, I've added two new types in `types.go` to define the Marshal and Unmarshal functions.
```go
// MarshalFunc encodes an object instance into JSON byte data.
type MarshalFunc func(v interface{}) (data []byte, err error)

// UnmarshalFunc decodes JSON byte data into an object instance.
type UnmarshalFunc func(data []byte, v interface{}) (err error)
```

As with other preferences, I've added them to the `Session` struct.
```go
// The function used to encode object instances
// into JSON data
//
// Defaults to json.Marshal
MarshalFunc MarshalFunc

// The function used to decode JSON data into
// object instances
//
// Defaults to json.Unmarshal
UnmarshalFunc UnmarshalFunc
```

Both values default to the internal `encoding/json` implementation on initialization by default.
```go
func New(args ...interface{}) (s *Session, err error) {
	// Create an empty Session interface.
	s = &Session{
		// ...
		MarshalFunc:            json.Marshal,
		UnmarshalFunc:          json.Unmarshal,
	}
	// ...
```

To utilize the defined functions, It was necessary to pass the `Session` instance to some functions which utilize `Marshal` and `Unmarshal`. Some types which are utilizing these functions as well (like `TimeStamp`), I left untouched to avoid changing the API.

Because I could not get the unit tests to run properly without authentication failures, I've created a quick and dirty example implementation with my fork and sonic as JSON parser. Here you can find the code of it: https://github.com/zekroTJA/discordgo-sonic

---

I've set this repo as `Draft` for the moment just to see if the pull request tests succeed with my changes and make further fixes based on that. Also, I want to implement my fork in one of my own Discord Bots which heavily bases on DiscordGo just to see if there are problems occurring during "field test".

Also, it would be really interesting to get some feedback what you think about this implementation and if you think it fits into the project and makes any sense to implement it this way. :)

Edit: I've now used this implementation in combination with pquerna/ffjson without any issues the last couple of days.

### DIFF
--- a/discord_test.go
+++ b/discord_test.go
@@ -7,6 +7,8 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
+
+	"github.com/joho/godotenv"
 )
 
 //////////////////////////////////////////////////////////////////////////////
@@ -23,6 +25,8 @@ var (
 )
 
 func init() {
+	godotenv.Load()
+
 	fmt.Println("Init is being called.")
 	if envBotToken != "" {
 		if d, err := New(envBotToken); err == nil {
--- a/discord_test.go
+++ b/discord_test.go
@@ -7,8 +7,6 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
-
-	"github.com/joho/godotenv"
 )
 
 //////////////////////////////////////////////////////////////////////////////
@@ -25,8 +23,6 @@ var (
 )
 
 func init() {
-	godotenv.Load()
-
 	fmt.Println("Init is being called.")
 	if envBotToken != "" {
 		if d, err := New(envBotToken); err == nil {
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,5 @@ go 1.13
 
 require (
 	github.com/gorilla/websocket v1.4.2
-	github.com/joho/godotenv v1.3.0
 	golang.org/x/crypto v0.0.0-20210421170649-83a5a9bb288b
 )
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,6 @@ go 1.13
 
 require (
 	github.com/gorilla/websocket v1.4.2
+	github.com/joho/godotenv v1.3.0
 	golang.org/x/crypto v0.0.0-20210421170649-83a5a9bb288b
 )
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,7 @@
 github.com/gorilla/websocket v1.4.2 h1:+/TMaTYc4QFitKJxsQ7Yye35DkWvkdLcvGKqM+x0Ufc=
 github.com/gorilla/websocket v1.4.2/go.mod h1:YR8l580nyteQvAITg2hZ9XVh4b55+EU/adAjf1fMHhE=
+github.com/joho/godotenv v1.3.0 h1:Zjp+RcGpHhGlrMbJzXTrZZPrWj+1vfm90La1wgB6Bhc=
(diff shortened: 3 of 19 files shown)


restapi.go
```diff
@@ -32,7 +31,7 @@ import (
 
 // All error constants
 var (
-	ErrJSONUnmarshal           = errors.New("json unmarshal")
+	ErrJSONUnmarshal           = errors.New("json s.unmarshal")
```
FedorLap2006, line 34: I think you should leave this error as it was, since the meaning didn't change.

discord_test.go
```diff
@@ -23,6 +25,8 @@ var (
 )
 
 func init() {
+	godotenv.Load()
```
FedorLap2006, line 28: I think would be better to keep out such libraries since (see another comment)

go.mod
```diff
@@ -4,5 +4,6 @@ go 1.13
 
 require (
 	github.com/gorilla/websocket v1.4.2
+	github.com/joho/godotenv v1.3.0
```
FedorLap2006, line 7: You add them to go.mod and they become necessary dependency for everyone who uses the library

restapi.go
```diff
@@ -149,9 +148,9 @@ func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b
 		}
 	case 429: // TOO MANY REQUESTS - Rate limiting
 		rl := TooManyRequests{}
-		err = json.Unmarshal(response, &rl)
+		err = s.UnmarshalFunc(response, &rl)
 		if err != nil {
-			s.log(LogError, "rate limit unmarshal error, %s", err)
+			s.log(LogError, "rate limit s.unmarshal error, %s", err)
```
FedorLap2006, line 153: Same here.

FedorLap2006 (changes requested):
I think that's a great start, but your code has some problems right now.

go.mod
zekroTJA: Oh yeah, of course. Added it actually just for local testing but yeah, I'm still too accustomed to just do `git add .` sadly. :D

zekroTJA:
> I think that's a great start, but your code has some problems right now.

Thank you very much for your review and feedback. :)

go.sum
```diff
@@ -1,5 +1,7 @@
 github.com/gorilla/websocket v1.4.2 h1:+/TMaTYc4QFitKJxsQ7Yye35DkWvkdLcvGKqM+x0Ufc=
 github.com/gorilla/websocket v1.4.2/go.mod h1:YR8l580nyteQvAITg2hZ9XVh4b55+EU/adAjf1fMHhE=
+github.com/joho/godotenv v1.3.0 h1:Zjp+RcGpHhGlrMbJzXTrZZPrWj+1vfm90La1wgB6Bhc=
+github.com/joho/godotenv v1.3.0/go.mod h1:7hK45KPybAkOC6peb+G5yklZfMxEjkZhHbwpqxOKXbg=
```
FedorLap2006, line 4: Not sure why this was added here.

FedorLap2006 (changes requested):

FedorLap2006:
Also note the conflicts.

FedorLap2006:
After a bit of consideration, we've decided that we should do global scoped marshal/unmarshal instead.
The development will be done in the separate PR, so I'm going to close this one.

PR #972, opened 2021-07-20
Fix slash_commands example  unregister on shutdown

lammel:
The current slash_commands example has on option to unregister commands on shutdown that was not implemented.

This PR adds the unregister on shutdown by remembering all registered commands and unregisters them later.


lammel:
This is a duplicate of PR #911.

PR #975, opened 2021-07-25
add some data to applications

Kesigomon:
Add some data to applications.
https://discord.com/developers/docs/resources/application#application-object-application-structure

### DIFF
--- a/oauth2.go
+++ b/oauth2.go
@@ -56,12 +56,12 @@ type Application struct {
 	Bot                 *User     `json:"bot"`
 	Team                *Team     `json:"team"`
 	VerifyKey           string    `json:"verify_key"`
-	RpcOrigins          *[]string `json:"rpc_origins,omitempty"`
-	TeamsOfServiceUrl   string    `json:"teams_of_service_url,omitempty"`
-	PrivacyPolicyUrl    string    `json:"privacy_policy_url,omitempty"`
+	RPCOrigins          *[]string `json:"rpc_origins,omitempty"`
+	TeamsOfServiceURL   string    `json:"teams_of_service_url,omitempty"`
+	PrivacyPolicyURL    string    `json:"privacy_policy_url,omitempty"`
 	Summary             string    `json:"summary"`
-	GuildId             string    `json:"guild_id,omitempty"`
-	PrimarySkuId        string    `json:"primary_sku_id,omitempty"`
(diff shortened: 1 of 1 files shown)


oauth2.go
```diff
@@ -55,6 +55,15 @@ type Application struct {
 	Owner               *User     `json:"owner"`
 	Bot                 *User     `json:"bot"`
 	Team                *Team     `json:"team"`
+	VerifyKey           string    `json:"verify_key"`
+	RPCOrigins          *[]string `json:"rpc_origins,omitempty"`
```
CarsonHoffman, line 59: Is having a pointer to the slice really necessary here? The field is not nullable, and I don't think that we really need to differentiate between present and not present—I don't think that the field would be sent if the array were empty.

Kesigomon: Fixed!

oauth2.go
```diff
@@ -55,6 +55,15 @@ type Application struct {
 	Owner               *User     `json:"owner"`
 	Bot                 *User     `json:"bot"`
 	Team                *Team     `json:"team"`
+	VerifyKey           string    `json:"verify_key"`
+	RPCOrigins          []string  `json:"rpc_origins,omitempty"`
+	TeamsOfServiceURL   string    `json:"teams_of_service_url,omitempty"`
+	PrivacyPolicyURL    string    `json:"privacy_policy_url,omitempty"`
+	Summary             string    `json:"summary"`
+	GuildID             string    `json:"guild_id,omitempty"`
+	PrimarySkuID        string    `json:"primary_sku_id,omitempty"`
```
CarsonHoffman, line 64: A bit of a nit, but see: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
> Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". As an example: ServeHTTP not ServeHttp. For identifiers with multiple initialized "words", use for example "xmlHTTPRequest" or "XMLHTTPRequest".

Although it looks a bit weird, `PrimarySKUID` is the way to go here. I know it's small, but given that all this change does is add fields, we might as well name them right. 🙂

FedorLap2006:
Merged in by 70e829694d9938f5c0fe87e05fad138ca6c408b0

PR #987, opened 2021-08-03
Add UserMessages* functions

damaredayo:
While writing my bot I noticed a lack of functions to make it easy to send Users messages in the same way you can normal channels without extra boilerplate in the actual function, so to remove that, I added these to the fork for the bot as well as a fork on my personal github so it can be shared.

If you feel some optimizations and improvements could be made, feel free.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1833,6 +1833,129 @@ func (s *Session) ChannelNewsFollow(channelID, targetID string) (st *ChannelFoll
 }
 
 // ------------------------------------------------------------------------------------------------
+// Functions specific to Direct Messages
+// ------------------------------------------------------------------------------------------------
+
+// UserMessageSend sends a message to a user.
+// userID : The ID of a User.
+// content   : The message to send.
+func (s *Session) UserMessageSend(userID, content string) (*Message, error) {
+	userChannel, err := s.UserChannelCreate(userID)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.ChannelMessageSendComplex(userChannel.ID, &MessageSend{
+		Content: content,
+	})
+}
+
+// UserMessageSendEmbed sends a message to the given private channel with embedded data.
+// userID : The ID of a User.
+// embed : The embed data to send.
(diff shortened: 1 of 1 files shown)


restapi.go
```diff
@@ -1832,6 +1832,129 @@ func (s *Session) ChannelNewsFollow(channelID, targetID string) (st *ChannelFoll
...
+// content   : The message to send.
+func (s *Session) UserMessageSend(userID, content string) (*Message, error) {
+	userChannel, err := s.UserChannelCreate(userID)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.ChannelMessageSendComplex(userChannel.ID, &MessageSend{
+		Content: content,
+	})
```
Romitou, line 1850: tell me if i'm wrong, but it might be better to use this instead
```suggestion
	return s.ChannelMessageSend(userChannel.ID, content)
```

restapi.go
```diff
@@ -1832,6 +1832,129 @@ func (s *Session) ChannelNewsFollow(channelID, targetID string) (st *ChannelFoll
...
+// embed : The embed data to send.
+func (s *Session) UserMessageSendEmbed(userID string, embed *MessageEmbed) (*Message, error) {
+	userChannel, err := s.UserChannelCreate(userID)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.ChannelMessageSendComplex(userChannel.ID, &MessageSend{
+		Embed: embed,
+	})
```
Romitou, line 1864:
```suggestion
	return s.ChannelMessageSendEmbed(userChannel.ID, embed)
```

restapi.go
damaredayo: The ChannelMessageSend calls ChannelMessageSendComplex the exact same way I do, internally, there is no need to use it because it's there for convenience. It's calling another function for pretty much no reason.

vertex-admin:
Is there a difference between the errors received with UserChannelCreate and ChannelMessageSend? If so, it would make sense to keep them separated.

PR #991, opened 2021-08-20
feat: add Banner & Accent to User

rigormorrtiss:
Reference: https://github.com/discord/discord-api-docs/pull/3448﻿

### DIFF
--- a/user.go
+++ b/user.go
@@ -74,6 +74,12 @@ type User struct {
 	// The flags on a user's account.
 	// Only available when the request is authorized via a Bearer token.
 	Flags int `json:"flags"`
+
+	// Banner is the user's banner, or an empty string if unset.
+	Banner string `json:"banner,omitempty"`
+
+	// Accent is the user's banner color encoded as an integer representation of hexadecimal color code
+	Accent int `json:"accent_color,omitempty"`
 }
 
 // String returns a unique identifier of the form username#discriminator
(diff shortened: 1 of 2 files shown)


CarsonHoffman:
Both `Banner` and `Accent` are marked as nullable in the PR you linked; if these are sent as `null`, they will fail to unmarshal. We should probably have these fields as pointers to the types to allow that nullability.

user.go
```diff
@@ -105,3 +111,18 @@ func (u *User) AvatarURL(size string) string {
...
+	if u.Banner == "" {
+		URL = ""
+	} else if strings.HasPrefix(u.Banner, "a_") {
+		URL = EndpointUserBannerAnimated(u.ID, u.Avatar)
+	} else {
+		URL = EndpointUserBanner(u.ID, u.Banner)
+	}
+
+	return URL
+}
```
CarsonHoffman, line 128: You're now trying to use a `*string` as a `string` in this function. I would also recommend adding a check for `u.Banner == nil` here before dereferencing it such that you stay consistent with the comment of "An empty string is returned if the user banner is unset."; unset might correspond to the case of `null`/`nil`.

CarsonHoffman:
There are changes that need to be made here, however the original account now seems to be deleted.

codemicro:
What changes need to be made?

PR #1002, opened 2021-09-05
Add omitempty to channel_id in MessageReference struct

corbmr:
[Per documentation](https://discord.com/developers/docs/resources/channel#message-types-replies) channel_id and guild_id are not required. GuildID is already marked as omitempty but ChannelId can be as well.

### DIFF
--- a/message.go
+++ b/message.go
@@ -403,7 +403,7 @@ type MessageApplication struct {
 // MessageReference contains reference data sent with crossposted messages
 type MessageReference struct {
 	MessageID string `json:"message_id"`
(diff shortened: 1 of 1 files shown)


ayntee:
Screenshot_20210906_154602

corbmr:
`omitempty` only applies with encoding so I would think it should be considered optional for the sake of sending, not related to whichever fields are returned on receive.

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

P.S. Sorry it took that long.

PR #1004, opened 2021-09-10
Add default permission to application commands

jolheiser:
This adds default permissions to application commands, as I noticed it was discussed in #943 

Since Discord defaults to `true`, I decided it would be easiest (and non-breaking) to invert the field when marshaling.  
This also means we don't need to deal with a null-bool implementation or deal with `omitempty`.

### Reference
[Discord Docs](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-structure)

default_permission? | boolean (default `true`) | whether the command is enabled by default when the app is added to a guild
-- | -- | --

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -39,6 +39,20 @@ type ApplicationCommand struct {
 	Version     string `json:"version,omitempty"`
 	// NOTE: Chat commands only. Otherwise it mustn't be set.
 	Options []*ApplicationCommandOption `json:"options"`
+	// Discord defaults permission to true, so we invert it in MarshalJSON
+	DefaultPermissionDisabled bool `json:"-"`
+}
+
+// MarshalJSON is a method for marshaling ApplicationCommand to a JSON object.
+func (a ApplicationCommand) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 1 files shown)


jolheiser:
For the record, I don't mind if #943 wants to adopt this change, I just figured it would be a small PR to add and _technically_ isn't dependent on it.

FedorLap2006:
@jolheiser As one of the implementation consider one from [my comment on the PR](https://github.com/bwmarrin/discordgo/pull/943#issuecomment-895078585). Your approach is nice, but not sure about the look.

jolheiser:
> @jolheiser As one of the implementation consider one from [my comment on the PR](https://github.com/bwmarrin/discordgo/pull/943#issuecomment-895078585). Your approach is nice, but not sure about the look and usage.

I did see yours, I just didn't particularly like it. 
I'm not a huge fan of double negatives like my implementation, however I cannot change the discord API.  
That bring said, I (personally) wasn't a fan of a null-bool implementation either.

Whichever gets implemented is fine with me, just my thoughts on the matter. 🙂

PR #1006, opened 2021-09-16
Add BeforeDelete to GuildMemberRemove struct

sizumita:
Added a Member copy in the GuildMemberRemove event, just like MessageDelete and GuildDelete.


sizumita:
Is there any reasons not to merge this PR?

PR #1009, opened 2021-09-18
Add RepliedUser field to MessageAllowedMentions

ayntgl:


FedorLap2006:
I think you could also add `mention` argument to `Session.ChannelMessageSendReply` :thinking:

restapi.go
```diff
@@ -1611,13 +1611,17 @@ func (s *Session) ChannelMessageSendEmbed(channelID string, embed *MessageEmbed)
 // channelID : The ID of a Channel.
 // content   : The message to send.
 // reference : The message reference to send.
-func (s *Session) ChannelMessageSendReply(channelID string, content string, reference *MessageReference) (*Message, error) {
+// mention   : Whether to mention the author of the reference message in the message or not.
+func (s *Session) ChannelMessageSendReply(channelID string, content string, reference *MessageReference, mention bool) (*Message, error) {
```
FedorLap2006, line 1615: Probably would be better to have `mention` parameter before the reference

FedorLap2006 (approved):
I'd say everything looks good to me, except one thing. But I think it's more controversial, so, yeah.

PR #1022, opened 2021-10-24
Modify GuildEmojiEdit function of restapi.go

awildboop:
Changes function to not omit empty role information. This allows for properly resetting the roles of an emoji, otherwise it appears to be presently impossible to reset the roles (without including a role ID in the edit request).

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1385,7 +1385,7 @@ func (s *Session) GuildEmojiCreate(guildID, name, image string, roles []string) 
 // guildID : The ID of a Guild.
 // emojiID : The ID of an Emoji.
 // name    : The Name of the Emoji.
-// roles   : The roles for which this emoji will be whitelisted, can be nil.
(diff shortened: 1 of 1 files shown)


restapi.go
```diff
@@ -1390,7 +1390,7 @@ func (s *Session) GuildEmojiEdit(guildID, emojiID, name string, roles []string)
 
 	data := struct {
 		Name  string   `json:"name"`
-		Roles []string `json:"roles,omitempty"`
+		Roles []string `json:"roles"`
```
FedorLap2006, line 1393: While it would be a breaking change, I think that using `*[]string` (both for the `roles` parameter and field) would be better.

FedorLap2006 (changes requested):
Sorry it took a long time.

FedorLap2006:
Thanks for your contribution!

PR #1026, opened 2021-11-05
feat: min and max constraints for slash command options

FedorLap2006:
https://github.com/discord/discord-api-docs/discussions/3327

### DIFF
--- a/examples/slash_commands/main.go
+++ b/examples/slash_commands/main.go
@@ -59,6 +59,8 @@ var (
 					Type:        discordgo.ApplicationCommandOptionInteger,
 					Name:        "integer-option",
 					Description: "Integer option",
+					MinValue:    1,
+					MaxValue:    10,
 					Required:    true,
 				},
 				{
--- a/interactions.go
+++ b/interactions.go
@@ -89,10 +89,15 @@ type ApplicationCommandOption struct {
 	// NOTE: This feature was on the API, but at some point developers decided to remove it.
 	// So I commented it, until it will be officially on the docs.
 	// Default     bool                              `json:"default"`
-	Required     bool                              `json:"required"`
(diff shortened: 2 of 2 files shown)


jolheiser:
I don't know that I particularly like `interface{}` for this. We can use `int` or `double` according to Discord docs, but I still think `float64` might make more sense. Maybe with some truncating internally if it's an integer field?

FedorLap2006:
> I don't know that I particularly like `interface{}` for this. We can use `int` or `double` according to Discord docs, but I still think `float64` might make more sense. Maybe with some truncating internally if it's an integer field?

The only downside of this approach - when value is `0` - `encoding/json` omits it, but shouldn't, and `interface{}` eliminates this problem.

iopred:
Using *float64 makes sense then, we've done that in a few places.

FedorLap2006:
It's quite inconvenient for the end user tho

FedorLap2006:
@CarsonHoffman What is your opinion on this?

FedorLap2006:
Alright, @iopred @CarsonHoffman should be ready for review

sentinelb51:
We've got this far:
`MinValue` and `MaxValue` as interface{} for `ApplicationCommandOption`

**What was NOT included are the following:**

Add a new enum (value 10) for ApplicationCommandOptionType
`ApplicationCommandOptionNumber ApplicationCommandOptionType = 10`

For `func (t ApplicationCommandOptionType) String() string`, add:
case ApplicationCommandOptionNumber: return "Number"

Summary:
`ApplicationCommandOptionNumber` should be used when you want to work with floats, otherwise `ApplicationCommandOptionInteger`

FedorLap2006:
Not entirely sure about merging this, because this is definitely going to break a lot of stuff. So will hold it for a couple of days.

Nv7-GitHub:
NEED THIS

PR #1030, opened 2021-11-19
fix shadowed variables that were preventing errors from being returned properly

danbrakeley:
I pulled latest and opened it up in VS Code, and there were 3 `ineffectual assignment to err (ineffassign)` errors from the `go-golangci-lint` linter, found in `restapi.go` on line 2548, `state.go` on line 857, and `state.go` on line 873.

The "ineffectual assignment to err" means that the code is assigning a value to a variable `err`, but then the variable goes out of scope without ever being used. Looking at the code, all 3 cases are due to the `:=` assignment operator creating a second `err` variable in the local scope, instead of re-using the `err` that is a named return value.

More info: [Golang Variable Shadowing](https://rpeshkov.net/blog/golang-variable-shadowing/).

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2536,7 +2536,7 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 // appID       : The application ID.
 // interaction : Interaction instance.
 // resp        : Response message data.
-func (s *Session) InteractionRespond(interaction *Interaction, resp *InteractionResponse) (err error) {
+func (s *Session) InteractionRespond(interaction *Interaction, resp *InteractionResponse) error {
 	endpoint := EndpointInteractionResponse(interaction.ID, interaction.Token)
 
 	if resp.Data != nil && len(resp.Data.Files) > 0 {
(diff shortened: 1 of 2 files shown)


FedorLap2006:
Oh, wow, I totally forgot about this while I was writing that function, thank you.

CarsonHoffman:
I'm not sure that the changes in `state.go` actually do anything—yes, we're shadowing `err`, but the returns immediately after those calls then refer to the inner `err`, no? Shadowing `err` is a common and typically innocuous practice. I agree that the `InteractionRespond` case is an issue.

danbrakeley:
@CarsonHoffman So in those two cases in `state.go`, you are right, the code behaves the same with this PR. But if you look at the rest of the function, my changes bring those case blocks consistent with the way the rest of the switch is written. And it removes linter errors, which means the linter output will be cleaner and easier to spot real issues. Also, removing the shadowing now means less chance of a minor change in the future not realizing that things are being shadowed.

FedorLap2006:
Thanks for your contribution!

PR #1043, opened 2021-12-02
fix: Fix debug message printing when debug is disabled.

racerxdl:
This message was happening a lot here, the thing is that when successfully read, the ioutil closes the body. So it will always trigger an error.

Also I think that message was mean't to be enabled only in debug. So I added the debug flag check :D


FedorLap2006:
Thanks for your contribution!

PR #1045, opened 2021-12-03
fix: Handle InteractionResponseDeferredChannelMessageWithSource in slash commands example. Also refactoring.

trofchik:
In /responses command `InteractionResponseDeferredChannelMessageWithSource` was listed in options but was not handled inside handler function. In result when typing command /responses and selecting option "Deffered response..." bot would either crash or go into infinite loop not knowing how to handle request from user.

Moved handler functions and commands definitions to separate files.


examples/slash_commands/handlers.go
```diff
@@ -0,0 +1,236 @@
...
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/bwmarrin/discordgo"
+)
+
+func SelectHandler(s *discordgo.Session, i *discordgo.InteractionCreate) {
```
FedorLap2006, line 12: This is not a critical problem, but I think this should be renamed to something else, it is quite confusing, since we have select components.

FedorLap2006 (approved):
Hi, everything looks good to me. But want to mention that we don't usually make go modules inside examples. Also I'm not sure if the slash commands example needs a readme, since it could be built with simple `go build`.

examples/slash_commands/handlers.go
trofchik: Ok, I will look into this on weekend.

trofchik: It's done. Take a look.

FedorLap2006 (approved):
Looks good to me

examples/slash_commands/main.go
```diff
@@ -1,386 +1,60 @@
...
-	RemoveCommands = flag.Bool("rmcmd", true, "Remove all commands after shutdowning or not")
-)
+// If you are getting "type not declared by package" error,
+// try executing "go get github.com/bwmarrin/discordgo@master".
+// It happens because "go get" pulls package's latest release instead
+// of the latest commit.
 
-var s *discordgo.Session
+const GuildID string = "Bot token goes here"
+const BotToken string = "Guild where the bot will be used"
```
FedorLap2006, line 18: I'm not sure if this is a good idea, since most of the time sensitive stuff is put into flags, plus it is easy to replace these values without recompiling the application again.

FedorLap2006 (changes requested):
Whoops, missed one thing

FedorLap2006:
Hi. Just wanted to check, if everything is okay and if there are any updates.
Since this would go pretty well with #1120 and #1129.

trofchik:
Sorry, no updates. Busy doing other things.

trofchik:
Hi, I've started looking into this again. What should I do in presence of #1120 and #1129?

FedorLap2006:
> Hi, I've started looking into this again. What should I do in presence of #1120 and #1129?

Good to know! Just wait until they'll be merged. However, while you wait, you can integrate the changes from the base branch.

trofchik:
Closed it accidentally. I'm working on it right now and will reopen PR once I have something to commit.

trofchik:
Changes from #1129 are integrated. If you decide to get #1120 done first it's fine, I can wait.

examples/slash_commands/handlers.go
```diff
@@ -0,0 +1,288 @@
...
+	"log"
+	"strings"
+	"time"
+
+	"github.com/bwmarrin/discordgo"
+)
+
+func Login(ds *discordgo.Session, r *discordgo.Ready) {
+	log.Printf("Logged in as: %v#%v", ds.State.User.Username, ds.State.User.Discriminator)
+}
```
FedorLap2006, line 14: I think having anonymous handler would be better in this case, since it's just oneliner.

examples/slash_commands/handlers.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	defer removeCommandsOnFailure(s, i.GuildID)
+	// This way commands will be deleted even if panic occurs inside one of the handlers
+	h(s, i)
+}
+
+func removeCommandsOnFailure(ds *discordgo.Session, guildID string) {
+	r := recover()
+	if r != nil {
```
FedorLap2006, line 31:
```suggestion
	if err := recover(); err != nil {
```

examples/slash_commands/handlers.go
```diff
@@ -0,0 +1,288 @@
...
+		deleteCommands(ds, guildID)
+		log.Fatal("Error occured inside command handler")
```
FedorLap2006, line 33:
```suggestion
		log.Fatalf("Error occured inside command handler: %s", err)
```

examples/slash_commands/handlers.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+}
+
+func Basic(s *discordgo.Session, i *discordgo.InteractionCreate) {
+	panic("test")
```
FedorLap2006, line 38:
```suggestion
```

examples/slash_commands/main.go
```diff
@@ -2,503 +2,77 @@ package main
...
-			s.FollowupMessageDelete(s.State.User.ID, i.Interaction, msg.ID)
+	flag.Parse()
 
-			s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
-				Content: "For those, who didn't skip anything and followed tutorial along fairly, " +
-					"take a unicorn :unicorn: as reward!\n" +
-					"Also, as bonus... look at the original interaction response :D",
-			})
-		},
+	ds, err := discordgo.New("Bot " + *BotToken)
```
FedorLap2006, line 49: Not sure why you named the session `ds`, the convention is usually `s`

examples/slash_commands/main.go
```diff
@@ -2,503 +2,77 @@ package main
...
 
 	log.Println("Adding commands...")
-	registeredCommands := make([]*discordgo.ApplicationCommand, len(commands))
-	for i, v := range commands {
-		cmd, err := s.ApplicationCommandCreate(s.State.User.ID, *GuildID, v)
-		if err != nil {
-			log.Panicf("Cannot create '%v' command: %v", v.Name, err)
-		}
-		registeredCommands[i] = cmd
+	createCommands(ds, *GuildID)
```
FedorLap2006, line 64: I think it would be better to keep this functionality in `main`, rather than separating into a function. Especially since it's called only once.

examples/slash_commands/main.go
```diff
@@ -2,503 +2,77 @@ package main
...
+func deleteCommands(ds *discordgo.Session, guildId string) {
+	for _, cmd := range commands {
+		err := ds.ApplicationCommandDelete(
+			ds.State.User.ID,
+			guildId,
+			cmd.ID,
+		)
+		if err != nil {
+			log.Fatalf("Could not delete command %s: %v\n", cmd.Name, err)
+		}
```
FedorLap2006, line 27: Since the function is related to commands, wouldn't it be better to move it to `commands.go`?

examples/slash_commands/main.go
```diff
@@ -2,503 +2,77 @@ package main
...
-	for i, v := range commands {
-		cmd, err := s.ApplicationCommandCreate(s.State.User.ID, *GuildID, v)
-		if err != nil {
-			log.Panicf("Cannot create '%v' command: %v", v.Name, err)
-		}
-		registeredCommands[i] = cmd
+	createCommands(ds, *GuildID)
+	if *RemoveCommands {
+		defer deleteCommands(ds, *GuildID)
+		defer log.Println("Removing commands...")
```
FedorLap2006, line 67: Since this example is often read by beginners, I think it would be better to make it into one defer, to avoid confusion.

examples/slash_commands/main.go
```diff
@@ -2,503 +2,77 @@ package main
...
-			},
-		},
-		{
-			Name:        "followups",
-			Description: "Followup messages",
-		},
+func createCommands(ds *discordgo.Session, guildId string) {
+	var err error
+	for i, cmd := range commands {
+		commands[i], err = ds.ApplicationCommandCreate(
```
FedorLap2006, line 34: Since it's a refactoring of the example I think it would be better to replace `ApplicationCommandCreate` with `ApplicationCommandBulkOverwrite` here. While using `Create` & `Delete` is a lot cleaner (you don't affect other commands) and straightforward, they get ratelimited pretty quickly when using them on multiple commands.

examples/slash_commands/main.go
```diff
@@ -2,503 +2,77 @@ package main
...
 
 	"github.com/bwmarrin/discordgo"
 )
 
-// Bot parameters
 var (
 	GuildID        = flag.String("guild", "", "Test guild ID. If not passed - bot registers commands globally")
 	BotToken       = flag.String("token", "", "Bot access token")
-	RemoveCommands = flag.Bool("rmcmd", true, "Remove all commands after shutdowning or not")
+	RemoveCommands = flag.Bool("rmcmd", true, "Weather to remove all commands after shutdowning or not")
```
FedorLap2006, line 15: Also speaking of performance and ratelimit issues, I think it would be better to set the default value to `false` here. That way each time you run the application you won't need to wait until it deletes the commands.

Otherwise, in addition to usage of `ApplicationCommandBulkOverwrite`, we can replace this parameter with `-clean`, which when enabled would use `ApplicationCommandCreate` & `ApplicationCommandDelete` instead of `ApplicationCommandBulkOverwrite`. That way it also might make sense to separate functionality into `createCommands` function instead of doing it directly in main.

examples/slash_commands/handlers.go
```diff
@@ -0,0 +1,288 @@
...
+	s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
+		Type: discordgo.InteractionResponseChannelMessageWithSource,
+		Data: &discordgo.InteractionResponseData{
+			Content: "Hey there! Congratulations, you've just executed your first slash command!",
+		},
+	})
+}
+
+func BasicWithFile(s *discordgo.Session, i *discordgo.InteractionCreate) {
+
```
FedorLap2006, line 48:
```suggestion
```

examples/slash_commands/handlers.go
```diff
@@ -0,0 +1,288 @@
...
+			Content: fmt.Sprintf(
+				msgformat,
+				margs...,
+			),
+		},
+	})
+}
+
+func Subcommands(s *discordgo.Session, i *discordgo.InteractionCreate) {
+
```
FedorLap2006, line 151:
```suggestion
```

examples/slash_commands/handlers.go
```diff
@@ -0,0 +1,288 @@
...
+	s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
+		Type: discordgo.InteractionResponseChannelMessageWithSource,
+		Data: &discordgo.InteractionResponseData{
+			Content: content,
+		},
+	})
+}
+
+func Responses(s *discordgo.Session, i *discordgo.InteractionCreate) {
+
```
FedorLap2006, line 180:
```suggestion
```

FedorLap2006 (changes requested):

FedorLap2006:
> Changes from #1129 are integrated. If you decide to get #1120 done first it's fine, I can wait.

I think yeah, better wait until that one is merged

examples/slash_commands/main.go
trofchik: Am I correct to assume that if we pass empty list of commands to `ApplicationCommandBulkOverwrite` it will delete all of the commands?

PR #1047, opened 2021-12-05
Added ChannelMessageSendRaw for raw JSON payload

Thurston00:
Example payload:

```json
{
  "content": "Hello, World!",
  "tts": false,
}
```

Useful for a microservice approach towards sending complex data that can't (currently) be correctly decoded in discordgo.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -1599,6 +1599,22 @@ func (s *Session) ChannelMessageSendComplex(channelID string, data *MessageSend)
 	return
 }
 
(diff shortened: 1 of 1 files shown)


FedorLap2006:
Can you provide more context why and what exactly can't be correctly encoded?

Thurston00:
To be honest, this PR is no longer required as the problem it attempted to resolve has since been fixed. I'd forgotten that it was open, cheers anyway! @FedorLap2006

FedorLap2006:
No worries. Glad the problem was resolved!.

PR #1052, opened 2021-12-11
remove omiteempty for SelectMenu MinValues

plally:
0 is a valid value for MinValues, it can be used to allow a user to remove all options on a select menu that defaults to having items selected.

this would change the behavior of the SelectMenu component, instead of defaulting to `1` (this is what discord defaults to if you dont provide a MinValues) it would default to 0 because we would always be passing min_values to discord. If this is not desirable I could instead make MinValues a pointer, I think like this is easier to use for most people though.

### DIFF
--- a/components.go
+++ b/components.go
@@ -171,7 +171,7 @@ type SelectMenu struct {
 	// The text which will be shown in the menu if there's no default options or all options was deselected and component was closed.
 	Placeholder string `json:"placeholder"`
 	// This value determines the minimal amount of selected items in the menu.
-	MinValues int `json:"min_values,omitempty"`
+	MinValues int `json:"min_values"`
 	// This value determines the maximal amount of selected items in the menu.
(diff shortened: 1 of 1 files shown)


FedorLap2006 (changes requested):
Great PR! But this unfortunately will break any usage of current select menus.
How it might be solved:

- You can put a comment with a NOTE that you must specify 1 for the menu to be a single value select. 
- You can change it to pointer, that would be self-describing, and also will give the developers errors about usage of current implementation.

FedorLap2006:
Also I think it would need additional feedback from the developers. So I'm going to hold it for a couple of days.

plally:
This was just the least likely fix to break stuff i could think off. I think we should move towards using pointers for all option fields in the future though.

plally:
this pr is no longer necessary

FedorLap2006:
Sorry, forgot to close it.

PR #982, opened 2021-07-30
v8 & v9 stuff

vysiondev:
First off, sorry about any confusion for my past PRs, I am still not too familiar with git. This time I tried and made sure everything was set up correctly before opening. If it's caused any trouble when I randomly close a PR, i apologize.

- Support for [threads](https://discord.com/developers/docs/topics/threads), introduced in API version 9. It covers all of the new endpoints for threads, as well as state updates when receiving gateway events related to threads. 
- Support for [stickers](https://discord.com/developers/docs/resources/sticker) 
- Support for [stage instances](https://discord.com/developers/docs/resources/stage-instance), a new type of voice channel.
- New function pertaining to invites, `InviteComplex()`, which gets invite information from the API with an option to include `with_counts` and `with_expiration`. Currently, `InviteWithCounts()` was the only way to retrieve # of uses of the invite, and there was no way to access the expiration date of the invite.
- ~~The new component, [Select Menus](https://discord.com/developers/docs/interactions/message-components#select-menus). The components example has also been updated accordingly.~~ Already implemented by a different PR.
- Added [missing audit log change keys and actions](https://discord.com/developers/docs/resources/audit-log).

PR #1058, opened 2021-12-18
Threads reloaded

FedorLap2006:
So, seeing _still ongoing_ problem with lack of threads implementation, I decided to make my own PR on them. 
This PR should include everything that's on threads currently.

This PR also includes some code from #982 and #922 (it is meant to be continuation of those two, but written from scratch). So credits to the authors.

### The roadmap
- [x] Gateway
- [x] REST
- [x] State
  - [x] CRUD
  - [x] Membership


FedorLap2006:
:construction: **This is still under development**

FedorLap2006:
Woah, that went pretty fast. Anyways, should be everything now.

FedorLap2006:
@CarsonHoffman Awaiting for your review.

orakaro:
Huge ++ to this work, and I would love to have Thread support on DiscordGo also. Appreciate if @CarsonHoffman could make this happen.

structs.go
```diff
@@ -286,25 +294,41 @@ type Channel struct {
...
+	Members []*ThreadMember `json:"-"`
 }
 
 // Mention returns a string which mentions the channel
 func (c *Channel) Mention() string {
 	return fmt.Sprintf("<#%s>", c.ID)
 }
 
+// Thread is a subchannel inside an existing channel
+type Thread = Channel
```
CarsonHoffman, line 325: I don't really see the point of this type alias. It seems like it's only going to confuse terminology further—knowing that a thread is simply a type of channel is, IMO, necessary to using them in many ways (e.g. knowing how to send a message to one). Below, the use of the alias in `IsThread` also is somewhat confusing—why would a method called `IsThread` on a receiver of type `Thread` return false? I'd just stick with `Channel`.

state.go
```diff
@@ -913,7 +1052,34 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 		if s.TrackChannels {
 			err = s.ChannelRemove(t.Channel)
 		}
-	case *MessageCreate:
+	case *ThreadCreate:
+		if s.TrackThreads {
+			msglog(LogDebug, 1, "thread created %s (%q) in %s", t.ID, t.Name, t.ParentID)
```
CarsonHoffman, line 1057: We don't do this for any other objects in the state; I'd recommend dropping these logs.

state.go
```diff
@@ -529,6 +561,101 @@ func (s *State) ChannelRemove(channel *Channel) error {
...
+				return err
+			}
+		}
+	}
+	s.Lock()
+	thread.MemberCount = tmu.MemberCount
+	s.Unlock()
+
+	return nil
+}
```
CarsonHoffman, line 646: This could easily add a member to the thread twice—if two updates happen concurrently, both could go through the deletion phase (the second one to execute wouldn't actually do anything), then both could go through the append phase. We really need one large lock on this entire method. I recognize that `MemberAdd` and `PresenceAdd` already take out locks, making this a bit more difficult—it might be worth adding unexported state methods that do the bulk of the work, and keep the exported methods as thin shims that take out the state lock. (Or, to be frank, it might be better to drop locking from the state altogether and have the caller handle the lock—currently, it's impossible to call multiple state functions in one locked sequence—but we might be too far down that road at this point.)

restapi.go
```diff
@@ -2162,11 +2162,20 @@ func (s *Session) WebhookDeleteWithToken(webhookID, token string) (st *Webhook,
 // webhookID: The ID of a webhook.
 // token    : The auth token for the webhook
 // wait     : Waits for server confirmation of message send and ensures that the return struct is populated (it is nil otherwise)
-func (s *Session) WebhookExecute(webhookID, token string, wait bool, data *WebhookParams) (st *Message, err error) {
+// threadID :	Sends a message to the specified thread within a webhook's channel. The thread will automatically be unarchived.
+func (s *Session) WebhookExecute(webhookID, token string, wait bool, threadID string, data *WebhookParams) (st *Message, err error) {
```
CarsonHoffman, line 2166: I'm not sure that changing the signature of the established method is the right way here—adding another method might be a better solution. If we're forcing users to change their code, it should be for a necessary reason, not for adding `""` in the vast majority of cases.

CarsonHoffman (changes requested):

state.go
FedorLap2006: Oh sorry, was debugging state, forgot to remove.

orakaro:
Bumped! Thanks for your work and how is this going? I am looking forward to have Thread supported any time soon!

FedorLap2006:
> Bumped! Thanks for your work and how is this going? I am looking forward to have Thread supported any time soon!

Well, I think by now everything is finished, just waiting when @CarsonHoffman does review on this.

restapi.go
```diff
@@ -2414,6 +2435,226 @@ func (s *Session) RelationshipsMutualGet(userID string) (mf []*User, err error)
...
+	_, err := s.RequestWithBucketID("PUT", endpoint, nil, endpoint)
+	return err
+}
+
+// RemoveThreadMember removes another member from a thread
+func (s *Session) RemoveThreadMember(threadID, memberID string) error {
+	endpoint := EndpointThreadMember(threadID, memberID)
+	_, err := s.RequestWithBucketID("DELETE", endpoint, nil, endpoint)
+	return err
+}
```
CarsonHoffman, line 2523: The pattern typically used in the library is <Noun><Verb>, not <Verb><Noun>. See `GuildMemberRemove`, `ChannelMessageSend`, etc.

state.go
```diff
@@ -529,6 +568,98 @@ func (s *State) ChannelRemove(channel *Channel) error {
...
+
+	// This algorithm filters out archived or
+	// threads which are children of channels in channelIDs
+	// and then it adds all synced threads to guild threads and cache
+	index := 0
+	for _, t := range guild.Threads {
+		if !t.ThreadMetadata.Archived && tls.ChannelIDs != nil {
+			for _, v := range tls.ChannelIDs {
+				if t.ParentID == v {
+					goto remove
```
CarsonHoffman, line 589: I'm not really sure why the `goto` is necessary here. I'd probably just `delete` and `continue` here, then elide the `continue` on line 594.

state.go
```diff
@@ -668,6 +799,12 @@ func (s *State) MessageAdd(message *Message) error {
 	if len(c.Messages) > s.MaxMessageCount {
 		c.Messages = c.Messages[len(c.Messages)-s.MaxMessageCount:]
 	}
+
+	if c.IsThread() {
+		c.MessageCount++
+		c.LastMessageID = message.ID
+	}
```
CarsonHoffman, line 806: We don't do this for normal channels, and the eventual consistency model of Discord's events means that these might not even be true. I'd just leave these properties alone and leave it to the semantics defined in the API.

state.go
```diff
@@ -913,7 +1056,31 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
...
+		}
+	case *ThreadMembersUpdate:
+		if s.TrackThreadMembers {
+			err = s.ThreadMembersUpdate(t)
+		}
+	case *ThreadListSync:
+		if s.TrackThreads {
+			err = s.ThreadListSync(t)
+		}
+	case *MessageCreate: // TODO: last message id in thread
```
CarsonHoffman, line 1083: Were you meaning to resolve this TODO?

events.go
```diff
@@ -73,6 +73,52 @@ type ChannelPinsUpdate struct {
...
 
+// ThreadCreate is the data for a ThreadCreate event.
+type ThreadCreate struct {
+	*Channel
+}
+
+// ThreadUpdate is the data for a ThreadUpdate event.
+type ThreadUpdate struct {
+	*Channel
+	// TODO: BeforeUpdate
```
CarsonHoffman, line 84: Were you meaning to resolve this TODO?

CarsonHoffman (changes requested):

state.go
FedorLap2006: I'm not sure if this is a good idea or not, because this might be confusing why the Thread in the state does not reflect on the last message. I mean, yeah, you could fetch the thread, but still, that is confusing.

FedorLap2006: Alright, I removed this, but not sure if this is a good idea. It might be confusing for a user why the thread does not contain the last message, like normal channels.

FedorLap2006: Also, since we're removing this we might need to remove this logic from `MessageRemove` as well

omegasome: This is marked as "outdated" so I'm unclear whether it's resolved or not

trag-stripe:
@CarsonHoffman and @FedorLap2006  👋  - checking in if this pull is ready to go? 
Thanks for building this project for the community and making all our lives easier!

state.go
FedorLap2006: It's not quite resolved, since I'm personally unsure about removal of this.

FedorLap2006:
> @CarsonHoffman and @FedorLap2006 wave - checking in if this pull is ready to go? Thanks for building this project for the community and making all our lives easier!

Almost. If I didn't forget anything then it's gonna be only the `LastMessageID` thing and that will be it.

omegasome:
So we're just waiting for @CarsonHoffman to provide their input on the `LastMessageID` thing?

FedorLap2006:
> So we're just waiting for @CarsonHoffman to provide their input on the `LastMessageID` thing?

Yes.

srikanthrc:
This would be really cool extension to  what the library already allows for. Would love to see this land sooner than later. @CarsonHoffman any chance we can bump this up in the queue of things to get to? thanks in advance,

state.go
CarsonHoffman: I don't think my opinion has changed here. I don't think that we should modify data from Discord that makes it impossible for a user to see what has canonically been sent. Thread updates should correct this message count. (I would agree that taking the logic out of message removal would make sense, as well.)

FedorLap2006:
@CarsonHoffman I think it's ready for the final review.

shrapx:
@FedorLap2006 I found a missing threadID assignment intended here:
https://github.com/FedorLap2006/discordgo/blob/threads/restapi.go#L2223

FedorLap2006:
After deeper reading of Discord's documentation and our current source code, I've come to the conclusion that Carson with right, about `LastMessageID` and it should not be changed. On other hand I will open a PR so it would be easier to work this out

FedorLap2006:
Uhh, sorry for commit base, but the merge of master didn't go well, so I had to rebase the branch to edit out some of the parts.

FedorLap2006:
There are some unresolved conflicts with `master` branch.
I'm going to resolve them in a couple of hours.

PR #1061, opened 2021-12-20
Add guild member timeouts

QPixel:
This PR implements the newly added [guild timeout](https://github.com/discord/discord-api-docs/pull/4075) feature

I am unsure about the permissions field, but I believe I followed the set precedent.


FedorLap2006:
I suggest you to make this PR draft as well, since the PR on docs is not complete yet, and the feature isn't released publicly yet

FedorLap2006:
And fun story, they got it merged, so you gotta undraft it now, lol @QPixel :smile:

QPixel:
lmao

CarsonHoffman:
Both the field name and the JSON property name should be disable**d**, not disable, no?

QPixel:
You are completely correct. I must've read it wrong while I was reading the docs pull request.

QPixel:
I'm converting this back to a draft because I did not account that timeouts can also have reasons. I'm going to implement that and test it, then mark it ready for review again.

tempor1s:
👋  Hey! Any update on getting this merged? Would love to use it in my project :)

QPixel:
@CarsonHoffman can I get a new review? I've tested it in my own personal bot and it works fine

CarsonHoffman:
I just pushed a commit that reconciles the change I recently made (and was inspired by this PR) to use `time.Time` in the library instead of our custom `Timestamp` type (which was somewhat confusing to create manually). I also added the ability to remove the timeout by having the parameter be a pointer, which it doesn't seem like your PR was able to do previously. Let me know what you think of these changes according to your original work with the feature.

QPixel:
Just tested it on my personal bot, these changes seem to be perfectly fine.

sentinelb51:
What about including `X-Audit-Log-Reason` and an optional `reason` argument?
You could add GuildMuteCreateWithReason since that's like GuildBanCreateWithReason

PR #1074, opened 2022-01-18
Presence.ClientStatus support

sentinelb51:
An additional field for the Presence struct containing a user's status for each type of 3 platforms: desktop, mobile or web [browser]

Reference: [client_status object](https://discord.com/developers/docs/topics/gateway#presence-update)


structs.go
```diff
@@ -740,11 +740,20 @@ type VoiceState struct {
 }
 
 // A Presence stores the online, offline, or idle and game status of Guild members.
+// ClientStatus will be nil if the user is offline or invisible
```
CarsonHoffman, line 743: This probably fits better right next to the field inside the struct

structs.go
```diff
@@ -740,11 +740,20 @@ type VoiceState struct {
...
-	Activities []*Activity `json:"activities"`
-	Since      *int        `json:"since"`
+	User         *User         `json:"user"`
+	Status       Status        `json:"status"`
+	Activities   []*Activity   `json:"activities"`
+	ClientStatus *ClientStatus `json:"client_status"`
+	Since        *int          `json:"since"`
+}
+
+// A ClientStatus struct contains the online, offline, idle or dnd status for each type of platform
```
CarsonHoffman, line 752: To match the format desired by the linter, this should just start "ClientStatus contains…"

CarsonHoffman (changes requested):
Just a few small documentation nits, but with a change of this size, that's often how it goes 😛

Issue #965, opened 2021-07-05
Add support for per-guild avatars.


FedorLap2006:
Suggest to add support of per-guild avatars. Yes, this feature is restricted for usage only of some group of people but it's already on the API, so I guess we can implement it.

andresperezl:
I just saw this is a duplicate of https://github.com/bwmarrin/discordgo/pull/985, but the original author seems to be inactive after being asked for some change

w.r.t https://github.com/bwmarrin/discordgo/pull/985#discussion_r769135345

I saw other string fields which are also nullable to be left as just `string` (like `Nick`) so I followed the same, but I can change `Avatar` to `*string` is that is what is desired.

cc @CarsonHoffman  @FedorLap2006

PR #985, opened 2021-08-02
Per-guild user avatars

vysiondev:
Related to https://github.com/discord/discord-api-docs/pull/3081 and closes #965. I anticipate i'll have to update this as more info about per-guild user avatars are released/if they get added to the docs.

PR #1077, opened 2022-01-22
Add Guild Member avatar

andresperezl:
Fixes https://github.com/bwmarrin/discordgo/issues/965

This PR adds support for the Guild-specific member avatar, if any.

To follow the same pattern as `*User.AvatarURL`, also add the same method for `*Member.AvatarURL` , and create an util method `avatarURL` to handle both.

Also, set the `*Member.GuildID` after the `GuildMember` requests, as this is not included with the returned object.

I have tested these changes with my own bot

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -642,6 +642,8 @@ func (s *Session) GuildMember(guildID, userID string) (st *Member, err error) {
 	}
 
 	err = unmarshal(body, &st)
+	// The returned object doesn't have the GuildID attribute so we will set it here.
+	st.GuildID = guildID
 	return
 }
 
--- a/structs.go
+++ b/structs.go
@@ -925,6 +925,9 @@ type Member struct {
 	// Whether the member is muted at a guild level.
 	Mute bool `json:"mute"`
 
+	// The avatar of the member for the guild, if any.
+	Avatar string `json:"avatar"`
+
 	// The underlying user on which the member is based.
 	User *User `json:"user"`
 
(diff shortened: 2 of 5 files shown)


andresperezl:
I have tested these changes with my own bot

andresperezl:
I see this is a duplicate of #985, but the original author seems to be inactive after being asked for some changes

w.r.t https://github.com/bwmarrin/discordgo/pull/985#discussion_r769135345

I saw other `string` fields which are also nullable to be left as simple `string` (like `Nick`) so I followed the same, but I can change it to `*string` if that is what is desired.

cc @FedorLap2006 @CarsonHoffman

restapi.go
```diff
@@ -790,6 +790,8 @@ func (s *Session) GuildMember(guildID, userID string) (st *Member, err error) {
 	}
 
 	err = unmarshal(body, &st)
+	// The returned object doesn't have the GuildID attribute so we will set it here.
+	st.GuildID = guildID
```
FedorLap2006, line 794: Good catch! Not sure tho if it would be the right PR to include the fix.

structs.go
```diff
@@ -794,6 +794,9 @@ type Member struct {
 	// Whether the member is muted at a guild level.
 	Mute bool `json:"mute"`
 
+	// The avatar of the member for the guild, if any.
+	Avatar string `json:"avatar"`
```
FedorLap2006, line 798: The doc-comment might be a little confusing. Since this is a hash, not the avatar url.

FedorLap2006 (changes requested):
Overall looks good to me. But it requires a couple of changes.

restapi.go
andresperezl: This is actually needed for the `*Member.AvatarURL` method, else it wouldn't be able to form the URL https://github.com/bwmarrin/discordgo/pull/1077/files#diff-ed32211335d38b2d817f745b55cd71fcc6f824cc578a680cdbed3912002876b6R829, but I can move it if necessary

FedorLap2006: Oh, then it's all good. No need to move it to different PR.

FedorLap2006 (approved):

FedorLap2006:
Tip: you can resolve conversations when they're outdated or finished, anyways, thank you for the PR. +1

FedorLap2006:
It's ready to be merged, but please note the conflict.

andresperezl:
@FedorLap2006 I have rebased my changes with latest `master`

structs.go
FedorLap2006: Still no change in the comment. Maybe you forgot to push?

structs.go
```diff
@@ -925,6 +925,9 @@ type Member struct {
 	// Whether the member is muted at a guild level.
 	Mute bool `json:"mute"`
 
+	// The avatar of the member for the guild, if any.
```
FedorLap2006, line 928: The comment is still unchanged.

FedorLap2006 (changes requested):

FedorLap2006 (approved):

FedorLap2006:
Thanks for the contribution!

PR #1083, opened 2022-02-04
Feature/add http

JordanDeBeer:
This PR adds support for receiving outgoing webhooks over HTTP.

It works by implementing the http.ServeHTTP method so users can pass it to any standard http server as to allow users to have routes outside of the discord interactions.

Sample usage:

```
package main

import (
	"fmt"
	"log"
	"net/http"
	"os"

)

var s *discordgo.Session
var AppId string
var GuildId string
var BotToken string
var PublicKey string

func init() {
	AppId = os.Getenv("APP_ID")
	GuildId = os.Getenv("GUILD_ID")
	BotToken = os.Getenv("BOT_TOKEN")
	PublicKey = os.Getenv("PUBLIC_KEY")

	var err error
	s, err = discordgo.New(fmt.Sprint("Bot ", BotToken))
	if err != nil {
		log.Fatalf("error starting: %v", err)
	}

}

func main() {
	_, ok := s.ApplicationCommandCreate(AppId, GuildId, &discordgo.ApplicationCommand{
		Name:        "test",
		Description: "test_command",
	})
	if ok != nil {
		log.Fatalf("cannot create command: %v", ok)
	}
	mux := discordgo.NewWebhookServer(s, PublicKey)
	mux.AddHandler("test", func(s *discordgo.Session, i *discordgo.Interaction) {
		s.InteractionRespond(i, &discordgo.InteractionResponse{
			Type: discordgo.InteractionResponseChannelMessageWithSource,
			Data: &discordgo.InteractionResponseData{
				Content: "test works",
			},
		})
	})
	log.Fatal(http.ListenAndServe(":8080", mux))
}

```

I threw this together pretty quickly, so I know there could be some cleanup, but wanted to get a PR open and get feedback and make sure this was something useful to the project before investing more time.

### DIFF
--- a/http.go
+++ b/http.go
@@ -9,28 +9,27 @@ import (
 	"net/http"
 )
 
-type WebhookServer struct {
-	sess      *Session
+type InteractionsHTTPServer struct {
+	*Session
 	publicKey ed25519.PublicKey
-	handlers  map[string]WebhookHandler
+	handlers  map[string]HandlerFunc
 }
 
-type WebhookHandler func(*Session, *Interaction)
+type HandlerFunc func(*Session, *InteractionCreate)
 
-func NewWebhookServer(sess *Session, pubKeyString string) *WebhookServer {
+func NewInteractionsHTTPServer(sess *Session, pubKeyString string) *InteractionsHTTPServer {
 	key, err := hex.DecodeString(pubKeyString)
 	if err != nil {
 		log.Fatal("couldn't parse public key string")
 	}
-	return &WebhookServer{sess: sess, publicKey: key, handlers: make(map[string]WebhookHandler)}
+	return &InteractionsHTTPServer{sess, key, make(map[string]HandlerFunc)}
 }
 
-func (s *WebhookServer) AddHandler(name string, fn WebhookHandler) {
+func (s *InteractionsHTTPServer) HandleFunc(name string, fn func(*Session, *InteractionCreate)) {
 	s.handlers[name] = fn
 }
 
-func (s *WebhookServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	log.Printf("http request: %+v", r)
+func (s *InteractionsHTTPServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if ok := VerifyInteraction(r, s.publicKey); !ok {
 		http.Error(w, "invalid request signature", http.StatusUnauthorized)
 		return
@@ -60,7 +59,7 @@ func (s *WebhookServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.Write(b)
 	case InteractionApplicationCommand:
 		if h, ok := s.handlers[interaction.ApplicationCommandData().Name]; ok {
-			h(s.sess, &interaction)
+			h(s.Session, &InteractionCreate{&interaction})
 		}
 	default:
 		log.Printf("unrecognized type: %v", interaction.Type.String())
--- a/http.go
+++ b/http.go
@@ -12,21 +12,29 @@ import (
 type WebhookServer struct {
 	sess      *Session
 	publicKey ed25519.PublicKey
+	handlers  map[string]WebhookHandler
 }
+
+type WebhookHandler func(*Session, *Interaction)
 
 func NewWebhookServer(sess *Session, pubKeyString string) *WebhookServer {
 	key, err := hex.DecodeString(pubKeyString)
 	if err != nil {
 		log.Fatal("couldn't parse public key string")
 	}
-	return &WebhookServer{sess: sess, publicKey: key}
+	return &WebhookServer{sess: sess, publicKey: key, handlers: make(map[string]WebhookHandler)}
+}
+
+func (s *WebhookServer) AddHandler(name string, fn WebhookHandler) {
+	s.handlers[name] = fn
 }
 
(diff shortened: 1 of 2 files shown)


http.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"crypto/ed25519"
+	"encoding/hex"
+	"encoding/json"
+	"io"
+	"log"
+	"net/http"
+)
+
+type WebhookServer struct {
```
FedorLap2006, line 12: Not sure if that's a good idea to call this a `WebhookServer` if it serves only interactions.
I would suggest to rename it to something like `InteractionsWebhookServer` or `InteractionsHTTPServer`

http.go
```diff
@@ -0,0 +1,68 @@
...
+	sess      *Session
+	publicKey ed25519.PublicKey
+	handlers  map[string]WebhookHandler
+}
+
+type WebhookHandler func(*Session, *Interaction)
```
FedorLap2006, line 18: Yet again, confusing naming. Aside from that, I think would be better to make the second parameter `*InteractionCreate`, cause that's a standard for events, and I think it would be good to keep it here as well.

http.go
```diff
@@ -0,0 +1,68 @@
...
+	"crypto/ed25519"
+	"encoding/hex"
+	"encoding/json"
+	"io"
+	"log"
+	"net/http"
+)
+
+type WebhookServer struct {
+	sess      *Session
```
FedorLap2006, line 13: maybe just embed the session like that?
```suggestion
	*Session
```

http.go
```diff
@@ -0,0 +1,68 @@
...
+		response := InteractionResponse{Type: InteractionResponsePong}
+
+		b, err := json.Marshal(response)
+		if err != nil {
+			log.Printf("could not marshal response: %v", err)
+			return
+		}
+		w.Write(b)
+	case InteractionApplicationCommand:
+		if h, ok := s.handlers[interaction.ApplicationCommandData().Name]; ok {
```
FedorLap2006, line 62: Subcommands support?

FedorLap2006 (changes requested):
Good idea! But I'm uncertain if that would fit into discordgo itself.
Also note my comments

FedorLap2006:
I might help you out with this, but collecting feedback first would be a good idea.

Benricheson101:
> Good idea! But I'm uncertain if that would fit into discordgo.

Personally, I like the idea of having some basic HTTP interaction features included in discordgo, like request validation. I like the idea of being able to use discordgo types and methods with serverless apps. However, I don't think a full command router would fit very well into discordgo.

Just my two cents.

JordanDeBeer:
> Good idea! But I'm uncertain if that would fit into discordgo.

This was my concern as well. I would personally like to see it, but it may be possible to implement this entirely in a separate package. 

This PR would have probably been better suited to a WIP PR because it does lack things like sub-commands (and maybe autocomplete?)

I know there's at least some demand from searching the discord for things like "http" and "outbound webhook" but it really comes down to if discordgo wants to be a library for handling all types of discord interactions, or just websocket ones.

It doesn't seem to be much code to maintain as the implementation is pretty simple, but with webhooks being a big focus of discord in the future it could become pretty sizeable.

Not sure who makes the final decision, but I'd be happy to publish my own lib reusing types from discordgo but with HTTP support.

I'll fix the changes @FedorLap2006 mentioned, but don't want to go too far without a feel for if this is in-scope for this project.

FedorLap2006:
> > Good idea! But I'm uncertain if that would fit into discordgo.
> 
> This was my concern as well. I would personally like to see it, but it may be possible to implement this entirely in a separate package.
> 
> This PR would have probably been better suited to a WIP PR because it does lack things like sub-commands (and maybe autocomplete?)
> 
> I know there's at least some demand from searching the discord for things like "http" and "outbound webhook" but it really comes down to if discordgo wants to be a library for handling all types of discord interactions, or just websocket ones.
> 
> It doesn't seem to be much code to maintain as the implementation is pretty simple, but with webhooks being a big focus of discord in the future it could become pretty sizeable.
> 
> Not sure who makes the final decision, but I'd be happy to publish my own lib reusing types from discordgo but with HTTP support.
> 
> I'll fix the changes @FedorLap2006 mentioned, but don't want to go too far without a feel for if this is in-scope for this project.

I totally agree with you.

Well, the final decision is with @CarsonHoffman. So, we all just wait for his answer.

http.go
plally: I think the command handling should be separated from the http stuff so that if someone wanted they could pass in their own command handler instead of being forced to use the implementation here. 

Maybe you could pass an interface or function when you create the InteractionsHttpServer that simply takes an Interaction. And then discordgo could provide a default implementation

JordanDeBeer:
Hey there @CarsonHoffman, I might have some time to work on this soon if you think this would be worthwhile to be included in the project.

FedorLap2006:
Hey! Just wanted to say, after a bit of thinking, I don't think that adding it would be a good idea.
We try to provide only necessary stuff (structures, gateway, rest) to work with Discord API. So on my opinion a separate library would be better for such functionality.

FedorLap2006:
> Hey there @CarsonHoffman, I might have some time to work on this soon if you think this would be worthwhile to be included in the project.

Hi! After a bit of thinking, I don't think that adding it would be a good idea.
We try to provide only necessary stuff (structures, gateway, rest) to work with Discord API.
So in my opinion a separate library would be a better fit for such functionality.

JordanDeBeer:
No worries!

PR #1087, opened 2022-02-08
Dropping user account endpoints and functionality

FedorLap2006:
This PR drops all user account related functionality and endpoints.
Due to the fact it violates Discord ToS and due to [recent changes](https://github.com/discord/discord-api-docs/discussions/4456) in rules for Discord Developers Rommunity Resources it needs to be dropped.


discord.go
```diff
@@ -75,87 +55,8 @@ func New(args ...interface{}) (s *Session, err error) {
 	s.Identify.Properties.OS = runtime.GOOS
 	s.Identify.Properties.Browser = "DiscordGo v" + VERSION
 	s.Identify.Intents = MakeIntent(IntentsAllWithoutPrivileged)
```
vidhanio, line 57: if this is being refactored anyways, `MakeIntent` is a no-op and not needed to be used anymore.

FedorLap2006: Done.

bwmarrin:
I'm inclined to go forward to dropping user account exclusive code - but I'm open to discussion here and thoughts from anyone that thinks this will cause them any issues.  So we'll leave this open for a few days at least to allow for that.

FedorLap2006:
Should be all good now!

PR #1100, opened 2022-02-16
fix(interactions): DefaultPermission should not be a bool ptr

Lucaskyy:
DefaultPermission should not be a bool pointer.

You can no longer pass in `true` or `false`, it **must** be a variable and you need to take the address of it. This absolutely makes no sense.

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -35,7 +35,7 @@ type ApplicationCommand struct {
 	Version           string                 `json:"version,omitempty"`
 	Type              ApplicationCommandType `json:"type,omitempty"`
 	Name              string                 `json:"name"`
-	DefaultPermission *bool                  `json:"default_permission,omitempty"`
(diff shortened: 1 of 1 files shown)


FedorLap2006:
Additionally, the same will happen when #1026 will be merged. It has `*float64` fields to allow json encoder to omit them as well, to not set a max/min value.

FedorLap2006:
Also, @Lucaskyy, you can read more about this behavious [here](https://eli.thegreenplace.net/2020/optional-json-fields-in-go/)

Lucaskyy:
Ah yeah you're right, it's inconvenient but since there's no other way to do this in Go we'll have to live with it :)

FedorLap2006:
There is a couple of ways, but this one is the easiest, since it requires just one wrapper function.

PR #1101, opened 2022-02-20
Use Logger in restapi.go

dschmidt:
There are way more occurences of `log.Print(...)` in `restapi.go`

I'm just not sure how to go about for fixing those. They are not so important because they are hidden behind the `Session.Debug` flag anyway. They  are basically for printing out whole http requests to the console.

I played around with introducing a new `LogLevel` `LogTrace` and not using the `Session.Debug` flag for deciding, whether those messages should be logged or not.
We could keep `Session.Debug` for api compatibility, but it would still be a behavioral change ... so not sure, what you would like there.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -109,7 +109,7 @@ func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b 
 	defer func() {
 		err2 := resp.Body.Close()
 		if err2 != nil {
-			log.Println("error closing resp body")
+			s.log(LogDebug, "error closing resp body")
 		}
 	}()
 


FedorLap2006:
I think would be better to use `Debug` here. Since you probably won't really want to have this in error logs, unless if you're in `Debug` mode.

dschmidt:
Changed

FedorLap2006:
I meant `s.Debug`, not `s.log(LogDebug`

dschmidt:
`s.Debug` has a comment that says it's meant to go away ... in the end I don't care how we do it, I just want to get rid of this unconditional output on stdout.

By now I feel it might be easier, if you just push a fix yourself, as it's just a one liner anyway? :)

FedorLap2006:
> `s.Debug` has a comment that says it's meant to go away ... in the end I don't care how we do it, I just want to get rid of this unconditional output on stdout.
> 
> By now I feel it might be easier, if you just push a fix yourself, as it's just a one liner anyway? :)

Yeah, I think the best solution here is to use `s.Debug`, and just put a `TODO` to replace it to `s.log(LogDebug, ...)` or `s.log(LogError, ...)` in the future

FedorLap2006:
After a bit of thinking, I think I'll close this in favor of #1043.

dschmidt:
fine with me, happy we got rid of the output - thanks!

PR #1103, opened 2022-02-23
Removal procedure for slash commands example

Fahim-Ferdous:
Implemented command removing procedure on exit.


FedorLap2006:
Thanks for the contribution!

PR #1106, opened 2022-02-24
Remove `omitempty` for `Embeds`

fiatjaf:
Some context: https://discord.com/channels/613425648685547541/697489244649816084/946485657184911411


fiatjaf:
Added a new commit to allow `TextInputs` to have `required: false`, otherwise they're assumed as required by Discord.

FedorLap2006:
> Added a new commit to allow `TextInputs` to have `required: false`, otherwise they're assumed as required by Discord.

Good catch! I totally overlooked this. 
Sorry that review took a long time.

PR #929, opened 2021-05-12
Add missing Interaction field to Message struct

SteMak:

PR #1112, opened 2022-02-26
add interaction field to message struct

TopiSenpai:
This pr adds the interaction field to the message struct 
https://discord.com/developers/docs/resources/channel#message-object-message-structure

there is another pr for this which seems to be dead and is wrongly implemented https://github.com/bwmarrin/discordgo/pull/929

### DIFF
--- a/message.go
+++ b/message.go
@@ -529,13 +529,13 @@ func (m *Message) ContentWithMoreMentionsReplaced(s *Session) (content string, e
 	return
 }
 
-// MessageInteraction contains information about the application command interaction which generated a message
+// MessageInteraction contains information about the application command interaction which generated the message.
 type MessageInteraction struct {
 	ID   string          `json:"id"`
 	Type InteractionType `json:"type"`
 	Name string          `json:"name"`
 	User *User           `json:"user"`
 
-	// Member is only present when the interaction is from a guild
+	// Member is only present when the interaction is from a guild.
 	Member *Member `json:"member"`
 }
--- a/message.go
+++ b/message.go
@@ -534,6 +534,6 @@ type MessageInteraction struct {
 	ID     string          `json:"id"`
 	Type   InteractionType `json:"type"`
 	Name   string          `json:"name"`
-	User   User            `json:"user"`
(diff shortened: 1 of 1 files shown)


message.go
```diff
@@ -135,6 +135,11 @@ type Message struct {
 	// If the field exists but is null, the referenced message was deleted.
 	ReferencedMessage *Message `json:"referenced_message"`
 
+	// MessageInteraction is sent on the message object when the message is a response to an Interaction without an existing message.
+	// This means responses to Message Components do not include this property,
+	// instead including a MessageReference as components always exist on preexisting messages.
```
FedorLap2006, line 140:
```suggestion
	// Is sent when the message is a response to an Interaction, without an existing message.
	// This means responses to message component interactions do not include this property,
	// instead including a MessageReference, as components exist on preexisting messages.
```

message.go
```diff
@@ -523,3 +528,12 @@ func (m *Message) ContentWithMoreMentionsReplaced(s *Session) (content string, e
 	})
 	return
 }
+
+// MessageInteraction contains information about the application command interaction which generated this message
```
FedorLap2006, line 532:
```suggestion
// MessageInteraction contains information about the application command interaction which generated a message
```

message.go
```diff
@@ -523,3 +528,12 @@ func (m *Message) ContentWithMoreMentionsReplaced(s *Session) (content string, e
...
+type MessageInteraction struct {
+	ID     string          `json:"id"`
+	Type   InteractionType `json:"type"`
+	Name   string          `json:"name"`
+	User   User            `json:"user"`
```
FedorLap2006, line 537:
```suggestion
	User   *User            `json:"user"`
```

FedorLap2006 (changes requested):

message.go
TopiSenpai: as far as I can tell the user field is always present
https://discord.com/developers/docs/interactions/receiving-and-responding#message-interaction-object-message-interaction-structure

PR #1115, opened 2022-02-27
Resume packets

FedorLap2006:
This PR makes sending resume packets easier.
It's draft because there might be more changes made regarding this problem.

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -41,7 +41,7 @@ func New(token string) (s *Session, err error) {
 		MaxRestRetries:         3,
 		Client:                 &http.Client{Timeout: (20 * time.Second)},
 		UserAgent:              "DiscordBot (https://github.com/bwmarrin/discordgo, v" + VERSION + ")",
-		sequence:               new(int64),
+		Sequence:               new(int64),
 		LastHeartbeatAck:       time.Now().UTC(),
 	}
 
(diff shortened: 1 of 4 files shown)


Benricheson101:
If resuming the gateway connection fails, the gateway responds with an opcode 9 ("Invalid Session") message. Discordgo catches this and automatically reidentifies. This is the expected result of receiving an opcode 9 message when _automatically_ resuming, but not ideal when manually resuming.

IMO, the best way around this would probably be to add a dedicated `Session::Resume()` method that will attempt to resume and return an error if Discord responds with opcode 9.

The main use case for having such a method is to allow for zero-downtime restarts for very large bots. Due to the way resuming works, you must send a resume packet within a few seconds of disconnecting, or it will fail with opcode 9. If you are trying to resume a large number of shards, you must resume all of them at once, instead of doing it in batches like you do with identifies. If >16 shards fail to resume, they will get ratelimited when they automatically identify.

Relevant lines:
https://github.com/bwmarrin/discordgo/blob/9e96d380d4e1ddd1b6049064fabfb6c37a4b7b59/wsapi.go#L171-L178
https://github.com/bwmarrin/discordgo/blob/9e96d380d4e1ddd1b6049064fabfb6c37a4b7b59/wsapi.go#L542-L553

fr3fou:
Any updates on this?

PR #1117, opened 2022-02-27
feat: Add Guild Member Cover & Accent Color support

### LINKED ISSUE #1093: Adding User Banners?
Hey there. Banners have been out for a while but it seems that discordgo still has 0 support for them. Is there any ETA to when this will be added?

42Atomys:
I will add Guild Member Cover & Accent Color support to the library 

Resolve #1093

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -1170,8 +1170,8 @@ func (m *Member) AvatarURL(size string) string {
 
 }
 
-// BannerURL returns the URL of the member's cover image.
-//    size:    The size of the desired cover image as a power of two
+// BannerURL returns the URL of the member's banner image.
+//    size:    The size of the desired banner image as a power of two
 //             Image size can be any power of two between 16 and 4096.
 func (m *Member) BannerURL(size string) string {
 	return bannerURL(m.User.Banner, EndpointUserBanner(m.User.ID, m.User.Banner), EndpointUserBannerAnimated(m.User.ID, m.User.Banner), size)
--- a/structs.go
+++ b/structs.go
@@ -1170,13 +1170,6 @@ func (m *Member) AvatarURL(size string) string {
 
 }
 
-// BannerURL returns the URL of the member's banner image.
-//    size:    The size of the desired banner image as a power of two
-//             Image size can be any power of two between 16 and 4096.
-func (m *Member) BannerURL(size string) string {
-	return bannerURL(m.User.Banner, EndpointUserBanner(m.User.ID, m.User.Banner), EndpointUserBannerAnimated(m.User.ID, m.User.Banner), size)
-}
-
 // A Settings stores data for a specific users Discord client settings.
 type Settings struct {
 	RenderEmbeds           bool               `json:"render_embeds"`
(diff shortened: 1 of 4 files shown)


42Atomys:
Tested manually : works with static and animated user banner. 

@FedorLap2006 Ready for review

user.go
```diff
@@ -54,6 +54,13 @@ type User struct {
 	// Whether the user has multi-factor authentication enabled.
 	MFAEnabled bool `json:"mfa_enabled"`
 
+	// Banner the user's banner hash
```
FedorLap2006, line 57:
```suggestion
	// The hash of the user's banner
```

FedorLap2006 (changes requested):
Overall PR looks good to me, except a couple of things. Additionally, I see inconsistent naming of `cover` and `banner`. Please fix that.

user.go
```diff
@@ -54,6 +54,13 @@ type User struct {
...
+	// see https://discord.com/developers/docs/reference#image-formatting
```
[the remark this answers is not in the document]
42Atomys: "Each" time a field is about image formatting, this link is provided (`structs.go#805` for example)

It's because is user ?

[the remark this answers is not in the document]
42Atomys: I remove it

util.go
```diff
@@ -92,3 +92,19 @@ func avatarURL(avatarHash, defaultAvatarURL, staticAvatarURL, animatedAvatarURL,
...
+		URL = animatedCoverURL
+	} else {
+		URL = staticCoverURL
+	}
+
+	if size != "" {
+		return URL + "?size=" + size
+	}
+	return URL
+}
```
FedorLap2006, line 110:
```suggestion
func bannerURL(bannerHash, staticBannerURL, animatedBannerURL, size string) string {
	var URL string
	if bannerHash == "" {
		return ""
	} else if strings.HasPrefix(bannerHash, "a_") {
		URL = animatedBannerURL
	} else {
		URL = staticBannerURL
	}

	if size != "" {
		return URL + "?size=" + size
	}
	return URL
}
```

structs.go
```diff
@@ -1170,6 +1170,13 @@ func (m *Member) AvatarURL(size string) string {
 
 }
 
+// BannerURL returns the URL of the member's banner image.
+//    size:    The size of the desired banner image as a power of two
+//             Image size can be any power of two between 16 and 4096.
+func (m *Member) BannerURL(size string) string {
```
FedorLap2006, line 1176: Not sure if this would be necessary, since `Member` doesn't have `Banner` property

FedorLap2006 (changes requested):

FedorLap2006:
I'll give it a look tomorrow

structs.go
FedorLap2006: I think it would be better to just implement it when it goes out. DiscordGo is 1:1 mapping, so if API doesn't have it, we don't as well.

FedorLap2006:
Thank you for your contribution! Due to the fact that v0.24.0 is planned for today, I decided to make a couple of changes which I proposed to the PR myself. 

About the `Member.BannerURL` method, I think it would be best to implement it when it's going to be shipped. Additionally it would eliminate confusion on why it doesn't return customized guild banner.

Issue #752, opened 2020-03-23
Add support for context.Context


ewohltman:
I was considering a way to incorporate the usage of `context.Context` so that if a timeout I set when handling an event is exceeded, I can have finer grain control over canceling requests within a logical context and cleaning up the associated resources.

Simplified example:

```go
// VoiceStateUpdate is the callback function for the VoiceStateUpdate event from Discord
func (config *Config) VoiceStateUpdate(s *discordgo.Session, vsu *discordgo.VoiceStateUpdate) {
    ctx, cancelCtx := context.WithTimeout(context.Background(), 30 * time.Second)
    defer cancelCtx()

    user, err := s.UserWithContext(ctx, vsu.UserID)
    if err != nil {
        // Log err
        return
    }

    // Numerous additional API calls
}
```

The idea being that if the time to make the numerous API calls add up to exceed the timeout in my context before I finish the `VoiceStateUpdate` method, I would like to return from the method early.

Currently, the built-in `*http.Client` in `discordgo.Session` has an overall timeout set, but that governs each individual request. The usage of `context.Context` gives us a higher-level control over a logical grouping of requests.

NaniteFactory:
I was also looking for something like this. My bot has hundreds of requests being made concurrently and blocked by the rate limiter. I want to throw away some of the least important requests to stop them bottleneck other requests, but can’t find a way. Every request made by `(*Session)` should really be cancelable with `context.Context`.

orthoplex64:
The pull request was closed without comment; why was it closed? I too was looking for these "WithContext" variants of functions when I was reading the DiscordGo docs for the first time.

ewohltman:
@CarsonHoffman There might be demand for this, and `context.Context` is low level enough that `discordgo` should support it.

CarsonHoffman:
I'm more than happy to consider a solution that integrates contexts into the library; I'm not sure that the proposed solution was the best for overall maintainability and surface area of the API. Having something emulating `net.http.(*Request).WithContext` would be interesting, as it would allow it to be applied to any request with one additional method. I'm not sure how feasible this would be with the shared, concurrent nature of `Session`, but it might be something interesting to explore.

ewohltman:
[...]

PR #1119, opened 2022-02-28
Add context.Context parameter to all REST API functions

jalavosus:
Go's `context.Context` builtin is extremely common when performing HTTP requests of any kind. Currently, discordgo does not allow API consumers to pass a Context, which means that operations which might time out -- or should -- are more or less left in the void. 

In one project I work on, this has forced me to use the following workaround: 

```go
SendMessage(ctx context.Context, session *discordgo.Session, msg string, channelId string) error {
    ch := make(chan error)
    
    ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
    defer cancel()

    go func(ctx context.Context, session *discordgo.Session, msg string, ch chan<- error) {
        _, sendErr := session.ChannelMessageSend(channelId, msg)
        if sendErr != nil {
            ch <- errors.Wrapf(sendErr, "error sending message to channel ID %[1]s", channelId)
            return
        }

        ch <- nil
    }(ctx, session, msg, ch)

    for {
        select {
        case <-ctx.Done():
            return ctx.Err()
        case sendMsgErr := <-ch:
            return sendMsgErr
        }
    }
}
```

Even that isn't a great workaround, since whether or not the message actually ends up being sent is unknown - I just know that my context timed out. It could fail, which in the case of a context timeout is optimal, or it could send; in the case of a context timeout, functions which check for errors might queue up messages for a retry if they "fail", which would mean multiple sends. Multiple sends is anywhere from annoying to production blocking. 

Using a user-supplied `context.Context` for all REST API interactions therefore allows for idempotent interactions with the API, which can be extremely important depending on the use case. 

Issue reference: #752


FedorLap2006:
Hi. Thank you for the Pull Request. The only problem is that it would break a lot of existing code.
I'll give it a look nonetheless

FedorLap2006:
This PR would break 90% of the already existing code, so I'm not sure about merging it.
Additionally, note the checks.

jalavosus:
> While it would be good to have this in the long run, it would just break too much stuff.

I don't disagree, but I'm also not advocating this to be featured in a point release; that is to say, such a change would and should only be part of a major version release.

> P. S. Note the checks

I'll fix the formatting check when I'm able to, though I do appreciate the heads up on it

jalavosus:
@FedorLap2006 I do appreciate that you didn't immediately close the issue 😊 as a developer who uses this library in production, I very much understand -- and support -- the need to not introduce breaking changes in the short term. Heck, I'd be pissed if someone else introduced this change and wanted it in the next minor version release. 

That being said, the use case for user -supplied `context.Context` params is there and likely widely useful to more developers/API consumers than just myself. Even if not in the next major release, even one after it would.be far more valuable than letting the concept languish. 

Either way, thanks a bunch truly 🌈

caibirdme:
Hope this pr could be merged! We really need the context.Context to propagate tracing information and control the timeout

FedorLap2006:
Hi everyone!

In the latest release (v0.27), we've introduced the options, which included `discordgo.WithContext`. The option which achieves the same functionality as this PR outlined. Due to this, I would have to close this PR.

Thanks for patience from everyone who participated and waited for this. And thanks to @jalavosus for pointing out the issue in the first place.

PR #1120, opened 2022-03-04
clarify slash commands example code

switchupcb:
# What Was Changed
**Added comments to slash commands example with reference to API.**

**Fixed comments to responses example with reference to API.**

**Added comments to followups example with reference to API.**

**Fixed line reference with regards to deleting all commands created in the session vs. all commands in the application.**

**Clarified extraneous code (flags, etc) such that a beginner would understand.**

**Combined multiple init() references.**
> https://stackoverflow.com/questions/56039154/is-it-really-bad-to-use-init-in-go

_Reason: The use of init() in this file is excessive. Using functions would provide the reader with a scope that's easier to follow. Everything in init() could have been placed in a function called at the start of main(). This would clarify the code execution order. I didn't remove init() though._

**I've added a comment to the ready handler**, because this is an example and the actual action is unclear at a first-glance. The log states that you log in which is fine. However, it's important to understand that the handler only fires once the websocket is ready to receive events from Discord.

# What Wasn't Changed
Application commands are defined without comments. In contrast, their handlers maintain the comments description the definition of the command. You can fix this by moving comments that describe **how to define a command** (and its properties) to the `commands` literal from the `commandHandlers` literal.

# What Could Be Changed
A suggestion is to clarify the distinction between creating/removing commands and the actual handling of them. In this way, the reader understands that creating and removing a command (interfacing with the API) is **separate** from handling the response to an event (interaction in this case).

The code may be easier to follow if you create a folder for this example and separate each command into their own files: For example, `basic.go`, `subcommands.go`, `responses.go`, `followups.go`. Instead, the user currently has to sort through slice and map "literals" along with everythig else to understand what's going on.

Another way to implement this suggestion is to manage the runtime in one file and struct/func literals in another. These literals (representing commands and handlers) can be added to a single literal map or slice and returned _(or used in init()_). This is easily maintainable and promotes good architecture.

## Example
main.go: Collects all commands and handlers (maps/slices), adds them to listeners and starts the bot.

Folder [pingpong]
-`session.go`: defines functions `commands() []appcommand` and `handlers() map[string]func(session...)` which returns a literal map or slice for ping and pong commands and handlers.

---`ping.go` (defines literal /ping commands and /ping handlers and any methods specific to ping)

---`pong.go` (defines literal /pong and /pong handler, etc)

### DIFF
--- a/examples/slash_commands/main.go
+++ b/examples/slash_commands/main.go
@@ -12,28 +12,24 @@ import (
 	"github.com/bwmarrin/discordgo"
 )
 
-// Bot parameters
+// Parse the bot's parameters from the command line.
+// NOTE: Don't commit the BOT TOKEN on Git. This will give FULL ACCESS to your bot (to whoever can access your project).
+// Use an environment variable instead.
 var (
 	GuildID        = flag.String("guild", "", "Test guild ID. If not passed - bot registers commands globally")
 	BotToken       = flag.String("token", "", "Bot access token")
 	RemoveCommands = flag.Bool("rmcmd", true, "Remove all commands after shutdowning or not")
 )
 
-var s *discordgo.Session
-
-func init() { flag.Parse() }
-
-func init() {
-	var err error
-	s, err = discordgo.New("Bot " + *BotToken)
-	if err != nil {
-		log.Fatalf("Invalid bot parameters: %v", err)
-	}
-}
-
 var (
 	integerOptionMinValue = 1.0
 
+	// The discord session that represents your bot.
+	s *discordgo.Session
+
+	// Create the application commands (slash commands) the bot will add.
+	// https://discord.com/developers/docs/interactions/application-commands#application-commands
+	// https://discord.com/developers/docs/interactions/application-commands#registering-a-command
 	commands = []*discordgo.ApplicationCommand{
 		{
 			Name: "basic-command",
@@ -172,6 +168,8 @@ var (
 			Description: "Followup messages",
 		},
 	}
+
+	// Create the handlers for the commands defined in the commands slice.
 	commandHandlers = map[string]func(s *discordgo.Session, i *discordgo.InteractionCreate){
 		"basic-command": func(s *discordgo.Session, i *discordgo.InteractionCreate) {
 			s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
@@ -266,13 +264,12 @@ var (
 				},
 			})
 		},
-		"responses": func(s *discordgo.Session, i *discordgo.InteractionCreate) {
-			// Responses to a command are very important.
-			// First of all, because you need to react to the interaction
-			// by sending the response in 3 seconds after receiving, otherwise
-			// interaction will be considered invalid and you can no longer
-			// use the interaction token and ID for responding to the user's request
 
+		// The Initial Response to an Interaction must occur within 3 seconds of recieving it.
+		// Otherwise the interaction token will be invalidated (disables the interaction token and ID).
+		// Valid interaction tokens last for 15 minutes.
+		// https://discord.com/developers/docs/interactions/receiving-and-responding#responding-to-an-interaction
+		"responses": func(s *discordgo.Session, i *discordgo.InteractionCreate) {
 			content := ""
 			// As you can see, the response type names used here are pretty self-explanatory,
 			// but for those who want more information see the official documentation
@@ -323,10 +320,11 @@ var (
 				s.InteractionResponseDelete(s.State.User.ID, i.Interaction)
 			})
 		},
+
+		// Interaction tokens are valid for 15 minutes and can be used to send any amount of followup messages.
+		// Followup Messages allow you send or edit responses after an initial response.
+		// https://discord.com/developers/docs/interactions/receiving-and-responding#followup-messages
 		"followups": func(s *discordgo.Session, i *discordgo.InteractionCreate) {
-			// Followup messages are basically regular messages (you can create as many of them as you wish)
-			// but work as they are created by webhooks and their functionality
-			// is for handling additional messages after sending a response.
 
 			s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
 				Type: discordgo.InteractionResponseChannelMessageWithSource,
@@ -366,7 +364,21 @@ var (
 	}
 )
 
+// init() executes as soon as the package is imported.
+// In this case, before any other code is run.
 func init() {
+	// Parse the flags from the command-line.
+	flag.Parse()
+
+	// Instantiate a new Discord Bot Session.
+	var err error
+	s, err = discordgo.New("Bot " + *BotToken)
+	if err != nil {
+		log.Fatalf("Invalid bot parameters: %v", err)
+	}
+
+	// Add an event handler for created interactions (defined in the commandHandlers map).
+	// https://discord.com/developers/docs/topics/gateway#interaction-create
 	s.AddHandler(func(s *discordgo.Session, i *discordgo.InteractionCreate) {
 		if h, ok := commandHandlers[i.ApplicationCommandData().Name]; ok {
 			h(s, i)
@@ -375,22 +387,22 @@ func init() {
 }
 
 func main() {
+	// Create a log once the websocket is ready to recieve events from Discord.
 	s.AddHandler(func(s *discordgo.Session, r *discordgo.Ready) {
 		log.Printf("Logged in as: %v#%v", s.State.User.Username, s.State.User.Discriminator)
 	})
-	err := s.Open()
-	if err != nil {
-		log.Fatalf("Cannot open the session: %v", err)
+	if err := s.Open(); err != nil {
+		log.Fatalf("Cannot open the websocket connection to Discord: %v", err)
 	}
 
 	log.Println("Adding commands...")
-	registeredCommands := make([]*discordgo.ApplicationCommand, len(commands))
+	newCommands := make([]*discordgo.ApplicationCommand, len(commands))
 	for i, v := range commands {
-		cmd, err := s.ApplicationCommandCreate(s.State.User.ID, *GuildID, v)
+		command, err := s.ApplicationCommandCreate(s.State.User.ID, *GuildID, v)
 		if err != nil {
 			log.Panicf("Cannot create '%v' command: %v", v.Name, err)
 		}
-		registeredCommands[i] = cmd
+		newCommands[i] = command
 	}
 
 	defer s.Close()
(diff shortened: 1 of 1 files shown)


examples/slash_commands/main.go
```diff
@@ -402,22 +414,24 @@ func main() {
 
 	if *RemoveCommands {
 		log.Println("Removing commands...")
-		// // We need to fetch the commands, since deleting requires the command ID.
-		// // We are doing this from the returned commands on line 375, because using
-		// // this will delete all the commands, which might not be desirable, so we
-		// // are deleting only the commands that we added.
+		// Uncomment the following code to delete every command in the application.
+		// You must also change newCommands (line 427) to registeredCommands.
```
FedorLap2006, line 418: Note my comment for `newCommands`.

examples/slash_commands/main.go
```diff
@@ -402,22 +414,24 @@ func main() {
...
+		//
+		// Fetch the created command IDs from the Discord API.
 		// registeredCommands, err := s.ApplicationCommands(s.State.User.ID, *GuildID)
 		// if err != nil {
 		// 	log.Fatalf("Could not fetch registered commands: %v", err)
 		// }
 
-		for _, v := range registeredCommands {
+		// Otherwise, only delete the newly created commands for this session.
+		// These are stored in the new []*discordgo.ApplicationCommand on line 405.
```
FedorLap2006, line 427:
```suggestion
		// They are stored in the newCommands, which are defined on line 387.
```

FedorLap2006 (changes requested):

examples/slash_commands/main.go
```diff
@@ -375,22 +387,22 @@ func init() {
...
 	})
-	err := s.Open()
-	if err != nil {
-		log.Fatalf("Cannot open the session: %v", err)
+	if err := s.Open(); err != nil {
+		log.Fatalf("Cannot open the websocket connection to Discord: %v", err)
 	}
 
 	log.Println("Adding commands...")
-	registeredCommands := make([]*discordgo.ApplicationCommand, len(commands))
```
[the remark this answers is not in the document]
switchupcb: registered could imply that these are all the registered commands for the bot. While these commands are registered, they are only new commands.

examples/slash_commands/main.go
```diff
@@ -366,7 +364,21 @@ var (
...
+	// Instantiate a new Discord Bot Session.
+	var err error
+	s, err = discordgo.New("Bot " + *BotToken)
+	if err != nil {
+		log.Fatalf("Invalid bot parameters: %v", err)
+	}
+
+	// Add an event handler for created interactions (defined in the commandHandlers map).
+	// https://discord.com/developers/docs/topics/gateway#interaction-create
 	s.AddHandler(func(s *discordgo.Session, i *discordgo.InteractionCreate) {
```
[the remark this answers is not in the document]
switchupcb: In the original commit, code in init() is left in init(). I have moved this call to main() at your request.

examples/slash_commands/main.go
```diff
@@ -266,13 +264,12 @@ var (
...
-		"responses": func(s *discordgo.Session, i *discordgo.InteractionCreate) {
-			// Responses to a command are very important.
-			// First of all, because you need to react to the interaction
-			// by sending the response in 3 seconds after receiving, otherwise
-			// interaction will be considered invalid and you can no longer
-			// use the interaction token and ID for responding to the user's request
 
+		// The Initial Response to an Interaction must occur within 3 seconds of recieving it.
+		// Otherwise the interaction token will be invalidated (disables the interaction token and ID).
+		// Valid interaction tokens last for 15 minutes.
```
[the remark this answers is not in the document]
switchupcb: We must clarify that they are valid because invalid interaction tokens (that result from a lack of response within 3 seconsd) do not last for 15 minutes.

examples/slash_commands/main.go
```diff
@@ -12,28 +12,24 @@ import (
...
-	}
-}
-
 var (
 	integerOptionMinValue = 1.0
 
+	// The discord session that represents your bot.
+	s *discordgo.Session
+
+	// Create the application commands (slash commands) the bot will add.
```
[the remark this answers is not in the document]
switchupcb: If you are referring to 27, the comment connects the concept of a session with the bot. This example uses `s` to refer to the session, but `s` could be replaced with `bot` and the code will still make sense. Otherwise, line 30 has been changed to the pattern you suggested ("The application...") which describes the commands variable.

examples/slash_commands/main.go
```diff
@@ -402,22 +416,24 @@ func main() {
...
 		log.Println("Removing commands...")
-		// // We need to fetch the commands, since deleting requires the command ID.
-		// // We are doing this from the returned commands on line 375, because using
-		// // this will delete all the commands, which might not be desirable, so we
-		// // are deleting only the commands that we added.
+		// Uncomment the following code to delete every command in the application.
+		// You must also change newCommands (line 430) to registeredCommands.
+		//
+		// Fetch the created command IDs from the Discord API.
 		// registeredCommands, err := s.ApplicationCommands(s.State.User.ID, *GuildID)
```
FedorLap2006, line 423: In addition to my comment on `registered`, I think it would be also fair to replace `:=` with `=`, due to the fact that when this line will be uncommented both variables will be shadowed.

It's not a critical flaw, but since this example is copied a lot, it would be good to fix it.

examples/slash_commands/main.go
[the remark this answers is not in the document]
FedorLap2006: Double meaning of `registered` here is intentional. Due to the fact that it might not contain only new commands if line 409 (423) is uncommented.

examples/slash_commands/main.go
```diff
@@ -323,10 +320,11 @@ var (
 				s.InteractionResponseDelete(s.State.User.ID, i.Interaction)
 			})
 		},
+
+		// Interaction tokens are valid for 15 minutes and can be used to send any amount of followup messages.
+		// Followup Messages allow you send or edit responses after an initial response.
```
FedorLap2006, line 325: I think it would be better to change wording here a little. Currently it's very confusing, especially for a beginner.
```suggestion
		// Followup messages allow for handling additional interaction related messages after the initial response.
```

examples/slash_commands/main.go
```diff
@@ -12,28 +12,24 @@ import (
...
-func init() {
-	var err error
-	s, err = discordgo.New("Bot " + *BotToken)
-	if err != nil {
-		log.Fatalf("Invalid bot parameters: %v", err)
-	}
-}
-
 var (
 	integerOptionMinValue = 1.0
```
FedorLap2006, line 25: I think it would be better to move it closer to `commands`

examples/slash_commands/main.go
```diff
@@ -366,31 +364,47 @@ var (
...
+
+	// Add an event handler for created interactions (defined in the commandHandlers map).
+	// https://discord.com/developers/docs/topics/gateway#interaction-create
+	s.AddHandler(func(s *discordgo.Session, i *discordgo.InteractionCreate) {
+		if h, ok := commandHandlers[i.ApplicationCommandData().Name]; ok {
+			h(s, i)
+		}
+	})
+
+	// Open the web socket connection to Discord to recieve events (and handle them).
```
FedorLap2006, line 395:
```suggestion
	// Open the websocket connection to Discord to recieve events (and handle them).
```

FedorLap2006 (changes requested):
Sorry, review took a while.
Overall everything looks good. However there are a couple of minor issues.

switchupcb:
Ready for review @FedorLap2006

switchupcb:
Options are recursive, so we must parse them into a map in a recursively. Otherwise, you run into an issue where the developer only accesses top-level options.

Otherwise, I recommend reverting the revert and adding a note that using a `for` will only parse the top-level options.

switchupcb:
Regarding errors, you should decide whether to use `panic` (which makes sense for an example) or a version `log.Print` which is more likely to be used in production.

examples/slash_commands/main.go
```diff
@@ -395,21 +430,24 @@ var (
 						"message will be deleted.",
 				})
 				if err != nil {
-					s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
+					_, err := s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
```
FedorLap2006, line 433:
```suggestion
					_, err = s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
```

examples/slash_commands/main.go
```diff
@@ -383,9 +415,12 @@ var (
 				},
 			})
 			if err != nil {
-				s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
+				_, err := s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
```
FedorLap2006, line 418:
```suggestion
				_, err = s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
```

examples/slash_commands/main.go
```diff
@@ -369,9 +398,12 @@ var (
 					Type: discordgo.InteractionResponseType(i.ApplicationCommandData().Options[0].IntValue()),
 				})
 				if err != nil {
-					s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
+					_, err := s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
```
FedorLap2006, line 401:
```suggestion
					_, err = s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
```

examples/slash_commands/main.go
```diff
@@ -395,21 +430,24 @@ var (
...
+
+		// Followup Messages allow you to send/edit/delete additional responses to an interaction after an initial response.
+		// https://discord.com/developers/docs/interactions/receiving-and-responding#followup-messages
 		"followups": func(s *discordgo.Session, i *discordgo.InteractionCreate) {
-			// Followup messages are basically regular messages (you can create as many of them as you wish)
-			// but work as they are created by webhooks and their functionality
-			// is for handling additional messages after sending a response.
 
-			s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
+			err := s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
```
FedorLap2006, line 450:
```suggestion
			err = s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
```

examples/slash_commands/main.go
```diff
@@ -419,59 +457,83 @@ var (
...
+			if err != nil {
+				panic(err)
+			}
+
 			msg, err := s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
 				Content: "Followup message has been created, after 5 seconds it will be edited",
 			})
 			if err != nil {
-				s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
+				_, err := s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
```
FedorLap2006, line 468:
```suggestion
				_, err = s.FollowupMessageCreate(s.State.User.ID, i.Interaction, true, &discordgo.WebhookParams{
```

FedorLap2006 (changes requested):

PR #1122, opened 2022-03-05
Payload for member edit

Jarvvski:
When editing a guild member, you can do much more than just update their roles, i.e. their nickname

[Discord docs](https://discord.com/developers/docs/resources/guild#modify-guild-member)

This change the method signature to expose a payload struct devs can use to set what they want to change on the member. We ommit any value from the future json payload if it was not set

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -699,13 +699,8 @@ func (s *Session) GuildMemberDeleteWithReason(guildID, userID, reason string) (e
 // GuildMemberEdit edits the roles of a member.
 // guildID  : The ID of a Guild.
 // userID   : The ID of a User.
-// roles    : A list of role ID's to set on the member.
-func (s *Session) GuildMemberEdit(guildID, userID string, roles []string) (err error) {
-
-	data := struct {
-		Roles []string `json:"roles"`
-	}{roles}
-
+// data     : Struct of GuildMemberEditData
+func (s *Session) GuildMemberEdit(guildID, userID string, data GuildMemberEditData) (err error) {
 	_, err = s.RequestWithBucketID("PATCH", EndpointGuildMember(guildID, userID), data, EndpointGuildMember(guildID, ""))
 	return
 }
--- a/restapi.go
+++ b/restapi.go
@@ -710,12 +710,18 @@ func (s *Session) GuildMemberEdit(guildID, userID string, roles []string) (err e
 	return
 }
 
-// GuildMemberEditComplex edits the roles & nickname of a member.
+// GuildMemberEditComplex edits the nickname and roles of a member.
 // guildID  : The ID of a Guild.
 // userID   : The ID of a User.
-// data     : Struct of GuildMemberEditData
-func (s *Session) GuildMemberEditComplex(guildID, userID string, data *GuildMemberEditData) (err error) {
-	_, err = s.RequestWithBucketID("PATCH", EndpointGuildMember(guildID, userID), data, EndpointGuildMember(guildID, ""))
+// data     : A GuildMemberEditData struct with the new nickname and roles
+func (s *Session) GuildMemberEditComplex(guildID, userID string, data GuildMemberEditData) (st *Member, err error) {
+	var body []byte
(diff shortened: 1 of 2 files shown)


restapi.go
```diff
@@ -699,13 +699,8 @@ func (s *Session) GuildMemberDeleteWithReason(guildID, userID, reason string) (e
...
 // userID   : The ID of a User.
-// roles    : A list of role ID's to set on the member.
-func (s *Session) GuildMemberEdit(guildID, userID string, roles []string) (err error) {
-
-	data := struct {
-		Roles []string `json:"roles"`
-	}{roles}
-
+// data     : Struct of GuildMemberEditData
+func (s *Session) GuildMemberEdit(guildID, userID string, data *GuildMemberEditData) (err error) {
```
FedorLap2006, line 703: I think still would be better to name it `GuildMemberEditComplex`. To avoid breaking current code.

structs.go
```diff
@@ -1566,6 +1566,17 @@ type UserGuildSettingsEdit struct {
...
 
+// A GuildMemberEditData stores a payload to use in Session.GuildMemberEdit
+// https://discord.com/developers/docs/resources/guild#modify-guild-member
+type GuildMemberEditData struct {
+
+	// Value to set user's nickname to
+	Nick string `json:"nick,omitempty"`
+
+	// Array of role ids the member is assigned
+	Roles []string `json:"roles,omitempty"`
```
FedorLap2006, line 1577: Since `roles` are optional, I think it would be best to make it `*[]string`.
As alternative we can put up a comment with notice about usage of `GuildMemberRoles` to reset member roles, but I think the first approach is better in this case.

FedorLap2006 (changes requested):

structs.go
```diff
@@ -1566,15 +1566,15 @@ type UserGuildSettingsEdit struct {
 	ChannelOverrides     map[string]*UserGuildSettingsChannelOverride `json:"channel_overrides"`
 }
 
-// A GuildMemberEditData stores a payload to use in Session.GuildMemberEdit
+// A GuildMemberEditData stores a payload to use in Session.GuildMemberEditComplex
 // https://discord.com/developers/docs/resources/guild#modify-guild-member
 type GuildMemberEditData struct {
```
FedorLap2006, line 1571: To keep consistency between all edit structures I think it would be best to name it
```suggestion
type GuildMemberEdit struct {
```

structs.go
```diff
@@ -1566,15 +1566,15 @@ type UserGuildSettingsEdit struct {
 	ChannelOverrides     map[string]*UserGuildSettingsChannelOverride `json:"channel_overrides"`
 }
 
-// A GuildMemberEditData stores a payload to use in Session.GuildMemberEdit
+// A GuildMemberEditData stores a payload to use in Session.GuildMemberEditComplex
```
FedorLap2006, line 1569:
```suggestion
// GuildMemberEditData stores data needed for member update
```

restapi.go
```diff
@@ -710,6 +710,15 @@ func (s *Session) GuildMemberEdit(guildID, userID string, roles []string) (err e
 	return
 }
 
+// GuildMemberEditComplex edits the roles & nickname of a member.
+// guildID  : The ID of a Guild.
+// userID   : The ID of a User.
+// data     : Struct of GuildMemberEditData
+func (s *Session) GuildMemberEditComplex(guildID, userID string, data *GuildMemberEditData) (err error) {
```
FedorLap2006, line 717: Modify endpoints always return the entity with modification. It seems you forgot to unmarshal & add it to return parameters.

Additionally I think it's unnecessary to make `data` a pointer.

restapi.go
```diff
@@ -710,6 +710,15 @@ func (s *Session) GuildMemberEdit(guildID, userID string, roles []string) (err e
 	return
 }
 
+// GuildMemberEditComplex edits the roles & nickname of a member.
```
FedorLap2006, line 713:
```suggestion
// GuildMemberEditComplex edits nick and roles of a member and returns new member data.
```

FedorLap2006 (changes requested):

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1123, opened 2022-03-05
Removed appID comment from InteractionRespond method which does not have this parameter

shawntoffel:
The comment for the InteractionRespond method mentions an appID despite the lack of an appID parameter. Likely copy paste. This change removes mention of the appID. 

image


FedorLap2006:
Thanks for your contribution!

PR #1128, opened 2022-03-09
move marshal functionality to json

switchupcb:
The library has no packages, so it's easier to keep everything related to it's functionality. Marshal/Unmarshal is the function of preparing information sent from/to to the API. Whereas individual files (`message.go`, `components.go`) serve a function of the Discord API. The latter is already implemented, so this pull request simply enforces that organization structure.

### DIFF
--- a/components.go
+++ b/components.go
@@ -2,7 +2,6 @@ package discordgo
 
 import (
 	"encoding/json"
-	"fmt"
 )
 
 // ComponentType is type of component.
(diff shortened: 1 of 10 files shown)


FedorLap2006:
I think that the idea behind putting the marshal functions beside the structures themselves is to better navigate through all of that.

FedorLap2006:
I'll consider it however.

FedorLap2006:
Just wanted to point out, this might be an interesting idea, but it would require rethinking library structure as whole

Issue #1126, opened 2022-03-08
Improve Options Handling to Reduce Security Errors


switchupcb:
Examples are a reflection of the decisions made during the API's creation. Improving the examples allows us to improve the API by identifying issues with assumptions we've made on _how the API is used._ An additional benefit to better examples is less need for support, and better code throughout the Discord Go scene.

The provided examples assume that the user enters options in the order we define them. Unfortunately, this is not the case which results in multiple errors. This demonstration uses the "options" handler in [slash commands example](https://github.com/bwmarrin/discordgo/blob/master/examples/slash_commands/main.go) to demonstrate this issue. However, this coding pattern (among others) reoccurs throughout the provided examples. 

## Fix
There are a few issues we can solve by refactoring the examples:
1. **Don't assume the best-case.** In this specific example, multiple `if len(Options) > n { parse expectedOption }` statements are used. This causes an error when the user inevitably includes an option in position 4 (that is _supposed_ to be in position 7). 
2. **Removing unnecessary panics.** Similar to reason 1, we may want to encourage an alternate way to handle issues with a command that doesn't crash the bot.
3. **Better ways to handle options.** Options are stored as a slice. This works when a single option is required. Less so when it's optional or there is more then one (of the same type). There are ways around this (using len, then looping, etc), but it seems more fitting to **use a `map[option name][option value]` that stores the value**. This makes it easy to check when an option exists, still allows use of len(), and provides an easier way to handle them.
4. **We can't always assume that Discord provides validation on the client-side.**

## Explanation
An error occurs because the value at `Options[0]` is not a string.

### Defining Options
```go
{
Name:        "options",
Description: "Command for demonstrating options",
Options: []*discordgo.ApplicationCommandOption{

    {
	    Type:        discordgo.ApplicationCommandOptionString,
	    Name:        "string-option",
	    Description: "String option",
	    Required:    true,
    },
    {
	    Type:        discordgo.ApplicationCommandOptionNumber,
	    Name:        "number-option",
	    Description: "Float option",
	    MaxValue:    10.1,
	    Required:    true,
    },
    {
	    Type:        discordgo.ApplicationCommandOptionBoolean,
	    Name:        "bool-option",
	    Description: "Boolean option",
	    Required:    true,
    },
...
```
### Handling
```go
margs := []interface{}{
    i.ApplicationCommandData().Options[0].StringValue(),
    i.ApplicationCommandData().Options[1].IntValue(),
[...]

PR #1129, opened 2022-03-09
Slash commands example: parse options into a map

switchupcb:
https://github.com/bwmarrin/discordgo/issues/1126 [Method 3](https://github.com/bwmarrin/discordgo/issues/1126#issuecomment-1062543250)

The resulting example output:
image

image

### DIFF
--- a/examples/slash_commands/main.go
+++ b/examples/slash_commands/main.go
@@ -201,7 +201,7 @@ var (
 			// Access options in the order provided by the user.
 			options := i.ApplicationCommandData().Options
 
-			// or convert the slice into a map
+			// Or convert the slice into a map
 			optionMap := make(map[string]*discordgo.ApplicationCommandInteractionDataOption, len(options))
 			for _, opt := range options {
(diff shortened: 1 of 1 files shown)


switchupcb:
If options can be recursive, we can implement that to the existing loop while unmarshalling. Otherwise, this should be good to go!

switchupcb:
If the subcommand example is correct, there can be parsed recursive options. However, they can't have the same name as other options. I will implement a recursive parse and example tomorrow.

switchupcb:
image
image
image
image

FedorLap2006 (changes requested):
As I said in the original issue, DiscordGo is 1:1 mapping of Discord API. I don't think that adding an extra option, which would be just a reflection of the original, would be a good idea.
That said, I think it would fair to still showcase parsing options into a map in the example.

Additionally, note my comments for some other stuff.

FedorLap2006 (approved):

PR #1130, opened 2022-03-10
Move RESTError to restapi.go

switchupcb:
The library has no packages, so it's easier to keep everything related to it's functionality. `types.go` only contains (error) functionality related to the the REST API, but it's not placed in `restapi.go`. Whereas [`wsapi.go`](https://github.com/bwmarrin/discordgo/blob/master/wsapi.go
) errors remain in `wsapi.go` functionality. The latter is already implemented, so this pull request simply enforces that organization structure.

Also discussed: https://github.com/bwmarrin/discordgo/discussions/1114


FedorLap2006:
Thanks for your contribution!

FedorLap2006:
Thanks for your contribution!

PR #1131, opened 2022-03-11
Reconfigure Golangci Linter

switchupcb:
These aren't technically breaking changes because no code logic is _supposed to be_ changed. However, there may be errors in translation. These improvements are beneficial over the long term for a more maintainable (and contributor-friendly) code base.

## Style
Removed over 200 occurrences of nakedreturns. A lot of unhandled errors _(that are now handled)_. A lot of setting errors (inside the if statement) before returning (as opposed to simply returning the error).

I placed recommendations for disabled linters in `.golangci.yml`. Edits by maintainers is allowed (as usual).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,8 @@ linters:
     - nlreturn # Recommended
     - stylecheck # Recommended
     - paralleltest
+    - predeclared
+    - goerr113 # Recommended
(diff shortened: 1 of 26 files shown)


switchupcb:
https://github.com/bwmarrin/discordgo/issues/1132

PR #1147, opened 2022-03-31
Update Go version in CI

Jleagle:


FedorLap2006:
It seems that GolangCI isn't very happy with these changes. I think we can update it's version

FedorLap2006:
I was also talking about manual install, which is right under the pre-defined action

FedorLap2006:
Thanks for your contribution!

PR #1150, opened 2022-04-03
Added Guild Members Search Function

Pancake-e:


restapi.go
```diff
@@ -2933,3 +2933,29 @@ func (s *Session) GuildScheduledEventUsers(guildID, eventID string, limit int, w
 	err = unmarshal(body, &st)
 	return
 }
+
+// SearchGuildMembers returns a list of guild member objects whose username or nickname starts with a provided string
+// guildID  : The ID of a Guild
+// query    : Query string to match username(s) and nickname(s) against
+// limit    : Max number of members to return (1-1000)
+func (s *Session) SearchGuildMembers(guildID, query string, limit int) (st []*Member, err error) {
```
FedorLap2006, line 2941: I think it would be better to name it `GuildMembersSearch`, to be consistent with the endpoint and other `Session` methods.

FedorLap2006 (changes requested):

restapi.go
```diff
@@ -2933,3 +2933,29 @@ func (s *Session) GuildScheduledEventUsers(guildID, eventID string, limit int, w
...
+	queryParams.Set("query", query)
+	if limit > 1 {
+		queryParams.Set("limit", strconv.Itoa(limit))
+	}
+
+	if len(queryParams) > 0 {
+		uri += "?" + queryParams.Encode()
+	}
+
+	body, err := s.RequestWithBucketID("GET", uri, nil, EndpointGuildMembersSearch(guildID))
```
FedorLap2006, line 2954: As `query` is always present and therefore `len(queryParams)` will be always greater than 0, I think we can do the following
```suggestion
	body, err := s.RequestWithBucketID("GET", uri + "?" + queryParams.Encode(), nil, uri)
```

FedorLap2006 (changes requested):

restapi.go
FedorLap2006, line 2954: Since `query` is always present in `queryParams`, the check is unnecessary here, the `len` will be always greater than zero.
```suggestion
	body, err := s.RequestWithBucketID("GET", uri + "?" + queryParams.Encode(), nil, uri)
```

restapi.go
```diff
@@ -2933,3 +2933,29 @@ func (s *Session) GuildScheduledEventUsers(guildID, eventID string, limit int, w
 	err = unmarshal(body, &st)
 	return
 }
+
+// GuildMembersSearch returns a list of guild member objects whose username or nickname starts with a provided string
+// guildID  : The ID of a Guild
+// query    : Query string to match username(s) and nickname(s) against
+// limit    : Max number of members to return (1-1000)
+func (s *Session) GuildMembersSearch(guildID, query string, limit int) (st []*Member, err error) {
```
FedorLap2006, line 2941: Since this method lists guild members, I think it would be better to place it between `GuildMembers` and `GuildMember`.

FedorLap2006 (changes requested):

FedorLap2006:
Thanks for your contribution!

PR #1151, opened 2022-04-09
Add option to disable rate-limited request retry

### LINKED ISSUE #1011: Add option to disable retrying of rate limited requests
Sometimes it's better for a rate limited request to fail with an error and let the library user decide what to do.

For example, I update a channel's topic periodically with info from an external data source. The rate limiting on channel edits is quite severe, two requests in 10 minutes, and hitting that limit causes DiscordGo to sleep for several minutes before retrying my now stale request. It would be better if I could send a new request with updated data instead.

Ideally retry would be disabled on a per-request basis, but I can't see a way to implement that without major API changes. Instead a field like `ShouldRetryOnRateLimit` could be added to Session, defaulting to true. Unfortunately this would force the user to decide the rate limiting policy globally for their application, and it couldn't be changed in a thread-safe manner. Finally, the automatic retry logic could be removed entirely and left to the user. Obviously this would be a breaking change. Other solutions would be appreciated.

njhanley:
Some requests are time sensitive and become stale if not completed quickly. It's better to let the library user choose how to handle a rate limit being exceeded in those cases. `ShouldRetryOnRateLimit` is a session option which disables automatic retry and instead returns an error the user can use to handle retry manually. 

Closes #1011.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -73,6 +73,17 @@ func (r RESTError) Error() string {
 	return "HTTP " + r.Response.Status + ", " + string(r.ResponseBody)
 }
 
+// RateLimitError is returned when a request exceeds a rate limit
+// and ShouldRetryOnRateLimit is false. The request may be manually
+// retried after waiting the duration specified by RetryAfter.
+type RateLimitError struct {
+	*RateLimit
+}
+
+func (e RateLimitError) Error() string {
+	return "Rate limit exceeded on " + e.URL + ", retry after " + e.RetryAfter.String()
+}
+
 // Request is the same as RequestWithBucketID but the bucket id is the same as the urlStr
 func (s *Session) Request(method, urlStr string, data interface{}) (response []byte, err error) {
 	return s.RequestWithBucketID(method, urlStr, data, strings.SplitN(urlStr, "?", 2)[0])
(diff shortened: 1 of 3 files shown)


restapi.go
```diff
@@ -186,14 +197,20 @@ func (s *Session) RequestWithLockedBucket(method, urlStr, contentType string, b
...
+			s.handleEvent(rateLimitEventType, &RateLimit{TooManyRequests: &rl, URL: urlStr})
 
-		response, err = s.RequestWithLockedBucket(method, urlStr, contentType, b, s.Ratelimiter.LockBucketObject(bucket), sequence)
+			time.Sleep(rl.RetryAfter)
+			// we can make the above smarter
+			// this method can cause longer delays than required
+
+			response, err = s.RequestWithLockedBucket(method, urlStr, contentType, b, s.Ratelimiter.LockBucketObject(bucket), sequence)
+		} else {
+			s.log(LogInformational, "Rate Limiting %s, retry is disabled", urlStr)
```
FedorLap2006, line 211: Since the function already returns the error, I think that we can just get rid of this log message.

restapi.go
```diff
@@ -73,6 +73,17 @@ func (r RESTError) Error() string {
...
 }
 
+// RateLimitError is returned when a request exceeds a rate limit
+// and ShouldRetryOnRateLimit is false. The request may be manually
+// retried after waiting the duration specified by RetryAfter.
+type RateLimitError struct {
+	*RateLimit
+}
+
+func (e RateLimitError) Error() string {
```
FedorLap2006, line 83:
```suggestion
// Error returns a rate limit error with rate limited endpoint and retry time.
func (e RateLimitError) Error() string {
```

FedorLap2006 (changes requested):

FedorLap2006:
Thanks for your contribution!

PR #1156, opened 2022-04-11
Updated RequestGuildMembers method and GuildMembersChunk event struct

Pancake-e:
I'm not sure if the "guild_id" field should be an array because according to the [documentation](https://discord.com/developers/docs/topics/gateway#request-guild-members-guild-request-members-structure), it should be a string. The endpoint accepts both array and string without errors.

### DIFF
--- a/events.go
+++ b/events.go
@@ -187,11 +187,13 @@ type GuildEmojisUpdate struct {
 
 // A GuildMembersChunk is the data for a GuildMembersChunk event.
 type GuildMembersChunk struct {
-	GuildID    string      `json:"guild_id"`
-	Members    []*Member   `json:"members"`
-	ChunkIndex int         `json:"chunk_index"`
-	ChunkCount int         `json:"chunk_count"`
-	Presences  []*Presence `json:"presences,omitempty"`
+	GuildID    string        `json:"guild_id"`
+	Members    []*Member     `json:"members"`
+	ChunkIndex int           `json:"chunk_index"`
+	ChunkCount int           `json:"chunk_count"`
+	NotFound   []json.Number `json:"not_found,omitempty"`
+	Presences  []*Presence   `json:"presences,omitempty"`
+	Nonce      string        `json:"nonce,omitempty"`
 }
 
 // GuildIntegrationsUpdate is the data for a GuildIntegrationsUpdate event.
--- a/events.go
+++ b/events.go
@@ -187,13 +187,13 @@ type GuildEmojisUpdate struct {
 
 // A GuildMembersChunk is the data for a GuildMembersChunk event.
 type GuildMembersChunk struct {
-	GuildID    string        `json:"guild_id"`
-	Members    []*Member     `json:"members"`
-	ChunkIndex int           `json:"chunk_index"`
-	ChunkCount int           `json:"chunk_count"`
-	NotFound   []json.Number `json:"not_found,omitempty"`
-	Presences  []*Presence   `json:"presences,omitempty"`
-	Nonce      string        `json:"nonce,omitempty"`
+	GuildID    string      `json:"guild_id"`
+	Members    []*Member   `json:"members"`
+	ChunkIndex int         `json:"chunk_index"`
+	ChunkCount int         `json:"chunk_count"`
+	NotFound   []string    `json:"not_found,omitempty"`
+	Presences  []*Presence `json:"presences,omitempty"`
+	Nonce      string      `json:"nonce,omitempty"`
 }
 
 // GuildIntegrationsUpdate is the data for a GuildIntegrationsUpdate event.
--- a/wsapi.go
+++ b/wsapi.go
@@ -409,12 +409,12 @@ func (s *Session) UpdateStatusComplex(usd UpdateStatusData) (err error) {
 }
 
 type requestGuildMembersData struct {
-	GuildIDs  []string       `json:"guild_id"`
-	Query     *string        `json:"query,omitempty"`
-	UserIDs   *[]json.Number `json:"user_ids,omitempty"`
-	Limit     int            `json:"limit"`
-	Nonce     string         `json:"nonce,omitempty"`
-	Presences bool           `json:"presences"`
+	GuildIDs  []string  `json:"guild_id"`
+	Query     *string   `json:"query,omitempty"`
+	UserIDs   *[]string `json:"user_ids,omitempty"`
+	Limit     int       `json:"limit"`
+	Nonce     string    `json:"nonce,omitempty"`
+	Presences bool      `json:"presences"`
 }
 
 type requestGuildMembersOp struct {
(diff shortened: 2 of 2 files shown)


wsapi.go
```diff
@@ -426,12 +428,22 @@ type requestGuildMembersOp struct {
 // query     : String that username starts with, leave empty to return all members
 // limit     : Max number of items to return, or 0 to request all members matched
 // presences : Whether to request presences of guild members
-func (s *Session) RequestGuildMembers(guildID string, query string, limit int, presences bool) (err error) {
+// userIDs   : Used to specify which users you wish to fetch
+// nonce     : Nonce to identify the Guild Members Chunk response
+func (s *Session) RequestGuildMembers(guildID, query string, limit int, presences bool, userIDs *[]string, nonce string) (err error) {
```
FedorLap2006, line 433: Otherwise, if we're keeping current approach, I think you can remove pointer from `userIDs` and do an internal check for length (`len(userIDs) > 0`).

wsapi.go
```diff
@@ -446,7 +458,7 @@ func (s *Session) RequestGuildMembers(guildID string, query string, limit int, p
 func (s *Session) RequestGuildMembersBatch(guildIDs []string, query string, limit int, presences bool) (err error) {
```
FedorLap2006, line 458: Looks like you forgot to add `userIDs` here.

FedorLap2006 (changes requested):

wsapi.go
```diff
@@ -426,12 +428,22 @@ type requestGuildMembersOp struct {
...
 // presences : Whether to request presences of guild members
-func (s *Session) RequestGuildMembers(guildID string, query string, limit int, presences bool) (err error) {
+// userIDs   : Used to specify which users you wish to fetch
+// nonce     : Nonce to identify the Guild Members Chunk response
+func (s *Session) RequestGuildMembers(guildID, query string, limit int, presences bool, userIDs *[]string, nonce string) (err error) {
+	// One of query or user_ids
+	var queryPt *string
+	if userIDs == nil {
+		queryPt = &query
+	}
```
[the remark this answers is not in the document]
Pancake-e: `omitempty` will not include `query` if the string is empty, hence the usage of pointer is to prevent the field being omitted when it is explicitly set to `""` (return all members). 

The endpoint will throw an error if both query and userIDs are not provided.

wsapi.go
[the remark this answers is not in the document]
Pancake-e: I think this approach is much cleaner compared to the pointer method.

FedorLap2006 (changes requested):

wsapi.go
```diff
@@ -425,29 +427,66 @@ type requestGuildMembersOp struct {
...
+// nonce     : Nonce to identify the Guild Members Chunk response
 // presences : Whether to request presences of guild members
-func (s *Session) RequestGuildMembersBatch(guildIDs []string, query string, limit int, presences bool) (err error) {
+func (s *Session) RequestGuildMembersBatchList(guildIDs []string, userIDs []string, limit int, nonce string, presences bool) (err error) {
+	if len(userIDs) == 0 {
+		return
+	}
+
+	// []json.Number is used here to prevent retry loop in
+	// 'ready' callback caused by non-digit characters
```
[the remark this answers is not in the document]
Pancake-e: The endpoint will return an error if there are non-digit characters in the `userIDs` string array. By using `json.Number`, the method will fail silently on this line instead of sending the invalid request to the endpoint.
https://github.com/bwmarrin/discordgo/blob/16efe0842b51604ee4bc7a647c6ad59f14984b7f/wsapi.go#L467

While the case above is not an issue when used normally, I found that it will cause an infinite retry loop when used in the 'ready' callback as demonstrated below (when `[]string` is used):
```
func ready(s *discordgo.Session, m *discordgo.Ready) {
	fmt.Println("Retry loop test")
        // Invalid userID "abc"
	s.RequestGuildMembersList(guildID, []string{"abc"}, 0, "", false)
}
```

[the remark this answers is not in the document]
FedorLap2006: I think it would be better to fix this in a separate PR.

events.go
```diff
@@ -187,11 +187,13 @@ type GuildEmojisUpdate struct {
...
-	GuildID    string      `json:"guild_id"`
-	Members    []*Member   `json:"members"`
-	ChunkIndex int         `json:"chunk_index"`
-	ChunkCount int         `json:"chunk_count"`
-	Presences  []*Presence `json:"presences,omitempty"`
+	GuildID    string        `json:"guild_id"`
+	Members    []*Member     `json:"members"`
+	ChunkIndex int           `json:"chunk_index"`
+	ChunkCount int           `json:"chunk_count"`
+	NotFound   []json.Number `json:"not_found,omitempty"`
```
FedorLap2006, line 194:
```suggestion
	NotFound   []string `json:"not_found,omitempty"`
```

wsapi.go
```diff
@@ -425,29 +427,55 @@ type requestGuildMembersOp struct {
...
 // presences : Whether to request presences of guild members
-func (s *Session) RequestGuildMembers(guildID string, query string, limit int, presences bool) (err error) {
+func (s *Session) RequestGuildMembers(guildID, query string, limit int, nonce string, presences bool) error {
+	return s.RequestGuildMembersBatch([]string{guildID}, query, limit, nonce, presences)
+}
+
+// RequestGuildMembersList requests guild members from the gateway
+// The gateway responds with GuildMembersChunk events
+// guildID   : Single Guild ID to request members of
+// userIDs   : Used to specify which users you wish to fetch
```
FedorLap2006, line 439:
```suggestion
// userIDs   : IDs of users to fetch
```

wsapi.go
```diff
@@ -425,29 +427,55 @@ type requestGuildMembersOp struct {
...
 	err = s.requestGuildMembers(data)
 	return
 }
 
-// RequestGuildMembersBatch requests guild members from the gateway
+// RequestGuildMembersBatchList requests guild members from the gateway
 // The gateway responds with GuildMembersChunk events
 // guildID   : Slice of guild IDs to request members of
-// query     : String that username starts with, leave empty to return all members
+// userIDs   : Used to specify which users you wish to fetch
```
FedorLap2006, line 469:
```suggestion
// userIDs   : IDs of users to fetch
```

wsapi.go
```diff
@@ -425,29 +427,55 @@ type requestGuildMembersOp struct {
...
 // limit     : Max number of items to return, or 0 to request all members matched
+// nonce     : Nonce to identify the Guild Members Chunk response
 // presences : Whether to request presences of guild members
-func (s *Session) RequestGuildMembersBatch(guildIDs []string, query string, limit int, presences bool) (err error) {
+func (s *Session) RequestGuildMembersBatchList(guildIDs []string, userIDs []string, limit int, nonce string, presences bool) (err error) {
```
FedorLap2006, line 473: According to the [API](https://discord.com/developers/docs/topics/gateway#request-guild-members), this functionality was removed.
```suggestion
//
//  NOTE: this function is deprecated, please use RequestGuildMembersList instead
func (s *Session) RequestGuildMembersBatchList(guildIDs []string, userIDs []string, limit int, nonce string, presences bool) (err error) {
```

wsapi.go
```diff
@@ -425,29 +427,55 @@ type requestGuildMembersOp struct {
...
+}
+
+// RequestGuildMembersBatch requests guild members from the gateway
+// The gateway responds with GuildMembersChunk events
+// guildID   : Slice of guild IDs to request members of
+// query     : String that username starts with, leave empty to return all members
+// limit     : Max number of items to return, or 0 to request all members matched
+// nonce     : Nonce to identify the Guild Members Chunk response
+// presences : Whether to request presences of guild members
+func (s *Session) RequestGuildMembersBatch(guildIDs []string, query string, limit int, nonce string, presences bool) (err error) {
```
FedorLap2006, line 454: According to the [API](https://discord.com/developers/docs/topics/gateway#request-guild-members), this functionality was removed.
```suggestion
//
//  NOTE: this function is deprecated, please use RequestGuildMembers instead
func (s *Session) RequestGuildMembersBatch(guildIDs []string, query string, limit int, nonce string, presences bool) (err error) {
```

FedorLap2006 (changes requested):

FedorLap2006 (approved):
LGTM! Thanks for your contribution!

PR #1166, opened 2022-04-20
fix: check if opus created

khodand:
Occasionally it causes panic when secretbox.open failed.
Below in the code we read `extlen := binary.BigEndian.Uint16(p.Opus[2:4])`
But opus can be nil.
I encountered this problem myself.


FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

Issue #914, opened 2021-04-24
Impossible to remove content in webhook message edit


plally:
The struct `WebhookEdit` does not allow deleting the content of a message. This happens because an empty string will not be included in the outputted json due to the omitempty tag.

this could be fixed by making `Content` a pointer in `WebhookEdit ` similar to `MessageEdit` but would break existing code.

PR #1174, opened 2022-05-01
make WebhookEdit fields pointers

plally:
These need to be pointers to allow removing embeds, content, and components from an existing message.

omitempty was added to components to allow keeping the existing components on a message when calling WebhookEdit


plally:
This addresses #914

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1177, opened 2022-05-05
fix: type of AuditLogActionApplicationCommandPermissionUpdate

ChristianGaertner:
AuditLogActionApplicationCommandPermissionUpdate was just of type `int` instead of `AuditLogAction`


FedorLap2006:
Thanks for your contribution!

PR #1178, opened 2022-05-05
Use `MessageFlags` as the type for `Flags` in `InteractionResponseData` and `WebhookParams`  

andresperezl:
According to the Discord documentation, the values for `InteractionResponseData.Flags` and `WebhookParams.Flags` are just a subset of the standard `MessageFlags`:

https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-messages
https://discord.com/developers/docs/resources/webhook#execute-webhook-jsonform-params
https://discord.com/developers/docs/resources/channel#message-object-message-flags

So use that type instead and put a note indicating the valid subset of flags for each of them. 

This is a breaking change, as the type is changing, so any code that is setting those flags will need to change to use the `MessageFlags` values instead, as it was done in the examples.

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -534,8 +534,10 @@ type InteractionResponseData struct {
 	Components      []MessageComponent      `json:"components"`
 	Embeds          []*MessageEmbed         `json:"embeds"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
-	Flags           uint64                  `json:"flags,omitempty"`
 	Files           []*File                 `json:"-"`
+
+	// NOTE: only MessageFlagsSuppressEmbeds and MessageFlagsEphemeral can be set
+	Flags MessageFlags `json:"flags,omitempty"`
 
 	// NOTE: autocomplete interaction only.
 	Choices []*ApplicationCommandOptionChoice `json:"choices,omitempty"`
--- a/interactions.go
+++ b/interactions.go
@@ -536,7 +536,7 @@ type InteractionResponseData struct {
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 	Files           []*File                 `json:"-"`
 
-	// NOTE: only MessageFlagsSuppressEmbeds and MessageFlagsEphemeral can be set
+	// NOTE: only MessageFlagsSuppressEmbeds and MessageFlagsEphemeral can be set.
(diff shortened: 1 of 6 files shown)


webhook.go
```diff
@@ -35,8 +35,8 @@ type WebhookParams struct {
 	Components      []MessageComponent      `json:"components"`
 	Embeds          []*MessageEmbed         `json:"embeds,omitempty"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
-	// NOTE: Works only for followup messages.
-	Flags uint64 `json:"flags,omitempty"`
+	// NOTE: Works only for followup messages and only MessageFlagsSuppressEmbeds can be set
```
FedorLap2006, line 38: Actually `flags` are supported outside of followup endpoints. Additionally, the followup create endpoint exclusively allows for `EPHEMERAL` flag.
So I think it would be better to rephrase this.
```suggestion
	// Only MessageFlagsSuppressEmbeds and MessageFlagsEphemeral can be set.
	// MessageFlagsEphemeral can only be set when using Followup Message Create endpoint.
```

FedorLap2006 (changes requested):

FedorLap2006 (changes requested):
LGTM. However you missed some raw value usages in the examples (including changed ones).

andresperezl:
Made the additional changes in the examples and rebased from master

interactions.go
```diff
@@ -534,9 +534,11 @@ type InteractionResponseData struct {
 	Components      []MessageComponent      `json:"components"`
 	Embeds          []*MessageEmbed         `json:"embeds"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
-	Flags           uint64                  `json:"flags,omitempty"`
 	Files           []*File                 `json:"-"`
 
+	// NOTE: only MessageFlagsSuppressEmbeds and MessageFlagsEphemeral can be set
```
FedorLap2006, line 539:
```suggestion
	// NOTE: only MessageFlagsSuppressEmbeds and MessageFlagsEphemeral can be set.
```

FedorLap2006 (changes requested):

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1179, opened 2022-05-06
Add `Dialer` property to `Session`

uWynell:
A little attempt to make this pull request.
Why would I need such an update? For example, if I would be able to pass custom `Dialer`, then I could use proxy ([click](https://stackoverflow.com/questions/33585587/creating-a-go-socks5-client)).
```go
dialSocksProxy, _:= proxy.SOCKS5("tcp", "proxy_ip", nil, proxy.Direct)
s, _:= discordgo.New("token")
s.Dialer = &websocket.Dialer{Dial: dialSocksProxy.Dial}
```
Sorry if I'm wrong. I'll be very happy to hear where exactly I made a mistake, if there's one.

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -17,6 +17,8 @@ import (
 	"net/http"
 	"runtime"
 	"time"
+
+	"github.com/gorilla/websocket"
 )
 
 // VERSION of DiscordGo, follows Semantic Versioning. (http://semver.org/)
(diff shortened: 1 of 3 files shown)


FedorLap2006 (changes requested):

FedorLap2006:
Thanks for the contribution!

P.S. While ideally the voice (see the comment above) related changes should have been implemented here, I'll be adding them in a separate PR, since you have disabled maintainer edits on the branch.

uWynell:
Sorry for inactivity, I hadn't even logged into my account since this PR.

FedorLap2006:
No problem, and as you can see, I've adressed it in the PR mentioned above.

PR #1091, opened 2022-02-11
Guild Template

QPixel:
This PR adds the Guild Template API to discordgo. 

API Reference: [discord api docs](https://discord.com/developers/docs/resources/guild-template)

I am curious if I should rename `EndpointGuildTemplateSync` since it technically does more than syncing a guild template's "state". Some feedback on that would be appreciated.

PR #1191, opened 2022-06-01
Fix GuildTemplate struct and documentation

QPixel:
This is a fix to PR https://github.com/bwmarrin/discordgo/pull/1091. It seems that the ``UsageCount`` field was incorrectly set to a string instead of a int. This PR has been made to fix this. I've tested this fix on my own machine.

Relevant Documentation: https://discord.com/developers/docs/resources/guild-template

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -952,7 +952,7 @@ type GuildTemplate struct {
 	Code string `json:"code"`
 
 	// The name of the template
-	Name *string `json:"name,omitempty"`
+	Name  `json:"name,omitempty"`
 
 	// The description for the template
 	Description *string `json:"description,omitempty"`
--- a/structs.go
+++ b/structs.go
@@ -952,7 +952,7 @@ type GuildTemplate struct {
 	Code string `json:"code"`
 
 	// The name of the template
-	Name  `json:"name,omitempty"`
+	Name string `json:"name,omitempty"`
 
 	// The description for the template
 	Description *string `json:"description,omitempty"`
--- a/structs.go
+++ b/structs.go
@@ -952,10 +952,10 @@ type GuildTemplate struct {
 	Code string `json:"code"`
 
 	// The name of the template
-	Name string `json:"name"`
+	Name *string `json:"name,omitempty"`
 
 	// The description for the template
-	Description string `json:"description"`
+	Description *string `json:"description,omitempty"`
 
 	// The number of times this template has been used
 	UsageCount int `json:"usage_count"`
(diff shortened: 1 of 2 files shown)


structs.go
```diff
@@ -958,7 +958,7 @@ type GuildTemplate struct {
 	Description string `json:"description"`
```
FedorLap2006, line 958: `omitempty` because:

image

`*string` because [Modify Guild Template](https://discord.com/developers/docs/resources/guild-template#modify-guild-template-json-params) requires it:

image

Can't use just `omitempty`, because it would be impossible to set the description to empty.

```suggestion
	Description *string `json:"description,omitempty"`
```
---

P. S. Same goes for the `name`, to work correctly in "Modify Guild Template" it should be a pointer.

FedorLap2006, line 958: Since we're fixing the `GuildTemplate` struct — this field requires `omitempty` and should be a pointer, to work correctly in [`Modify Guild Template`](https://discord.com/developers/docs/resources/guild-template#modify-guild-template) and it is also optional in [`Create Guild Template`](https://discord.com/developers/docs/resources/guild-template#modify-guild-template). endpoint

```suggestion
	Description *string `json:"description,omitempty"`
```
---

P. S. Same goes for the `name`, to work correctly in [`Modify Guild Template`](https://discord.com/developers/docs/resources/guild-template#modify-guild-template) it should be `omitempty`.

FedorLap2006 (changes requested):

FedorLap2006 (changes requested):
LGTM, however you have some unnecessary files commited.

structs.go
```diff
@@ -946,19 +946,19 @@ type GuildScheduledEventUser struct {
...
 
-// A GuildTemplate represents
+// A GuildTemplate represents a replicable template for guild creation
 type GuildTemplate struct {
 	// The unique code for the guild template
 	Code string `json:"code"`
 
 	// The name of the template
-	Name string `json:"name"`
+	Name *string `json:"name,omitempty"`
```
FedorLap2006, line 955: While making it a pointer would be good for consistency, it cannot be set to `""` (in contrast to `description`).
So I think just `omitempty` would be enough.
```suggestion
	Name string `json:"name,omitempty"`
```

FedorLap2006 (changes requested):

QPixel:
@FedorLap2006 Anything else you need changed?

FedorLap2006 (approved):

PR #1192, opened 2022-06-01
Added SelectMenu Values slice

JABirchall:
From what I know, this change is only for modals and is undocumented. 
They only return the selectmenu name and the values.


JABirchall:
Update from discord devs, they are waiting for the new release android to ship.
They also hinted that they might be rewriting modals ingeneral, as atm they use a webview, and they are thinking and making them native.
If they decide to move models to native that will delay future modal additions.

PR #1196, opened 2022-06-07
Add support for `with_localizations` flag

### LINKED ISSUE #1195: Support `with_localizations` query string for Get Application Commands endpoints
The endpoints "Get Global Application Commands" and "Get Guild Application Commands", which are implemented in `(*Session).ApplicationCommands(string, string)`, don't return `name_localizations` and `description_localizations` by default, as described in the [Discord Docs](https://discord.com/developers/docs/interactions/application-commands#retrieving-localized-commands). Instead, those endpoints supply `name_localized` and `description_localized`, which only contain one localization based on the request. The full `name_localizations` and `description_localizations` can only be obtained by supplying `with_localizations=true` in the query string. 

As far as I know this is not currently possible with DiscordGo. Furthermore `name_localized` and `description_localized` aren't accessible with the current `ApplicationCommand` struct.

EliasStar:
Fixes #1195

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2814,7 +2814,7 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 		endpoint = EndpointApplicationGuildCommands(appID, guildID)
 	}
 
-	body, err := s.RequestWithBucketID("GET", endpoint + "?with_localizations=true", nil, endpoint)
+	body, err := s.RequestWithBucketID("GET", endpoint+"?with_localizations=true", nil, endpoint)
 	if err != nil {
 		return
 	}
--- a/restapi.go
+++ b/restapi.go
@@ -2814,6 +2814,8 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 		endpoint = EndpointApplicationGuildCommands(appID, guildID)
 	}
 
+	endpoint += "?with_localizations=true"
+
 	body, err := s.RequestWithBucketID("GET", endpoint, nil, endpoint)
 	if err != nil {
 		return
--- a/restapi.go
+++ b/restapi.go
@@ -2814,9 +2814,7 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 		endpoint = EndpointApplicationGuildCommands(appID, guildID)
 	}
 
-	endpoint += "?with_localizations=true"
-
-	body, err := s.RequestWithBucketID("GET", endpoint, nil, endpoint)
+	body, err := s.RequestWithBucketID("GET", endpoint + "?with_localizations=true", nil, endpoint)
 	if err != nil {
 		return
 	}


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/bwmarrin/discordgo
+module github.com/EliasStar/discordgo
```
FedorLap2006, line 1:
```suggestion
module github.com/bwmarrin/discordgo
```

restapi.go
```diff
@@ -2808,12 +2808,16 @@ func (s *Session) ApplicationCommand(appID, guildID, cmdID string) (cmd *Applica
...
-func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*ApplicationCommand, err error) {
+func (s *Session) ApplicationCommands(appID, guildID string, withLocalizations bool) (cmd []*ApplicationCommand, err error) {
 	endpoint := EndpointApplicationGlobalCommands(appID)
 	if guildID != "" {
 		endpoint = EndpointApplicationGuildCommands(appID, guildID)
 	}
 
+	if withLocalizations {
+		endpoint += "?with_localizations=true"
+	}
```
FedorLap2006, line 2819: I totally overlooked the fact that the endpoint doesn't return full dicts without `with_localizations=true`.
However I think would be better to avoid the breaking change
```suggestion
    endpoint += "?with_localizations=true"
```

interactions.go
```diff
@@ -37,16 +37,20 @@ type ApplicationCommand struct {
...
 
 	// NOTE: Chat commands only. Otherwise it mustn't be set.
-
 	Description              string                      `json:"description,omitempty"`
 	DescriptionLocalizations *map[Locale]string          `json:"description_localizations,omitempty"`
 	Options                  []*ApplicationCommandOption `json:"options"`
+
+	// NOTE: Sometimes return by Discord API. Mustn't be set.
+	NameLocalized 		 *string     `json:"name_localized,omitempty"`
+	DescriptionLocalized *string     `json:"description_localized,omitempty"`
```
FedorLap2006, line 53: I think would be better not to include these yet, to avoid confusion. As this implementation does not have the functionality for the language customization (`Accept-Language` and `X-Discord-Locale` headers). 

Plus we have some plans on the implementation of header-based features. So I think we should leave them out, at least before the proper implementation of headers.

FedorLap2006 (changes requested):

go.mod
EliasStar: whoops, didn't know pull requests reference  HEAD. thought they reference the latest commit at the time of creation. my bad🤦

interactions.go
EliasStar: sounds good

restapi.go
```diff
@@ -2814,6 +2814,8 @@ func (s *Session) ApplicationCommands(appID, guildID string) (cmd []*Application
 		endpoint = EndpointApplicationGuildCommands(appID, guildID)
 	}
 
+	endpoint += "?with_localizations=true"
+
 	body, err := s.RequestWithBucketID("GET", endpoint, nil, endpoint)
```
FedorLap2006, line 2819: I think would be better to leave out the query params out of the bucket name (second `endpoint` specifies the bucket name that will be used when handling the ratelimit).
```suggestion
	body, err := s.RequestWithBucketID("GET", endpoint + "?with_localizations=true", nil, endpoint)
```

FedorLap2006 (changes requested):

FedorLap2006:
Please note the checks.

EliasStar:
I would appreciate if we could merge this, so I can retarget my bot to this repo away from the fork.

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1198, opened 2022-06-09
add Invite ExpiresAt field

Aderlx:
As you can see. What do you think about adding a ExpiresAt field ?

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -246,6 +246,8 @@ type Invite struct {
 	// will only be filled when using InviteWithCounts
 	ApproximatePresenceCount int `json:"approximate_presence_count"`
 	ApproximateMemberCount   int `json:"approximate_member_count"`
+
+	ExpiresAt time.Time `json:"expires_at"`
 }
 
 // ChannelType is the type of a Channel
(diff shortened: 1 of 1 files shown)


structs.go
```diff
@@ -246,6 +246,8 @@ type Invite struct {
 	// will only be filled when using InviteWithCounts
 	ApproximatePresenceCount int `json:"approximate_presence_count"`
 	ApproximateMemberCount   int `json:"approximate_member_count"`
+
+	ExpiresAt time.Time `json:"expires_at"`
```
FedorLap2006, line 250: According to docs this field is optional, maybe better would be to make it a pointer?
```suggestion
	ExpiresAt *time.Time `json:"expires_at"`
```

FedorLap2006 (changes requested):

structs.go
Aderlx: sure

Aderlx (commented):
optional expires_at

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1199, opened 2022-06-18
Allow setting bool fields of `ChannelEdit` to false

iokill:
Setting e.g. `ChannelEdit.Archived = false` is currently not possible as Go will treat a `false` value as empty and will omit the `archived` property from the resulting JSON object. The same goes for `Locked`, `NSFW` and `Invitable`.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -360,7 +360,7 @@ func (c *Channel) IsThread() bool {
 type ChannelEdit struct {
 	Name                 string                 `json:"name,omitempty"`
 	Topic                string                 `json:"topic,omitempty"`
-	NSFW                 bool                   `json:"nsfw,omitempty"`
+	NSFW                 *bool                  `json:"nsfw,omitempty"`
 	Position             int                    `json:"position"`
 	Bitrate              int                    `json:"bitrate,omitempty"`
 	UserLimit            int                    `json:"user_limit,omitempty"`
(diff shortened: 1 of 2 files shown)


structs.go
```diff
@@ -370,7 +370,7 @@ type ChannelEdit struct {
 
 	// NOTE: threads only
 
-	Archived            bool `json:"archived,omitempty"`
+	Archived            bool `json:"archived,bool"`
```
FedorLap2006, line 373: Haven't seen this JSON tag before, can you link the source where it is described?

FedorLap2006: But according to your description, I think you meant to do this:
```suggestion
	Archived            *bool `json:"archived,omitempty"`
```

iokill: Yeah sorry, I just copied what's done two lines below for `Locked` without really thinking much. :flushed: 

Using a pointer is probably the better idea here, but will change the user-facing API for `ChannelEdit``. If that is not an issue, let's do it like this.  The Discord API seems to really allow omitting this field, although the official docs do not really indicate this: Note the `boolean` and not `?boolean` for `archived` and other properties [here](https://discord.com/developers/docs/resources/channel#modify-channel-json-params-thread)

Should I also fix the other `bool` properties (`Locked`,`Invitable`, `NSFW`) here while I'm at it?

iokill: > Should I also fix the other bool properties (Locked,Invitable, NSFW) here while I'm at it?

Adapted to PR to do this.

FedorLap2006:
Please note the checks.

FedorLap2006 (approved):

PR #1205, opened 2022-06-23
Fix slash commands example

ck3g:
Add Application ID command flag to use it to create and delete
application commands.

[`Session.ApplicationCommandCreate()`](https://github.com/bwmarrin/discordgo/blob/a7f037862343f3ed29da9543b9ea4f8c57810620/restapi.go#L2717) and [`Session.ApplicationCommandDelete()`](https://github.com/bwmarrin/discordgo/blob/a7f037862343f3ed29da9543b9ea4f8c57810620/restapi.go#L2777) requires Application ID as a first
argument instead of a Bot's user ID

### DIFF
--- a/examples/slash_commands/main.go
+++ b/examples/slash_commands/main.go
@@ -17,6 +17,7 @@ import (
 var (
 	GuildID        = flag.String("guild", "", "Test guild ID. If not passed - bot registers commands globally")
 	BotToken       = flag.String("token", "", "Bot access token")
+	AppID          = flag.String("app", "", "Application ID")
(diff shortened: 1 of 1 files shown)


FedorLap2006 (commented):
Sorry for a late review. 
I'm not really sure if this is a necessary change, as application ID and Bot's user ID are always the same.
Having a separate argument for application id would make sense for a HTTP based interaction example though.

ck3g:
You're right. For some reason that example didn't work for me at first. Thanks for the hint

PR #1214, opened 2022-08-01
Add communication_disabled_until to GuildMemberParams

### LINKED ISSUE #1213: Timeout User
There is no way to timeout a user.

0xAozora:
This allows to timeout an user.

Closes #1213

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -1699,6 +1699,8 @@ type GuildMemberParams struct {
 	Nick string `json:"nick,omitempty"`
 	// Array of role ids the member is assigned
 	Roles *[]string `json:"roles,omitempty"`
+	// Timeout user until as ISO8601 timestamp
+	CommunicationDisabledUntil string `json:"communication_disabled_until,omitempty"`
 }
 
 // An APIErrorMessage is an api error message returned from discord


structs.go
```diff
@@ -1699,6 +1699,8 @@ type GuildMemberParams struct {
 	Nick string `json:"nick,omitempty"`
 	// Array of role ids the member is assigned
 	Roles *[]string `json:"roles,omitempty"`
+	// Timeout user until as ISO8601 timestamp
+	CommunicationDisabledUntil string `json:"communication_disabled_until,omitempty"`
```
FedorLap2006, line 1703: We usually use `time.Time` for timestamps.
Additionally this field must be both omittable and nullable (to unmute someone, a `null` value must be set).
There are two ways to implement that:
- Custom `MarshalJSON`
- `NullTime` type

I'm not really sure which one would be the best here. Additionally since there's already a `GuildMemberTimeout`, which serves the same purpose and more fit for the task, I think it would be the best to leave this field out for now.

FedorLap2006 (changes requested):

FedorLap2006:
Since `GuildMemberParams` had other unimplemented params and the `communication_disabled_until` needs custom marshalling for the `null`, I've decided to implement it all together in #1226. Therefore I'm going to close this PR.

0xAozora:
Cool!

PR #1230, opened 2022-08-18
Continue to next packet if received broken packet

ozraru:
Broken packet should be ignored and continue to next packet.
This goroutine should not exit until close.


FedorLap2006:
Hi, sorry for a delay, will check it today.

PR #1235, opened 2022-08-26
feat: state.GuildIDs() to get ids of guilds the bot is present in

telephrag:
To be able to get guildID at runtime thus, saving a hustle for end user of getting manually after enabling developer mod. Useful for self-hosted single guild bots.


FedorLap2006:
We already have `Session.State.Ready.Guilds` for that purpose.

PR #1241, opened 2022-09-12
Guild edit

Suhaibinator:


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/bwmarrin/discordgo
+module github.com/suhaibinator/discordgo
```
FedorLap2006, line 1:
```suggestion
module github.com/bwmarrin/discordgo
```

FedorLap2006 (changes requested):
LGTM, I think it would be better to separate out the formatting and `Icon` changes into different PRs.

FedorLap2006:
After looking more into documentation, this would be the incorrect way of modifying guild icon.
The guild icon must be encoded, as all images in the api, into base64 and that is why its type was string.

PR #1245, opened 2022-09-17
graceful shutdown: notify using syscall.SIGTERM

solodeathkill:
`os.Kill` can't be caught it's just there for killing other processes. `os.Kill` is like kill -9 which kills a process immediately. `syscall.SIGTERM` is equivalent to kill which allows the process time to cleanup.


Nv7-GitHub:
`syscall.SIGTERM` is already in that list

PR #1254, opened 2022-10-01
fix(structs): fix type of `GuildScheduledEvent*` enum values

invakid404:
Seems like the types of the enum values of `GuildScheduledEventStatus` and
`GuildScheduledEventEntityType` were overlooked and were left as `int`,
resulting in the need to cast them each time you want to use them.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -973,13 +973,13 @@ type GuildScheduledEventStatus int
 
 const (
 	// GuildScheduledEventStatusScheduled represents the current event is in scheduled state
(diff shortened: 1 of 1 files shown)


FedorLap2006:
> @FedorLap2006 how is that supposed to be a breaking change?

My bad. Usually change in types results in a breaking change, though in this case, you would be converting it everywhere already.

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1253, opened 2022-09-30
Add files to MessageEdit

ShadiestGoat:
This PR adds the ability to edit the attachments of a message, following the discord API.

PR #1256, opened 2022-10-03
fix(MultipartBodyWithJSON): form file field name format

FedorLap2006:
Currently `MultipartBodyWithJSON` uses `file<index>` format for the `name` of the form field for files. However this format is not correct, according to the [documentation](https://discord.com/developers/docs/reference#uploading-files). The correct format is `files[<index>]`.

The implementation was working all that time, because all the endpoints with support of attachments append them automatically and you do not need to reference them, however when you would try to, you would just get an "Unknown attachment" error. 
This PR should fix this problem.

Initially I discovered this when I was performing manual tests on #1253.

### DIFF
--- a/util.go
+++ b/util.go
@@ -51,7 +51,7 @@ func MultipartBodyWithJSON(data interface{}, files []*File) (requestContentType 
 
 	for i, file := range files {
 		h := make(textproto.MIMEHeader)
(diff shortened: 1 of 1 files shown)


FedorLap2006:
Whoops, forgot to push the local changes. It's `files`, not `file` 😅

PR #1260, opened 2022-10-19
Continue to next packet if received broken packet

ozraru:
Same as #1230 .
Sorry, I made a mistake and closed it.

#1166 has avoid panic but cause a silent fail.
This fixes it.
Broken packet should be ignored and continue to next packet.
This goroutine should not exit until close.


FedorLap2006:
Thanks for the contribution!

PR #1261, opened 2022-10-19
Overwrite entire emoji list on GuildEmojiUpdate

ozraru:
deleted emoji should be deleted from cache

### DIFF
--- a/state.go
+++ b/state.go
@@ -735,21 +735,6 @@ func (s *State) EmojisAdd(guildID string, emojis []*Emoji) error {
 	return nil
 }
 
-// EmojisUpdate updates entire emoji list for a guild id.
-func (s *State) EmojisUpdate(guildID string, emojis []*Emoji) error {
-	guild, err := s.Guild(guildID)
-	if err != nil {
-		return err
-	}
-
-	s.Lock()
-	defer s.Unlock()
-
-	guild.Emojis = emojis
-
-	return nil
-}
-
 // MessageAdd adds a message to the current world state, or updates it if it exists.
 // If the channel cannot be found, the message is discarded.
 // Messages are kept in state up to s.MaxMessageCount per channel.
@@ -1038,7 +1023,14 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 		}
 	case *GuildEmojisUpdate:
 		if s.TrackEmojis {
-			err = s.EmojisUpdate(t.GuildID, t.Emojis)
+			var guild *Guild
+			guild, err = s.Guild(t.GuildID)
+			if err != nil {
+				return err
+			}
+			s.Lock()
+			defer s.Unlock()
+			guild.Emojis = t.Emojis
 		}
 	case *ChannelCreate:
 		if s.TrackChannels {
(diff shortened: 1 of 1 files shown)


state.go
```diff
@@ -1023,7 +1038,7 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 		}
 	case *GuildEmojisUpdate:
 		if s.TrackEmojis {
-			err = s.EmojisAdd(t.GuildID, t.Emojis)
+			err = s.EmojisUpdate(t.GuildID, t.Emojis)
```
FedorLap2006, line 1041: Since the implementation is very simple and short, I think it would be better to implement it inline (see `GuildMembersChunk` and `GuildMemberRemove` as examples)

FedorLap2006 (changes requested):

state.go
```diff
@@ -735,6 +735,21 @@ func (s *State) EmojisAdd(guildID string, emojis []*Emoji) error {
 	return nil
 }
 
+// EmojisUpdate updates entire emoji list for a guild id.
```
FedorLap2006, line 738:
```suggestion
// EmojisUpdate updates emoji list for a guild.
```

FedorLap2006 (changes requested):

ozraru:
I fixed it, can you review please?

state.go
```diff
@@ -1023,7 +1023,14 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 		}
 	case *GuildEmojisUpdate:
 		if s.TrackEmojis {
-			err = s.EmojisAdd(t.GuildID, t.Emojis)
+			var guild *Guild
+			guild, err = s.Guild(t.GuildID)
+			if err != nil {
+				return err
+			}
+			s.Lock()
```
FedorLap2006, line 1031: I would personally add a line here, to separate the fetching section, but that's just a style change.
```suggestion
			}
			
			s.Lock()
```

FedorLap2006 (approved):
LGTM 🚀

state.go
```diff
@@ -1023,7 +1023,14 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
...
+			defer s.Unlock()
```
FedorLap2006, line 1032: Nvm, lock isn't needed here.
```suggestion

```

FedorLap2006 (changes requested):

state.go
```diff
@@ -1023,7 +1023,14 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
...
+			guild.Emojis = t.Emojis
```
FedorLap2006, line 1033: Additionally, after looking at more functions in the `state`, I think it would have made more sense to actually separate it out (that is why I deleted the original comment), like in `ThreadMemberUpdate`. But since change would not be significant, hope you don't mind if I make it myself.

FedorLap2006 (commented):
After thinking a bit more about this, it would have actually made sense to separate out the `GuildEmojisUpdate`. However, I'll probably do it in another PR, since it's been already a long time. Don't want to delay the PR even more.

PR #1263, opened 2022-10-23
Integrate handling of interactions via outgoing webhooks

topi314:
This is an alternative solution to https://github.com/bwmarrin/discordgo/pull/1181

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -12,6 +12,7 @@
 package discordgo
 
 import (
+	"crypto/ed25519"
 	"encoding/json"
 	"fmt"
 	"math"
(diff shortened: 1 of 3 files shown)


structs.go
```diff
@@ -115,6 +116,9 @@ type Session struct {
 	handlers     map[string][]*eventHandlerInstance
 	onceHandlers map[string][]*eventHandlerInstance
 
+	httpInteractions   map[string]func(response *InteractionResponse) error
+	httpInteractionsMu sync.Mutex
```
FedorLap2006, line 120:
```suggestion
	HttpInteractions   map[string]func(response *InteractionResponse) error
	HttpInteractionsMu sync.Mutex
```

TopiSenpai: I don't think this should be exposed as it's an interal

PR #1268, opened 2022-10-27
Add MarshalJSON to message and save raw response data

Suhaibinator:


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/bwmarrin/discordgo
```
bwmarrin, line 1: We can't make this change for sure :)  This should not be part of your PR.

bwmarrin:
@Suhaibinator 
Can you elaborate on what the goal of this change is, the use case for the change, why you want this?  Or some details to help us understand :)

go.mod
Suhaibinator: Oh yeah, sorry about that. I just needed to commit this so that I could use the code when I import the module in my go.mod for my other repo that this code was using. We can remove that part of the PR definitely.

Suhaibinator:
> 

So if I want test a bot with historical messages, or train an AI model on historical data, I want to be able to feed message data into the model exactly as they would appear coming from discord. The raw representation of the message data is the JSON marshal of the message object. Calling json.Marshal() on the message object does not give the same output as the json received from the Discord API which is why I wanted a way to store the raw response.

go.mod
FedorLap2006: A small tip, you could use `replace` directive in `go.mod` to get the same result, but without changing the target module 😄

FedorLap2006:
> I want to be able to feed message data into the model exactly as they would appear coming from discord

You can hook up a raw event handler (`func(s *discordgo.Session, e *discordgo.Event)`) from which you would be able to access `RawData` of the event:

```go
s.AddHandler(func(s *discordgo.Session, e *discordgo.Event) {
    if e.Type != "MESSAGE_CREATE" {
        return
   }
   fmt.Println(e.RawData)
})
```

> So if I want test a bot with historical messages, or train an AI model on historical data
I would recommend additionally asking about this in DDevs to not violate privacy policy by any chance.

> Calling json.Marshal() on the message object does not give the same output as the json received from the Discord API which is why I wanted a way to store the raw response.
That's by design, this data structure was meant only to be received and read. It will differ from the data sent by Discord due to it being partial sometimes.

go.mod
Suhaibinator: Yes, I am sorry for not thinking about this. 

> First of all, thanks for making the PR!
> 
> Would like to share some of my own thoughts on the use case you have described, if you don't mind:
> 
> > I want to be able to feed message data into the model exactly as they would appear coming from discord
> 
> You can hook up a raw event handler — `func(s *discordgo.Session, e *discordgo.Event)`, from which you would be able to access `RawData` of the event:
> 
> ```go
> s.AddHandler(func(s *discordgo.Session, e *discordgo.Event) {
>    if e.Type != "MESSAGE_CREATE" {
>         return
>    }
>    fmt.Println(e.RawData)
> })
> ```
> 
> > Calling json.Marshal() on the message object does not give the same output as the json received from the Discord API which is why I wanted a way to store the raw response.
> 
> That's by design, this data structure was meant only to be received and read. It will differ from the data sent by Discord due to it being partial sometimes.

Thanks for the feedback. OK so the solution you are proposing would only work for websocket connections, correct? There is not a way to integrate this with the REST API, for example to marshal the []*discordgo.Message received from session.ChannelMessages(channelID string, limit int, beforeID, afterID, aroundID string)?

FedorLap2006:
@Suhaibinator 
> There is not a way to integrate this with the REST API, for example to marshal the []*discordgo.Message received from session.ChannelMessages(channelID string, limit int, beforeID, afterID, aroundID string)?

Yes. However you can inspect how `ChannelMessages` uses `RequestWithBucketID`. And make the request yourself.

Suhaibinator:
> ```go
> Event
> ```

I see. Yeah, I think that works. We can close this then.

FedorLap2006:
Was glad to help.

Issue #1054, opened 2021-12-14
Support for the X-Audit-Log-Reason header


Cyb3r-Jak3:
Continued from #386

According to the [docs](https://discord.com/developers/docs/resources/audit-log#audit-logs) for every admin action performed you can give a reason by adding the `X-Audit-Log-Reason` header to the request and the reason will show up in the server's audit log.

There are functions like [GuildBanCreate](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L690) have a separate one like [GuildBanCreateWithReason](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L712).  However, the way that `GuildBanCreateWithReason` adds the reason with the query string is now deprecated.

Endpoints that support `X-Audit-Log-Reason` (Code links add to best of my ability):

* [Modify Guild](https://discord.com/developers/docs/resources/guild#modify-guild) | [Code](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L611)
* [Create Guild Channel](https://discord.com/developers/docs/resources/guild#create-guild-channel) | [Code](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L977)
* [Modify Guild Channel Positions](https://discord.com/developers/docs/resources/guild#modify-guild-channel-positions) | [Code](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L987)
* [Modify Guild Member](https://discord.com/developers/docs/resources/guild#modify-guild-member) | [Code](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L839)
* [Modify Current Member](https://discord.com/developers/docs/resources/guild#modify-current-member) | Not sure
* [Add Guild Member Role](https://discord.com/developers/docs/resources/guild#add-guild-member-role) | [Code](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L913)
* [Remove Guild Member Role](https://discord.com/developers/docs/resources/guild#remove-guild-member-role) | [Code](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L924)
* [Remove Guild Member](https://discord.com/developers/docs/resources/guild#remove-guild-member) | [Code](https://github.com/bwmarrin/discordgo/blob/v0.23.2/restapi.go#L815)
[...]

PR #1299, opened 2022-12-15
Options

FedorLap2006:
Since the beginning, DiscordGo did not have ability for adding per-call customization:

- What if you want to disable automatic rate limit handling when you call `ChannelMessageDelete`, instead handling the error by yourself?
- Or what if you want to set additional headers on `GuildBanCreate`, like `X-Audit-Log-Reason`, to comment your action in audit log?

We’ve been thinking on how to combat this problem, and came up with the following solutions:

- Builders pattern
    
    ```go
    s.WithAuditLogReason("my reason").GuildBanCreate(...)
    s.WithRetryOnRateLimit(false).ChannelMessageDelete(...)
    ```
    
- Functional options
    
    ```go
    s.GuildBanCreate(..., discordgo.WithAuditLogReason("reason"))
    s.ChannelMessageDelete(..., discordgo.WithRetryOnRateLimit(false))
    ```
    

This PR implements the latter. It provides a few pre-built options. However, you can always supply your custom option.

Options have access to a `RequestConfig` struct, which is internally used by `RequestX` functions. This config allows you to modify a few global settings, like `Client`, `ShouldRetryOnRateLimit` and `MaxRestRetries`, but it also allows you to modify the request directly.

A typical functional option would look like this:

```go
func WithQuery(key string, value string) discordgo.RequestOption {
	return func(cfg *discordgo.RequestConfig) {
		q := cfg.Request.Query()
		q.Add(key, value)
		cfg.Request.RawQuery = q.Encode()
	}
}
```

These are built-in options discordgo will provide:

- [x]  `WithClient` - Control `http.Client` used for the request (defaults to `s.Client`)
- [x]  `WithHeader` - Change a certain header in the request
    - [x]  `WithAuditLogReason` - Set `X-Audit-Log-Reason` header
    - [x]  `WithLocale` - Set `X-Discord-Locale` header
- [x]  `WithRetryOnRateLimit` - Control `Session.ShouldRetryOnRateLimit` field for the request
- [x]  `WithMaxRestRetries` - Control `Session.MaxRestRetries` field for the request
- [x]  `WithContext` - Provide a `context.Context` to the `http.Request`

This change will be affecting all of the REST endpoints. However it should be non-breaking.

If you have used any function signatures though, you would need to change them accordingly:

```diff
- func (s *Session) XXX(...) (...) {}
+ func (s *Session) XXX(..., options ...RequestOption) (...) {}
```

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -321,9 +321,9 @@ func unmarshal(data []byte, v interface{}) error {
 
 // User returns the user details of the given userID
 // userID    : A user ID or "@me" which is a shortcut of current user ID
-func (s *Session) User(userID string) (st *User, err error) {
+func (s *Session) User(userID string, options ...RequestOption) (st *User, err error) {
 
-	body, err := s.RequestWithBucketID("GET", EndpointUser(userID), nil, EndpointUsers)
+	body, err := s.RequestWithBucketID("GET", EndpointUser(userID), nil, EndpointUsers, options...)
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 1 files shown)


FedorLap2006:
This PR also closes #1054 and #752

PR #1301, opened 2022-12-19
Add size parameters to IconURL and BannerURL

sentinelb51:
Adds guild icon size parameter, guild banner size parameter, missing ClientStatus struct


FedorLap2006 (changes requested):
In terms of `ClientStatus`, we'll probably favor #1303, since it also updates the `State`. 

P.S. And it not having the comment is the reason why our linter is angry 😄

FedorLap2006 (approved):
I've done a few of my own changes. In particular refactored out the changes into a `iconURL` function. And added the documentation.

Thanks for your contribution nonetheless!

PR #1305, opened 2022-12-24
Automod trigger metadata update

ikafly144:
Add AllowList and MentionTotalLimit to AutoMod as shown in docs below.
https://discord.com/developers/docs/resources/auto-moderation#auto-moderation-rule-object-trigger-metadata

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -1449,7 +1449,7 @@ type AutoModerationTriggerMetadata struct {
 
 	// Substrings which should not trigger the rule.
 	// NOTE: should be only used with keyword or keyword preset trigger type.
-	AllowList []string `json:"allow_list,omitempty"`
+	AllowList []*string `json:"allow_list,omitempty"`
 
 	// Total number of unique role and user mentions allowed per message.
 	// NOTE: should be only used with mention spam trigger type.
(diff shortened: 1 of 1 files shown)


structs.go
```diff
@@ -1446,6 +1446,14 @@ type AutoModerationTriggerMetadata struct {
 	// Internally pre-defined wordsets which will be searched for in content.
 	// NOTE: should be only used with keyword preset trigger type.
 	Presets []AutoModerationKeywordPreset `json:"presets,omitempty"`
+
+	// Substrings which should not trigger the rule.
+	// NOTE: should be only used with keyword or keyword preset trigger type.
+	AllowList []string `json:"allow_list,omitempty"`
```
FedorLap2006, line 1452: Since allow list can be an empty list, it should be a pointer.

```suggestion
	AllowList *[]string `json:"allow_list,omitempty"`
```

FedorLap2006 (changes requested):

structs.go
ikafly144: OK, fixed

FedorLap2006 (approved):

Issue #1318, opened 2023-01-29
UserUpdate previous User data


ryanblenis:
UserUpdate does not currently send the previous User object.

I'm not sure if users should be added to State to accomplish this, if an additional dereferenced copy of the user should be included in GuildMemberUpdate BeforeUpdate copy creation (as currently the pointer to the User is copied, and thus identical in the GuildMemberUpdate BeforeUpdate member.User property during comparison).

To make matters more interesting, Discord tends to send UserUpdate, PresenceUpdate, and GuildMemberUpdate when a user has changed their username. I would have implemented this myself and added a pull request, but don't want to stomp on how you'd prefer these cases to be handled and which call (or all of them with a comparison of Presence / GuildMemberUpdate to detect changes and determine only the User was updated and send a single UserUpdate while suppressing Presence / GuildMemberUpdate calls).

Given that info, if there's a preferred implementation path you can offer, I'd be happy to try and implement it if you're overloaded.

Use case: The ability to keep a log of users username changes; typically for the purpose of detecting malicious users changing their name to mimic those higher rank(s) in guilds (mostly seen in paid discords) in an attempt to DM spam and defraud guild members by using the name and avatar of someone "trusted" to try to scam unsuspecting server members.

PR #1320, opened 2023-01-30
Properly reference previous user when user data is changed and GuildM…

ryanblenis:
…emberUpdate is fired with TrackMembers enabled

Minimal implementation for obtaining before/after snapshot of user changes. See additional information and items for consideration in #1318 if a different implementation path is preferred. 

Currently the User changed data is lost as the BeforeUpdate Member copies the pointer to the associated user, leaving both the current Member and BeforeUpdate Member to reference the now updated user properties.

Thank you.

### DIFF
--- a/state.go
+++ b/state.go
@@ -996,6 +996,10 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 			old, err = s.Member(t.GuildID, t.User.ID)
 			if err == nil {
 				oldCopy := *old
+				if oldCopy.User != nil {
+					oldUser := *oldCopy.User
+					oldCopy.User = &oldUser
+				}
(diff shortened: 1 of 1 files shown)


FedorLap2006 (approved):

FedorLap2006 (commented):
Sorry for a late response. I also probably was too quick to approve this. But anyway.

After investigating this myself, I haven't exactly found where the copy is happening into the `User`.
https://github.com/bwmarrin/discordgo/blob/0c2b5faa1afd56816facba4795b6a5ccc837cf48/state.go#L328
In `memberAdd` (linked above) we only replace the member object. And since we're in `OnInterface` - we should receive a new `User`. Thus only the pointer should be replaced, and in the copy it would stay as it was.

PR #1335, opened 2023-02-26
fix: ThreadMember flags not serialized

armsnyder:
The `Flags` field was being incorrectly de/serialized as `"Flags"`.

ThreadMember spec: https://discord.com/developers/docs/resources/channel#thread-member-object

I caught this with the `musttag` linter, so I went ahead and added it to the linter config.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -527,7 +527,7 @@ type ThreadMember struct {
 	// The time the current user last joined the thread
 	JoinTimestamp time.Time `json:"join_timestamp"`
 	// Any user-thread settings, currently only used for notifications
-	Flags int
+	Flags int `json:"flags"`
 }
 
 // ThreadsList represents a list of threads alongisde with thread member objects for the current user.


FedorLap2006 (commented):
LGTM, however not entirely sure about the usage of `musttag` linter. It does not (at least currently) properly fullfil our needs: as an example, we tend to have a lot of nested structures, but after some local testing, it doesn't seem to catch any changes in them.

In any case, it's going to be a good idea to separate the linter change into a different PR, to make things easier.

armsnyder:
@FedorLap2006 It sounds like you are concerned about false negatives, which is the case with most linters. They safeguard against some but not all bugs. In the case the `musttag` linter did catch the `Flags` bug.

I removed the commit from this PR and opened #1345

FedorLap2006:
Great! Since I'm still writing up the release notes for the v0.27.1, we can probably merge this as well.

PR #1337, opened 2023-02-27
fix: WithContext had no effect

### LINKED ISSUE #1336: WithContext has no effect
The `WithContext` request option is broken because it does not actually update the request. I'm working on a fix.

armsnyder:
Fixes #1336

### DIFF
--- a/restapi_test.go
+++ b/restapi_test.go
@@ -244,7 +244,7 @@ func TestWithContext(t *testing.T) {
 	ctx := context.WithValue(context.Background(), key{}, "foo")
 
 	// Set up a test client.
-	sess, err := New("")
+	session, err := New("")
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -252,7 +252,7 @@ func TestWithContext(t *testing.T) {
 	testErr := errors.New("test")
 
 	// Intercept the request to assert the context.
-	sess.Client.Transport = roundTripperFunc(func(r *http.Request) (*http.Response, error) {
+	session.Client.Transport = roundTripperFunc(func(r *http.Request) (*http.Response, error) {
 		val, _ := r.Context().Value(key{}).(string)
 		if val != "foo" {
 			t.Errorf("missing value in context (got %q, wanted %q)", val, "foo")
@@ -261,7 +261,7 @@ func TestWithContext(t *testing.T) {
 	})
 
 	// Run any client method using WithContext.
-	_, err = sess.User("", WithContext(ctx))
+	_, err = session.User("", WithContext(ctx))
 
 	// Verify that the assertion code was actually run.
 	if !errors.Is(err, testErr) {
--- a/restapi_test.go
+++ b/restapi_test.go
@@ -1,7 +1,9 @@
 package discordgo
 
 import (
+	"context"
 	"errors"
+	"net/http"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


FedorLap2006 (commented):
Hi, thanks for outlining the problem. Haven't noticed myself.
One thing to note though, I think the solution with replacing `Do(req)` to `Do(cfg.Request)` would look more consistent.

FedorLap2006:
> @FedorLap2006 I still prefer my first version, but I changed it to reassign `req` in the caller function instead. I didn't want to simply replace `cfg.Client.Do(req)` with `cfg.Client.Do(cfg.Request)` since there are more places that `req` is used.

Ideally all occurences should be replaced with the `cfg.Request`. Besides it's only a couple (`req.Header` & `newRestError`)

armsnyder:
@FedorLap2006 Hi, I followed your suggestion and updated the code. ☺️ 

My two cents are the first version with `*cfg.Request = *cfg.Request.WithContext(ctx)` is the cleaner and more consistent option.

More consistent: All the other `RequestOption`'s are mutating the original request. It seems that `WithContext` should also mutate it.

Cleaner: It's better to have only one request in-scope in the `RequestWithLockedBucket` function, instead of keeping a reference to the old `req` before applying options. It is a long function (111 lines!), so it may not be obvious for future code changes whether `req` or `cfg.Request` should be used.

You have final say, of course. 😅

FedorLap2006 (approved):
LGTM!

FedorLap2006:
Regarding the original solution. After some consideration I have to admit the original solution could be the right way to go. Stylistically, it doesn't match up with other options, and can be a little bit confusing when reading `RequestWithLocketBucketID`, with all `cfg.XXX` around it. But logically, since other options mutate the original value (ref: `WithHeader`), it does make sense.

As for the reassignment of `req` inside `RequestWithLockedBucketID`, it doesn't match up stylistically and logicially with other options at all, that's why I've suggested to replace it by `cfg.Request`.

FedorLap2006:
Sorry for taking matter into my own hands. After thinking for a while, the second solution (with reassigning `req`), does not affect readability that much. And has the least amount of changes.

P.S. Another way this could have been implemented was to use a `Context` field in `RequestConfig` and then reassign the `req` to `req.WithContext(cfg.Request)` (which is what I originally wanted to do), but `req = cfg.Request` would just be a simpler version of this.

restapi_test.go
```diff
@@ -235,3 +237,41 @@ func Test_unmarshal(t *testing.T) {
...
+		return nil, testErr
+	})
+
+	// Run any client method using WithContext.
+	_, err = sess.User("", WithContext(ctx))
+
+	// Verify that the assertion code was actually run.
+	if !errors.Is(err, testErr) {
+		t.Errorf("unexpected error %v returned from client", err)
+	}
```
FedorLap2006, line 269:
```suggestion
	dg, err := New("")
	if err != nil {
		t.Fatal(err)
	}

	testErr := errors.New("test")

	// Intercept the request to assert the context.
	dg.Client.Transport = roundTripperFunc(func(r *http.Request) (*http.Response, error) {
		val, _ := r.Context().Value(key{}).(string)
		if val != "foo" {
			t.Errorf("missing value in context (got %q, wanted %q)", val, "foo")
		}
		return nil, testErr
	})

	// Run any client method using WithContext.
	_, err = dg.User("", WithContext(ctx))

	// Verify that the assertion code was actually run.
	if !errors.Is(err, testErr) {
		t.Errorf("unexpected error %v returned from client", err)
	}
```

restapi_test.go
```diff
@@ -235,3 +237,41 @@ func Test_unmarshal(t *testing.T) {
 		t.Errorf("Unexpected error type: %T", err)
 	}
 }
+
+func TestWithContext(t *testing.T) {
+	// Set up a test context.
+	type key struct{}
+	ctx := context.WithValue(context.Background(), key{}, "foo")
```
FedorLap2006, line 244: I think it's going to be better to replace it with `value`, to be more consistent with the naming.

```suggestion
	ctx := context.WithValue(context.Background(), key{}, "value")
```

restapi_test.go
```diff
@@ -235,3 +237,41 @@ func Test_unmarshal(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	testErr := errors.New("test")
+
+	// Intercept the request to assert the context.
+	session.Client.Transport = roundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		val, _ := r.Context().Value(key{}).(string)
+		if val != "foo" {
```
FedorLap2006, line 257: In relation to https://github.com/bwmarrin/discordgo/pull/1337/files#r1131680827
```suggestion
		if val != "value" {
```

FedorLap2006:
Sorry for doing the commits right on the branch. Since the release is planned for today, I've decided to not prolong the things even more and apply all the edits myself.

Thank you for the contribution!

armsnyder:
@FedorLap2006 No worries! Ultimately what matters most is the bug is fixed. 😄 Thanks for the review.

PR #1343, opened 2023-03-07
74 byte handshake

FedorLap2006:
Update format of the UDP handshake when performing IP discovery.

```diff
+ direction (2 bytes, 1 - send, 2 - recv)
+ length (2 bytes, 70)
  ssrc (4 bytes)
- ip (16 bytes)
+ ip (64 bytes, null terminated)
  port (2 bytes)
```

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -627,7 +627,7 @@ func (v *VoiceConnection) udpOpen() (err error) {
 		return fmt.Errorf("received udp packet too small")
 	}
 
-	// Loop over position 8 through 68 to grab the IP address.
+	// Loop over position 8 through 71 to grab the IP address.
 	var ip string
 	for i := 8; i < len(rb)-2; i++ {
 		if rb[i] == 0 {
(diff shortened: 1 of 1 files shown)


voice.go
```diff
@@ -599,44 +599,45 @@ func (v *VoiceConnection) udpOpen() (err error) {
...
 
-	if rlen < 70 {
+	if rlen < 74 {
 		v.log(LogWarning, "received udp packet too small")
 		return fmt.Errorf("received udp packet too small")
 	}
 
-	// Loop over position 4 through 20 to grab the IP address
-	// Should never be beyond position 20.
+	// Loop over position 8 through 68 to grab the IP address.
```
CamiKaseM7, line 630: It's position 72, not 68

voice.go
```diff
@@ -599,44 +599,45 @@ func (v *VoiceConnection) udpOpen() (err error) {
...
 	var ip string
-	for i := 4; i < 20; i++ {
+	for i := 8; i < len(rb)-2; i++ {
 		if rb[i] == 0 {
 			break
 		}
 		ip += string(rb[i])
 	}
 
 	// Grab port from position 68 and 69
```
CamiKaseM7, line 639: And here we should also update the comment to the newer positions, 72 to 74

CamiKaseM7:
I think that's all we need. The deadline to implement this is march 15th, so the sooner we deploy this the better

PR #1347, opened 2023-03-12
feat(wsapi): Gracefully close the websocket connection

armsnyder:
This PR removes a `time.Sleep(1 * time.Second)` from `Session.Close()`.

Instead, I implemented the websocket close handshake described in this comment: https://github.com/gorilla/websocket/issues/448#issuecomment-469903859

The reason I would like to see this change is that the `time.Sleep` was causing my unit tests to run slowly.

I tested this change by running this program a few times:

```go
package main

import (
	"github.com/samber/lo"
)

func main() {
	session := lo.Must(discordgo.New(Getenv("BOT_TOKEN")))
	session.LogLevel = discordgo.LogDebug
	lo.Must0(session.Open())
	lo.Must0(session.Close())
}
```

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -921,24 +921,11 @@ func (s *Session) CloseWithCode(closeCode int) (err error) {
 	// this should force stop any reconnecting voice channels too
 
 	if s.wsConn != nil {
-
-		s.log(LogInformational, "sending close frame")
-		// To cleanly close a connection, a client should send a close
-		// frame and wait for the server to close the connection.
-		s.wsMutex.Lock()
-		err := s.wsConn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(closeCode, ""))
-		s.wsMutex.Unlock()
-		if err != nil {
-			s.log(LogInformational, "error closing websocket, %s", err)
-		}
-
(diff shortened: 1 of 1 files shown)


wsapi.go
```diff
@@ -951,3 +938,36 @@ func (s *Session) CloseWithCode(closeCode int) (err error) {
 
 	return
 }
+
+func (s *Session) doCloseHandshake(closeCode int) error {
```
FedorLap2006, line 942:
```suggestion
func (s *Session) gracefulClose(closeCode int) error {
```

FedorLap2006 (changes requested):

wsapi.go
FedorLap2006, line 942:
```suggestion
func (s *Session) gracefulClose(closeCode int) error {
```

FedorLap2006 (changes requested):

wsapi.go
armsnyder: This function does not actually close the connection though. The caller closes it. All this function does is the handshake.

FedorLap2006: My bad. Not really sure how I missed that 😅

PR #1350, opened 2023-03-13
Add toggle for voice reconnect on session reconnect

### LINKED ISSUE #1349: Session reconnect forces voice reconnect
I currently have an issue that is related to this part of code.
```go			
                               // I'm not sure if this is actually needed.
				// if the gw reconnect works properly, voice should stay alive
				// However, there seems to be cases where something "weird"
				// happens.  So we're doing this for now just to improve
				// stability in those edge cases.
				s.RLock()
				defer s.RUnlock()
				for _, v := range s.VoiceConnections {

					s.log(LogInformational, "reconnecting voice connection to guild %s", v.GuildID)
					go v.reconnect()

					// This is here just to prevent violently spamming the
					// voice reconnects
					time.Sleep(1 * time.Second)

				}
				return
```
 On session reconnect for example, which happens randomly, it forces my bot to exit and enter the voice channel. Currently I am working on a soundboard bot that basically stays in the voice channel of a server while members are there. Sometimes it reconnects as often as 3 times in a time span of 10-20 minutes which is quite annoying. I propose to wrap the code around a toggle. I created a fork to see if the removal of this code fixes my issue and I will come back with a PR if everything checks out.

loghinalexandru:
This PR adds a flag in order to prevent voice reconnect on session reconnect. The default value is set to `true` so it should not affect the initial behaviour.

Resolves #1349

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -42,6 +42,9 @@ type Session struct {
 	// Should the session reconnect the websocket on errors.
 	ShouldReconnectOnError bool
 
+	// Should the session reconnect also trigger voice connections reconnect.
+	ShouldReconnectVoiceConnOnError bool
+
 	// Should the session retry requests when rate limited.
 	ShouldRetryOnRateLimit bool
 
--- a/structs.go
+++ b/structs.go
@@ -42,8 +42,8 @@ type Session struct {
 	// Should the session reconnect the websocket on errors.
 	ShouldReconnectOnError bool
 
-	// Should the session reconnect also trigger voice connections reconnect.
-	ShouldReconnectVoiceConnOnError bool
+	// Should voice connections reconnect on session websocket errors.
+	ShouldReconnectVoiceOnSessionError bool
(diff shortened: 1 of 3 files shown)


FedorLap2006 (changes requested):

FedorLap2006:
Sorry for a long wait!

structs.go
```diff
@@ -42,6 +42,9 @@ type Session struct {
 	// Should the session reconnect the websocket on errors.
 	ShouldReconnectOnError bool
 
+	// Should voice connections reconnect on session websocket errors.
```
FedorLap2006, line 45:
```suggestion
	// Should voice connections be reconnected on a session reconnect.
```

FedorLap2006 (commented):
Actually after a bit of thinking, I think we're going to move forward with this doc comment.
As we don't reconnect only on errors.

PR #1372, opened 2023-04-28
Fix bug with error not being returned in WebhookEditWithToken

quackduck:
Earlier, if there was an error it would not be returned since the `:=` operator was used instead of a `=` operator.


FedorLap2006 (approved):

FedorLap2006:
Sorry for a long wait. And thanks for noticing the fix!

quackduck:
okay I just tested whether this actually is a problem:

https://go.dev/play/p/FvZC2TYqcCJ

It's not, so this PR hasn't really fixed anything lol oof

PR #1373, opened 2023-05-01
Voice discovery packet

Joe3160:
suport proxy of websoket listen

[0001-Support-to-Proxy-of-websocket-listennig.patch](https://github.com/bwmarrin/discordgo/files/11363891/0001-Support-to-Proxy-of-websocket-listennig.patch)

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -627,7 +627,7 @@ func (v *VoiceConnection) udpOpen() (err error) {
 		return fmt.Errorf("received udp packet too small")
 	}
 
-	// Loop over position 8 through 68 to grab the IP address.
+	// Loop over position 8 through 71 to grab the IP address.
 	var ip string
 	for i := 8; i < len(rb)-2; i++ {
 		if rb[i] == 0 {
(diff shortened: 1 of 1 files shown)


FedorLap2006:
These commits should be already in master. Due to this, I'm going to close this PR as invalid.

As for the change you have submitted in the `.patch` file, there are a couple of problems with it:
1. `cfg.Client` is never `nil`, and is inherited from `s` (`*Session`). By overriding it, you are discarding all of the global/option (those set by `WithXXX` funcs) settings (including the `Proxy`, see no. 2)
2. `cfg.Client.Transport.Proxy` can be already non-`nil` and differ from the `s.Dialer.Proxy`
3. Changing the proxy and allocating a new `http.Client` (can be solved by just assigning `cfg.Client.Transport.Proxy` to the desired value)  on each request to the same values just wastes time and memory.  
4. It's a probably a good idea to leave up to the user to decide the proxies. Because:
     - Proxy manipulation, without an explicit mention of it in docs, might lead to undesired problems
     - It's a one-liner when setting up the session
     - `nil` is a default and perfectly valid value for the `Proxy`

PR #1371, opened 2023-04-25
feat(auditlog): add auto moderation actions and options

Earlopain:
https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object-audit-log-events
https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object-optional-audit-entry-info

PR #1380, opened 2023-05-14
feat(auditlog): add application_id to AuditLogOptions

Earlopain:
Put into a seperate PR as per #1371

I added it above the moderator fields because that is the order they were added to the api.


FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1381, opened 2023-05-14
Update Github CI versions

Earlopain:
* Remove unused travis config
* Bump the actions in the workflow file
* Add Go 1.19 and 1.20 to the test matrix. I added quotes around the versions because 1.20 is being interpreted as 1.2 otherwise
* Bump the linter and apply needed formatting changes. I can't tell how some of these changes would be rendered in the docs, I just went with what the formatted wanted.


.github/workflows/ci.yml
```diff
@@ -54,7 +54,6 @@ jobs:
       - name: GolangCI-Lint
         if: github.event.name != 'pull_request' # See https://github.com/golangci/golangci-lint-action/issues/362
         run: |
-          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.45.2
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.52.2
```
FedorLap2006, line 57: Looking through releases of [golangci/golangci-lint-action[(https://github.com/golangci/golangci-lint-action), it seems their latest release (3.6) should fix this issue. So, this hacky workaround might be not needed anymore.
However, this needs to be tested.

Earlopain: From what I understand the workaround can be removed, I pushed a commit doing so. I have a test PR on my fork https://github.com/Earlopain/discordgo/pull/1 where you can look at some output.

I changed the trigger for push to only run on master because ci would run twice on PRs otherwise. They fall under both specified events. Previously the if checks on the steps took care of that.

Earlopain:
Redid on top of current master

Earlopain:
Rebased again, added Go 1.22

FedorLap2006 (changes requested):
Please remove formatting changes.

Earlopain:
I'm fairly sure the formatting changes are required by bumping `golangci-lint`. If you don't want that I can drop it, I will make a new PR in that case.

Earlopain:
This is probably out of date again

PR #1388, opened 2023-06-09
Add GlobalName field to User

merlinfuchs:
This field is being introduced with the username change and is displayed pretty much everywhere instead of the actual username.

https://discord.com/developers/docs/resources/user#user-object


FedorLap2006:
Thank you for making this PR, however, I will close it in favor of #1387, which also adds `GlobalName` to `User`.

PR #1390, opened 2023-06-09
Add missing webhook thread endpoints and allow overriding webhook message attachments

merlinfuchs:
This PR adds the missing `attachments` field to the `WebhookEdit` struct so you can remove and override attachments on webhook messages. See https://discord.com/developers/docs/resources/webhook#edit-webhook-message

It also adds the missing thread specific equivalents to the `WebhookMessage`, `WebhookMessageEdit`, and `WebhookMessageDelete` endpoints.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -2401,6 +2401,28 @@ func (s *Session) WebhookMessage(webhookID, token, messageID string, options ...
 	return
 }
 
+// WebhookMessage gets a webhook message.
+// webhookID : The ID of a webhook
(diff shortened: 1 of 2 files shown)


FedorLap2006:
Thank you for implementing these missing features, however, I think it's a good idea to separate the endpoint and attachment changes into different PRs, since they have completely different scopes.

merlinfuchs:
#1392 adds the attachments parameter so I will remove it from this PR

PR #1392, opened 2023-06-19
Add Attachment options to interactions

KokeCacao:
Add Attachment options to interactions so that editing uploaded files is possible.
Behavior documented here: https://discord.com/developers/docs/resources/webhook#edit-webhook-message


FedorLap2006:
Thanks for your contribution!

PR #1395, opened 2023-06-23
Remove remove duplicated slash in template endpoint

merlinfuchs:
`EndpointGuilds` already has a slash at the end so we end up with `/guilds//templates` which is not ideal. The Discord API ignores the double slash but if you are using some kind of proxy it doesn't work.


FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1401, opened 2023-06-29
feat(guild): implement onboarding

Earlopain:
Adds support for inspecting the onboarding configuration of a guild.

https://discord.com/developers/docs/resources/guild#guild-onboarding-object

I want to know which roles are assignable by users through onboarding, which this is needed for.

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -3250,8 +3250,8 @@ func (s *Session) GuildScheduledEventUsers(guildID, eventID string, limit int, w
 	return
 }
 
-// GuildOnboarding returns the onboarding flow for a guild.
-// guildID   : The ID of a Guild
+// GuildOnboarding returns onboarding configuration of a guild.
+// guildID   : The ID of the guild
 func (s *Session) GuildOnboarding(guildID string, options ...RequestOption) (onboarding *GuildOnboarding, err error) {
 	endpoint := EndpointGuildOnboarding(guildID)
 
@@ -3265,9 +3265,9 @@ func (s *Session) GuildOnboarding(guildID string, options ...RequestOption) (onb
 	return
 }
 
-// GuildOnboardingEdit edits Onboarding for a Guild.
-// guildID   : The ID of a Guild.
-// o 		     : A GuildOnboarding struct.
+// GuildOnboardingEdit edits onboarding configuration of a guild.
+// guildID   : The ID of the guild
+// o 		     : New GuildOnboarding data
 func (s *Session) GuildOnboardingEdit(guildID string, o *GuildOnboarding, options ...RequestOption) (onboarding *GuildOnboarding, err error) {
 	endpoint := EndpointGuildOnboarding(guildID)
 
--- a/restapi.go
+++ b/restapi.go
@@ -3265,6 +3265,22 @@ func (s *Session) GuildOnboarding(guildID string, options ...RequestOption) (onb
 	return
 }
 
+// GuildOnboardingEdit edits Onboarding for a Guild.
+// guildID   : The ID of a Guild.
+// o 		     : A GuildOnboardingParams struct.
+func (s *Session) GuildOnboardingEdit(guildID string, o *GuildOnboardingParams, options ...RequestOption) (onboarding *GuildOnboarding, err error) {
+	endpoint := EndpointGuildOnboarding(guildID)
+
+	var body []byte
+	body, err = s.RequestWithBucketID("PUT", endpoint, o, endpoint, options...)
+	if err != nil {
+		return
+	}
+
+	err = unmarshal(body, &onboarding)
+	return
+}
+
 // ----------------------------------------------------------------------
 // Functions specific to auto moderation
 // ----------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


FedorLap2006:
I could have done a rebase, but whatever 😅

restapi.go
```diff
@@ -3250,6 +3250,21 @@ func (s *Session) GuildScheduledEventUsers(guildID, eventID string, limit int, w
 	return
 }
 
+// GuildOnboarding returns the onboarding flow for a guild
+// guildID   : The ID of a Guild
+func (s *Session) GuildOnboarding(guildID string, options ...RequestOption) (onboarding GuildOnboarding, err error) {
```
FedorLap2006, line 3255: From REST functions we usually return pointers.

structs.go
```diff
@@ -1070,6 +1070,79 @@ type GuildScheduledEventUser struct {
...
+	// ID of the guild this onboarding is part of
+	GuildID string `json:"guild_id"`
+
+	// Prompts shown during onboarding and in customize community
+	Prompts []GuildOnboardingPrompt `json:"prompts"`
+
+	// Channel IDs that members get opted into automatically
+	DefaultChannelIDs []string `json:"default_channel_ids"`
+
+	// 	Whether onboarding is enabled in the guild
```
FedorLap2006, line 1085:
```suggestion
	// Whether onboarding is enabled in the guild
```

structs.go
```diff
@@ -1070,6 +1070,79 @@ type GuildScheduledEventUser struct {
...
+	RoleIDs []string `json:"role_ids"`
+
+	// Emoji of the option
+	Emoji Emoji `json:"emoji"`
+
+	// Title of the option
+	Title string `json:"title"`
+
+	// Description of the option
+	Description string `json:"description,omitempty"`
```
FedorLap2006, line 1143: I don't think there's a point in setting `omitempty` (it affects only `json.Marshal`), but `*string`, as docs specify type as nullable is a good idea.

P.S. On the other hand, if `description` cannot be set to `""`, I think it would be fair to get rid of the pointer. 
That requires a bit of testing though

restapi.go
```diff
@@ -3250,6 +3250,21 @@ func (s *Session) GuildScheduledEventUsers(guildID, eventID string, limit int, w
 	return
 }
 
+// GuildOnboarding returns the onboarding flow for a guild
```
FedorLap2006, line 3253: Not all our symbols have this (since it's a non-issue), but it's probably a good idea to put a period at the end.
```suggestion
// GuildOnboarding returns the onboarding flow for a guild.
```

Same goes for all other types.

FedorLap2006 (changes requested):
Overall, changes are good to go, however, a few tweaks are required.

Earlopain:
Thanks for the feedback, I appreciate your time on this. I'm currently on vacation and will get to it in a week or two.

structs.go
Earlopain: Yeah, that's my bad. Your suggestion is correct.

Earlopain:
I should now have addressed all your feedback.

structs.go
```diff
@@ -1070,6 +1070,94 @@ type GuildScheduledEventUser struct {
...
+	GuildOnboardingModeAdvanced = 1
+)
+
+// GuildOnboarding represents the onboarding flow for a guild.
+// https://discord.com/developers/docs/resources/guild#guild-onboarding-object
+type GuildOnboarding struct {
+	// ID of the guild this onboarding is part of.
+	GuildID string `json:"guild_id"`
+
+	// Prompts shown during onboarding and in customize community.
```
FedorLap2006, line 1091:
```suggestion
	// Prompts shown during onboarding and in the Channel & Roles tab.
```

FedorLap2006 (changes requested):

structs.go
```diff
@@ -1070,6 +1070,94 @@ type GuildScheduledEventUser struct {
...
+
+// Block containing known GuildOnboardingMode values
+const (
+	// GuildOnboardingModeDefault counts default channels towards constraints
+	GuildOnboardingModeDefault = 0
+	// GuildOnboardingModeAdvanced counts default channels and questions towards constraints
+	GuildOnboardingModeAdvanced = 1
+)
+
+// GuildOnboarding represents the onboarding flow for a guild.
```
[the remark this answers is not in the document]
Earlopain: I left this struct alone and added `GuildOnboardingParams`. Currently untested but is the idea fine? There are a few places that do it this way like `GuildParams` or`RoleParams` and I thought it makes sense that way.

FedorLap2006:
Sorry it took so long!

FedorLap2006:
> decided to make some of the changes myself

Well, a lot of changes. Want to merge this into the next release, that I'm going to make very soon. So decided to take it upon myself to finish the PR as quickly as possible.

FedorLap2006:
Thanks for your contribution!

Earlopain:
Much appreciated for pushing this over the finish line. My bad for not keeping up here but I pretty quickly found another way to better achieve what I wanted through role flags (https://github.com/bwmarrin/discordgo/pull/1410) so this wasn't as relevant to me anymore.

Very nice to hear about a potential new release soonish.

Issue #829, opened 2020-10-31
Voice websocket closed unexpectedly at random times


yms2772:
I'm making a bot which is playing songs using [github.com/jonas747/dca](github.com/jonas747/dca).
However it always leave the voice channel at random times (about after 1 hour) with below log and join the channel again.
My internet is stable and I even used Amazon AWS but it showed same problem.

```md
2020/10/31 15:14:36 [DG0] voice.go:376:wsListen() voice endpoint south-korea889.discord.media:443 websocket closed unexpectantly, websocket: close 1001 (going away)
2020/10/31 15:14:36 [DG0] voice.go:197:Close() error closing websocket, websocket: close sent
2020/10/31 15:14:37 [DG0] voice.go:206:Close() error closing websocket, write tcp 172.17.0.8:60728->162.159.128.235:443: i/o timeout
2020/10/31 15:14:49 [DG1] wsapi.go:648:ChannelVoiceJoin() error waiting for voice to connect, timeout waiting for voice
```

Mexican-Man:
I've been getting this and it's a real pain. No way to catch it or anything. For me at least, it doesn't always crash from it; only about half the time. Really hoping someone will give us an update. Even a basic `recover()` would be a godsend at this point

Mexican-Man:
I have an update on this one. Thanks to the @FlameInTheDark on [this issue here](https://github.com/jonas747/yagpdb/issues/284#issuecomment-475544178), it appears that using this:
```go
vc, err := discord.ChannelVoiceJoin(guildID, channelID, properties.Muted, properties.Deafened)
	if err != nil {
		if _, ok := discord.VoiceConnections[guildID]; ok {
			vc = discord.VoiceConnections[guildID]
		} else {
			return nil, err
		}
	}
```
—will fix your issue.

It almost looks like `(Session*) ChannelVoiceJoin()` just doesn't return properly sometimes. Not sure if this is a discordgo issue, or a Discord issue.

yms2772:
> I have an update on this one. Thanks to the @FlameInTheDark on [this issue here](https://github.com/jonas747/yagpdb/issues/284#issuecomment-475544178), it appears that using this:
> 
> ```go
> vc, err := discord.ChannelVoiceJoin(guildID, channelID, properties.Muted, properties.Deafened)
> 	if err != nil {
> 		if _, ok := discord.VoiceConnections[guildID]; ok {
> 			vc = discord.VoiceConnections[guildID]
> 		} else {
[...]

PR #1408, opened 2023-07-17
Fix session ID deadlockwhen opening a VoiceConnection

denverquane:
Closes https://github.com/bwmarrin/discordgo/issues/829

When checking the Voice Session ID in `VoiceConnection.open()`, the sessionID is populated asynchronously when websocket messages come in, see [wsapi.go](https://github.com/bwmarrin/discordgo/blob/master/wsapi.go#L736C19-L736C19). 

However, `open()` still holds the lock on the session, so the ID can never be populated, and we loop and accomplish nothing until releasing the lock upon exiting the function. 

This PR addresses this deadlock bug by momentarily releasing the lock anytime the process sleeps, allowing other processes to populate the sessionID appropriately - while not sacrificing concurrency safety by releasing the lock entirely.

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -76,7 +76,8 @@ type VoiceSpeakingUpdateHandler func(vc *VoiceConnection, vs *VoiceSpeakingUpdat
 // Speaking sends a speaking notification to Discord over the voice websocket.
 // This must be sent as true prior to sending audio and should be set to false
 // once finished sending audio.
-//  b  : Send true if speaking, false if not.
+//
+//	b  : Send true if speaking, false if not.
 func (v *VoiceConnection) Speaking(b bool) (err error) {
 
 	v.log(LogDebug, "called (%t)", b)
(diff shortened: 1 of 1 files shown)


voice.go
```diff
@@ -294,11 +294,15 @@ func (v *VoiceConnection) open() (err error) {
 		if v.sessionID != "" {
 			break
 		}
+
 		if i > 20 { // only loop for up to 1 second total
 			return fmt.Errorf("did not receive voice Session ID in time")
 		}
+		// release the lock so other processes can populate the sessionID if a ws message is received
```
FedorLap2006, line 301:
```suggestion
		// Release the lock, so sessionID can be populated upon receiving a VoiceStateUpdate event.
```

FedorLap2006:
Thanks for the contribution!

FedorLap2006:
I did the edits myself, to not hold it off for longer. Hope you understand and sorry for the wait.

PR #1334, opened 2023-02-23
Added support for the icon and unicode_emoji role struct fields

SamusAranX:
This partially addresses #1316 by adding support for the fields mentioned in this PR's title. 
I tested this on a server with role icons and emoji and encountered no issues.

Issue #1316, opened 2023-01-22
Guild Role struct


hff399:
Guild Role struct doesn't implement 3 discord API fields: icon, unicode_emoji, tags
https://discord.com/developers/docs/topics/permissions#role-object

hff399:
thx

PR #1410, opened 2023-07-24
feat(role): add flags

Earlopain:
Depends on #1334

https://discord.com/developers/docs/topics/permissions#role-object-role-flags

A better solution to what I was doing in my onboarding pr, wasn't documented at the time though.
I also thought about adding the `tags` field as per #1316 but that has weird nullable types I didn't want to bother with. The issue should also be reopened as the PR doesn't address it fully and isn't merged.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -1239,11 +1239,34 @@ type Role struct {
 	// This is a combination of bit masks; the presence of a certain permission can
 	// be checked by performing a bitwise AND between this int and the permission.
 	Permissions int64 `json:"permissions,string"`
+
+	// The hash of the role icon. Use Role.IconURL to retrieve the icon's URL.
+	Icon string `json:"icon"`
+
+	// The emoji assigned to this role.
+	UnicodeEmoji string `json:"unicode_emoji"`
 }
 
 // Mention returns a string which mentions the role
 func (r *Role) Mention() string {
 	return fmt.Sprintf("<@&%s>", r.ID)
+}
+
+// IconURL returns the URL of the users's banner image.
+//
+//	size:    The size of the desired role icon as a power of two
+//	         Image size can be any power of two between 16 and 4096.
+func (r *Role) IconURL(size string) string {
+	if r.Icon == "" {
+		return ""
+	}
+
+	URL := EndpointRoleIcon(r.ID, r.Icon)
+
+	if size != "" {
+		return URL + "?size=" + size
+	}
+	return URL
 }
 
 // RoleParams represents the parameters needed to create or update a Role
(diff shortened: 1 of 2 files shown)


structs.go
```diff
@@ -1239,13 +1239,51 @@ type Role struct {
...
+	// This is a combination of bit masks; the presence of a certain permission can
+	// be checked by performing a bitwise AND between this int and the flag.
+	Flags RoleFlags `json:"flags"`
 }
 
+// RoleFlags is the flags of "role" (see RoleFlags* consts)
+// https://discord.com/developers/docs/topics/permissions#role-object-role-flags
+type RoleFlags int
+
+// Block containing known RoleFlags values
```
FedorLap2006, line 1259:
```suggestion
// Block containing known RoleFlags values.
```

structs.go
```diff
@@ -1239,13 +1239,51 @@ type Role struct {
...
+const (
+	// RoleFlagInPrompt Role can be selected by members in an onboarding prompt.
```
FedorLap2006, line 1261:
```suggestion
	// RoleFlagInPrompt indicates a Role to be selectable by members in an onboarding prompt.
```

Earlopain: I copied the description straight from the discord docs like I usually do. Any reason to deviate here?

structs.go
```diff
@@ -1355,8 +1355,23 @@ type Role struct {
...
 	// The emoji assigned to this role.
 	UnicodeEmoji string `json:"unicode_emoji"`
+
+	// The flags of the role, which describe extra features of a role.
+	// This is a combination of bit masks; the presence of a certain permission can
+	// be checked by performing a bitwise AND between this int and the flag.
+	Flags RoleFlags `json:"flags"`
 }
 
+// RoleFlags is the flags of "role" (see RoleFlags* consts)
```
FedorLap2006, line 1365:
```suggestion
// RoleFlags represents the flags of a Role.
```

structs.go
```diff
@@ -1355,8 +1355,23 @@ type Role struct {
...
+// https://discord.com/developers/docs/topics/permissions#role-object-role-flags
+type RoleFlags int
+
+// Block containing known RoleFlags values.
+const (
+	// RoleFlagInPrompt indicates a Role to be selectable by members in an onboarding prompt.
```
FedorLap2006, line 1371:
```suggestion
	// RoleFlagInPrompt indicates a Role is selectable by members in an onboarding prompt.
```

FedorLap2006 (changes requested):

PR #1412, opened 2023-07-26
Add missing `thread_name` field for creating forum posts with webhooks

merlinfuchs:
https://discord.com/developers/docs/resources/webhook#execute-webhook-jsonform-params

### DIFF
--- a/webhook.go
+++ b/webhook.go
@@ -37,7 +37,8 @@ type WebhookParams struct {
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 	// Only MessageFlagsSuppressEmbeds and MessageFlagsEphemeral can be set.
 	// MessageFlagsEphemeral can only be set when using Followup Message Create endpoint.
-	Flags MessageFlags `json:"flags,omitempty"`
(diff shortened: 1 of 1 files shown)


webhook.go
```diff
@@ -37,7 +37,8 @@ type WebhookParams struct {
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 	// Only MessageFlagsSuppressEmbeds and MessageFlagsEphemeral can be set.
 	// MessageFlagsEphemeral can only be set when using Followup Message Create endpoint.
-	Flags MessageFlags `json:"flags,omitempty"`
+	Flags      MessageFlags `json:"flags,omitempty"`
+	ThreadName string       `json:"thread_name,omitempty"`
```
FedorLap2006, line 41: A nitpick, but I think it would be better to put it before the flags, to avoid visual separation from other fields.

FedorLap2006:
Also, sorry for the wait. Been quite busy

FedorLap2006:
Thank you for the contribution! Even though it all went on for quite some time.

PR #1415, opened 2023-07-30
nil send/receive channels on Close()

azekeil:
Without the below (or handling this explicitly by the caller), reusing the VoiceChannel causes a panic attempting to send on a closed channel. This ensures the channels are nil'd on Close(), meaning they can be reused without this additional protection.

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -210,6 +210,11 @@ func (v *VoiceConnection) Close() {
 
 		v.wsConn = nil
 	}
+
+	// Nil the send/receive channels in case the VoiceChannel gets reused,
+	// otherwise the nil check in onEvent means we attempt to send to a closed channel.
+	v.OpusSend = nil
+	v.OpusRecv = nil
 }
 
 // AddHandler adds a Handler for VoiceSpeakingUpdate events.


FedorLap2006:
Hi, sorry for the wait.

I don't think we close these two channels anywhere - thus dgo should be able to reuse them just fine.
Although, like with `v.close`, I think it's still a good idea to `close` and set them to `nil`.

cybtachyon:
This was merged in https://github.com/0x00-sys/discordgo/pull/11 and can be closed.

0x00-sys:
> This was merged in https://github.com/0x00-sys/discordgo/pull/11 and can be closed.

Hey, I do have my own maintained fork of the Discord Go library. It does include everything new that Discord has pushed in the last two to three years, which this repository hasn't really followed through and updated. This can cause some breaking changes to different bots depending on what it's oriented about. I'm trying to maintain it specifically because it helps me with my own work. So I'm sort of forced to maintain it myself as well. But you can give it a look and use that if you wish to. Everything that Discord has published, including the newest releases from this September, are already supported.

PR #1424, opened 2023-08-12
feat(session): Adding an event handler for monitoring purposes.

Jacobbrewer1:
Adding an event notifier to be used for monitoring purposes. The change can be used for things such as Prometheus monitoring. This is the result as there is no way to intercept messages through a middleware before the events are pushed to their respective handlers. This approach will force the events through the channel if it is not full.

I have included a respective code example below:

```go
func (a *App) Run() error {
	// Register bot.
	if err := a.RegisterBot(); err != nil {
		return fmt.Errorf("error registering bot: %w", err)
	}

	a.s.AddHandler(func(s *discordgo.Session, r *discordgo.Ready) {
		a.Info(fmt.Sprintf("Logged in as %s#%s", r.User.Username, r.User.Discriminator))
	})

	// Register commands.
	if err := a.RegisterCommands(); err != nil {
		return fmt.Errorf("error registering commands: %w", err)
	}

	if err := a.RegisterDiscordHandlers(); err != nil {
		return fmt.Errorf("error registering discord handlers: %w", err)
	}

	// Start event listener.
	go a.eventListener()

	// Open websocket.
	if err := a.s.Open(); err != nil {
		return fmt.Errorf("error opening connection to Discord: %w", err)
	}
	return nil
}

func (a *App) eventListener() {
	for e := range a.eventNotifier {
		switch t := e.(type) {
		case *discordgo.Event:
			monitoring.TotalDiscordEvents.WithLabelValues(t.Type).Inc()
		default:
			a.Error("Unknown event type", slog.String("type", fmt.Sprintf("%T", e)))
			monitoring.TotalDiscordEvents.WithLabelValues("unknown").Inc()
		}
	}
}
```

The code above will produce the respective graphs in Grafana by using the rate query with Prometheus.

image

### DIFF
--- a/event.go
+++ b/event.go
@@ -60,6 +60,37 @@ type eventHandlerInstance struct {
 	eventHandler EventHandler
 }
 
+// SetEventNotifier is to set the channel that will receive all events. This should be used for debugging and
+// monitoring purposes only. It is not recommended to use this for handling events. This should be set before calling
+// Open() or OpenWithIntents().
+//
+// It is recommended that this is a buffered channel.
+//
+// A good example of the use for this channel is for using it against Prometheus to monitor the number of events
+// received by the bot. This can be done by using the following code:
+//
+//	// Create a buffered channel with a size of 1000.
+//	eventNotifier := make(chan any, 1000)
+//
+//	// Set the event notifier.
+//	dg.SetEventNotifier(eventNotifier)
+//
+//	// Start a goroutine to receive events from the eventNotifier channel.
+//	go func() {
+//		for e := range a.eventNotifier {
+//			switch t := e.(type) {
+//			case *discordgo.Event:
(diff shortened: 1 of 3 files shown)


FedorLap2006:
I'm not sure this change is necessary, since for monitoring purposes you can already use `interface{}` or `*discordgo.Event` as the arguments in the handler to intercept all incoming events.

Jacobbrewer1:
> I don't think this change is necessary, since for monitoring purposes you can already use `interface{}` or `*discordgo.Event` as the data parameter in the handler to intercept all incoming events. They are handled asyncrhonously (there's a `SyncEvents` to disable that, but it's irrelevant). In case you need them to be in sync, you can always put this channel logic into the handler.

I see your point with regards to the event. The only thing I have found is that there is no easy way to intercept the messages with a `middleware` like approach. That was what this change was intended to solve.

PR #1426, opened 2023-08-13
feat(member): add helper function for display name

JulienTant:
This PR adds a simple helper on the `Member` struct to retrieve a member display name on the server, by either using the member nick if one is set, or the profile display name (called `GlobalName` in the API)

Would love to see this as part of https://github.com/bwmarrin/discordgo/milestone/25 :)

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -1394,6 +1394,15 @@ func (m *Member) AvatarURL(size string) string {
 
 }
 
+// DisplayName returns the member's guild nickname if they have one,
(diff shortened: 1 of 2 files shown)


ShadiestGoat:
Perhaps you should also add a fallback for the members's username? If I'm understanding the discord docs correctly, the global name isn't always set - in which case username should prob be used.

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1427, opened 2023-08-14
feat(MessageSend): add flags

ozraru:
https://discord.com/developers/docs/resources/channel#create-message

### DIFF
--- a/message.go
+++ b/message.go
@@ -234,7 +234,7 @@ type MessageSend struct {
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 	Reference       *MessageReference       `json:"message_reference,omitempty"`
 	StickerIDs      []string                `json:"sticker_ids"`
-	Flags           int                     `json:"flags,omitempty"`
+	Flags           MessageFlags            `json:"flags,omitempty"`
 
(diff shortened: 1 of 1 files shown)


message.go
```diff
@@ -234,6 +234,7 @@ type MessageSend struct {
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 	Reference       *MessageReference       `json:"message_reference,omitempty"`
 	StickerIDs      []string                `json:"sticker_ids"`
+	Flags           int                     `json:"flags,omitempty"`
```
Earlopain, line 237: Feels to me like this should be the flag type that  describes the possible values in the other PR. I know there are places that do it like this as well but that seems like legacy @FedorLap2006 ?

FedorLap2006 (changes requested):
As @Earlopain mentioned, `Flags` should have `MessageFlags` type instead of `int`.

message.go
ozraru: Fixed. Sorry for my rudimentary mistake.

FedorLap2006 (approved):

FedorLap2006:
Thanks for your contribution!

PR #1456, opened 2023-10-17
Adding sync_id for presences

ashkon91:
Adding support for `sync_Id` when unpacking Presence activities. This will give users the ability to get the `Spotify Track ID` from a presence event


mrbentarikau:
every id is capitalized in the code, so it should be SyncID instead of SyncId

PR #1475, opened 2024-01-04
fix(buttons): omit empty emoji field

### LINKED ISSUE #1474: Error creating message components without emojis
Starting today (few hours ago), I was having issues with my bot creating embeds that did not have emjois defined on the components. It appears that Discord changed their API to require the emoji field to be omitted when it is not populated, instead of an empty struct with no fields.

Error message:
```HTTP 400 Bad Request, {"message": "Invalid Form Body", "code": 50035, "errors": {"components": {"0": {"components": {"0": {"emoji": {"name": {"_errors": [{"code": "BUTTON_COMPONENT_INVALID_EMOJI", "message": "Invalid emoji"}]}}}}}}}}```

I forked this and confirmed that this fixes my issue and will create a PR for consideration or reference to fix in a larger batch of fixes.

TheGrandPackard:
Fixes #1474

### DIFF
--- a/components.go
+++ b/components.go
@@ -132,10 +132,10 @@ type ComponentEmoji struct {
 
 // Button represents button component.
 type Button struct {
-	Label    string         `json:"label"`
-	Style    ButtonStyle    `json:"style"`
-	Disabled bool           `json:"disabled"`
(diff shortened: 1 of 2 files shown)


FedorLap2006:
Thanks for your contribution!

PR #1476, opened 2024-01-04
Modify SelectMenuOption Emoji field to omitempty

phamill:
The emoji validation fixed in pr #1475 did not cover another emoji validation case with the `SelectMenuOption`. This fixes forward this breaking change from the API -

Example response from using a SelectMenuOption:

```
HTTP 400 Bad Request, {"message": "Invalid Form Body", "code": 50035, "errors": {"components": {"0": {"components": {"0": {"options": {"0": {"emoji": {"name": {"_errors": [{"code": "BUTTON_COMPONENT_INVALID_EMOJI", "message": "Invalid emoji"}]}}}}}}}}}}
```

### DIFF
--- a/components.go
+++ b/components.go
@@ -166,10 +166,10 @@ func (Button) Type() ComponentType {
 
 // SelectMenuOption represents an option for a select menu.
 type SelectMenuOption struct {
-	Label       string         `json:"label,omitempty"`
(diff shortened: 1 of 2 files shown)


FedorLap2006:
Thanks for your contribution!

PR #1477, opened 2024-01-04
update SelectMenuOption.Emoji to be a pointer, and omitted when empty

mpawlowski:
This change is similar to https://github.com/bwmarrin/discordgo/pull/1475 and related to https://github.com/bwmarrin/discordgo/issues/1474, where the Emoji field must be omitted entirely and not empty for SelectMenuOptions


mpawlowski:
This was done 10 min earlier from another PR :)

PR #1483, opened 2024-01-12
fix(message)!: omit empty Components and Embeds

AlexeyOplachko:
A while ago there was a small discussion on this topic in discord [down from this message](https://discord.com/channels/118456055842734083/155361364909621248/1172388479964028959)
Implemented as was described(at least I think that's what was meant about converting slices to slices of pointers), here's a recording of it working
DiscordPTB_2024-01-12_22-32-58
Snippet with example of code that's used for suppression.
```go
func MessageUpdate(s *discordgo.Session, m *discordgo.MessageUpdate) {
	if len(m.Embeds) > 0 {
		messageEdit := discordgo.NewMessageEdit(m.ChannelID, m.ID)
		messageEdit.Flags = discordgo.MessageFlagsSuppressEmbeds
		s.ChannelMessageEditComplex(messageEdit)
		s.ChannelMessageSend(m.ChannelID, "Suppressed embed")
	}
}
```

### DIFF
--- a/message.go
+++ b/message.go
@@ -233,7 +233,7 @@ type MessageSend struct {
 	Content         string                  `json:"content,omitempty"`
 	Embeds          []*MessageEmbed         `json:"embeds"`
 	TTS             bool                    `json:"tts"`
-	Components      []MessageComponent      `json:"components"`
+	Components      []*MessageComponent     `json:"components"`
 	Files           []*File                 `json:"-"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 	Reference       *MessageReference       `json:"message_reference,omitempty"`
@@ -251,8 +251,8 @@ type MessageSend struct {
 // is also where you should get the instance from.
 type MessageEdit struct {
 	Content         *string                 `json:"content,omitempty"`
-	Components      []MessageComponent      `json:"components"`
-	Embeds          []*MessageEmbed         `json:"embeds"`
+	Components      []*MessageComponent     `json:"components,omitempty"`
+	Embeds          []*MessageEmbed         `json:"embeds,omitempty"`
 	AllowedMentions *MessageAllowedMentions `json:"allowed_mentions,omitempty"`
 	Flags           MessageFlags            `json:"flags,omitempty"`
 	// Files to append to the message
--- a/message.go
+++ b/message.go
@@ -251,7 +251,7 @@ type MessageSend struct {
 // is also where you should get the instance from.
 type MessageEdit struct {
 	Content         *string                 `json:"content,omitempty"`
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -233,7 +233,7 @@ type MessageSend struct {
 	Content         string                  `json:"content,omitempty"`
 	Embeds          []*MessageEmbed         `json:"embeds"`
 	TTS             bool                    `json:"tts"`
-	Components      []MessageComponent      `json:"components"`
+	Components      []*MessageComponent     `json:"components"`
```
FedorLap2006, line 236: `MessageComponent` is an interface. If the array going to consist of pointers, it's going to be extremely inconvenient to pass things in. 
```go
var btn discordgo.MessageComponent = discordgo.Button{}
components := []*discordgo.MessageComponent{
    discordgo.Button{...}, // Not possible
    &btn,
}

File: message.go [hunk 2]
```diff
@@ -251,8 +251,8 @@ type MessageSend struct {
 // is also where you should get the instance from.
 type MessageEdit struct {
 	Content         *string                 `json:"content,omitempty"`
-	Components      []MessageComponent      `json:"components"`
-	Embeds          []*MessageEmbed         `json:"embeds"`
+	Components      []*MessageComponent     `json:"components,omitempty"`
```
FedorLap2006, line 254: Same as `MessageSend.Components`. However, since the field could be `[]` the array should be a pointer instead.
```suggestion
	Components      *[]MessageComponent     `json:"components,omitempty"`
```

message.go
```diff
@@ -251,8 +251,8 @@ type MessageSend struct {
...
+	Embeds          []*MessageEmbed         `json:"embeds,omitempty"`
```
FedorLap2006, line 255:
```suggestion
	Embeds          *[]*MessageEmbed         `json:"embeds,omitempty"`
```

FedorLap2006 (changes requested):

restapi.go
```diff
@@ -1797,14 +1797,14 @@ func (s *Session) ChannelMessageEditComplex(m *MessageEdit, options ...RequestOp
...
-			m.Embeds = []*MessageEmbed{m.Embed}
+			m.Embeds = &[]*MessageEmbed{m.Embed}
 		} else {
 			err = fmt.Errorf("cannot specify both Embed and Embeds")
 			return
 		}
 	}
 
-	for _, embed := range m.Embeds {
+	for _, embed := range *m.Embeds {
```
FedorLap2006, line 1807: `m.Embeds` can be `nil`, so a `if m.Embeds != nil` safeguard is required here

FedorLap2006 (changes requested):

FedorLap2006 (approved):

PR #1494, opened 2024-02-10
feat(message)!: use partial sticker objects

Earlopain:
https://discord.com/developers/docs/resources/channel#message-object
https://discord.com/developers/docs/resources/sticker#sticker-item-object

There exists a property called `stickers` that contain the full sticker information but that is deprecated and not part of the struct here.


Earlopain:
Ah, sorry. I somehow missed to update the doc comment.

PR #1500, opened 2024-02-27
feat(UserGuilds): support with_counts parameter

hhaste:
This PR adds support for the `with_count` query parameter on the `UserGuilds()` function. 

Additionally, the `limit` of returned guilds has been increased from 100 to 200.

Reference: https://discord.com/developers/docs/resources/user#get-current-user-guilds

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -424,10 +424,11 @@ func (s *Session) UserGuildMember(guildID string, options ...RequestOption) (st 
 }
 
 // UserGuilds returns an array of UserGuild structures for all guilds.
-// limit     : The number guilds that can be returned. (max 100)
-// beforeID  : If provided all guilds returned will be before given ID.
-// afterID   : If provided all guilds returned will be after given ID.
-func (s *Session) UserGuilds(limit int, beforeID, afterID string, options ...RequestOption) (st []*UserGuild, err error) {
+// limit       : The number guilds that can be returned. (max 200)
+// beforeID    : If provided all guilds returned will be before given ID.
+// afterID     : If provided all guilds returned will be after given ID.
+// withCounts  : Whether to include approximate member and presence counts or not
+func (s *Session) UserGuilds(limit int, beforeID, afterID string, withCounts bool, options ...RequestOption) (st []*UserGuild, err error) {
 
 	v := url.Values{}
 
@@ -439,6 +440,9 @@ func (s *Session) UserGuilds(limit int, beforeID, afterID string, options ...Req
 	}
 	if beforeID != "" {
 		v.Set("before", beforeID)
+	}
+	if withCounts {
+		v.Set("with_counts", "true")
 	}
 
 	uri := EndpointUserGuilds("@me")
--- a/restapi.go
+++ b/restapi.go
@@ -427,7 +427,7 @@ func (s *Session) UserGuildMember(guildID string, options ...RequestOption) (st 
 // limit       : The number guilds that can be returned. (max 200)
 // beforeID    : If provided all guilds returned will be before given ID.
 // afterID     : If provided all guilds returned will be after given ID.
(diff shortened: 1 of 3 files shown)


Earlopain:
I think it would be better to add a new method `UserGuildsWithCounts` to not break current code. The old method can then just dispatch to the new one. Sad that go doesn't have default args...

You could also update struct member docs for `ApproximateMemberCount` and `ApproximatePresenceCount` since that doesn't currently mention this new way to obtain the information. Should probably switch that over to the methods you can call in discordgo for this instead of the verbatim copy from the discord api docs.

hhaste:
> I think it would be better to add a new method `UserGuildsWithCounts` to not break current code. The old method can then just dispatch to the new one.

There would be no `withCounts` argument on the `UserGuildsWithCounts` function, so having the old method dispatch to the new one wouldn't work. I could still make a new function, but I felt it would be redundant.

Earlopain:
Oh, right. Tbh, I don't know how this is usually handled, I haven't been around here for very long. @FedorLap2006 should be able to give some input there.

FedorLap2006:
I think it would be better to make a breaking change, rather than introducing `UserGuildsWithCounts`.

restapi.go
```diff
@@ -424,10 +424,11 @@ func (s *Session) UserGuildMember(guildID string, options ...RequestOption) (st
...
 
 // UserGuilds returns an array of UserGuild structures for all guilds.
-// limit     : The number guilds that can be returned. (max 100)
-// beforeID  : If provided all guilds returned will be before given ID.
-// afterID   : If provided all guilds returned will be after given ID.
-func (s *Session) UserGuilds(limit int, beforeID, afterID string, options ...RequestOption) (st []*UserGuild, err error) {
+// limit       : The number guilds that can be returned. (max 200)
+// beforeID    : If provided all guilds returned will be before given ID.
+// afterID     : If provided all guilds returned will be after given ID.
+// withCounts  : Whether to include approximate member and presence counts or not
```
FedorLap2006, line 430:
```suggestion
// withCounts  : Whether to include approximate member and presence counts or not.
```

structs.go
```diff
@@ -1274,6 +1274,14 @@ type UserGuild struct {
...
 	Permissions int64          `json:"permissions,string"`
 	Features    []GuildFeature `json:"features"`
+
+	// Approximate number of members in this guild
+	// NOTE: this field is only filled when withCounts is true
+	ApproximateMemberCount int `json:"approximate_member_count"`
+
+	// Approximate number of members in this guild
+	// NOTE: this field is only filled when withCounts is true
+	ApproximatePresenceCount int `json:"approximate_presence_count"`
```
FedorLap2006, line 1284:
```suggestion

	// Approximate number of members in this guild.
	// NOTE: this field is only filled when withCounts is true.
	ApproximateMemberCount int `json:"approximate_member_count"`

	// Approximate number of members in this guild.
	// NOTE: this field is only filled when withCounts is true.
	ApproximatePresenceCount int `json:"approximate_presence_count"`
```

FedorLap2006 (changes requested):

structs.go
```diff
@@ -1274,6 +1274,14 @@ type UserGuild struct {
 	Owner       bool           `json:"owner"`
 	Permissions int64          `json:"permissions,string"`
 	Features    []GuildFeature `json:"features"`
+
+	// Approximate number of members in this guild.
+	// NOTE: this field is only filled when withCounts is true.
+	ApproximateMemberCount int `json:"approximate_member_count"`
+
+	// Approximate number of members in this guild.
```
FedorLap2006, line 1282:
```suggestion
	// Approximate number of non-offline members in this guild.
```

FedorLap2006 (approved):

PR #1516, opened 2024-04-09
add nonce for interaction and message

Sherlockouo:


FedorLap2006 (changes requested):
Please remove changes unrelated to the `Interaction` and `Message` structs

Also after looking into discord's documentation, it seems `Interaction.Nonce` doesn't exist, or at least is not documented. Usually we try not to add undocumented features.

PR #1521, opened 2024-04-17
Updating documentation to reflect the deprecation of Session.UserAvatar

arekouzounian:
I found that the description for the 'Avatar' field in user.go tells you to use Session.UserAvatar to grab the actual avatar, but the corresponding doc for Session.UserAvatar notes that it's deprecated, so I thought I would make that change in user.go as well to make sure people aren't calling a deprecated function.

### DIFF
--- a/user.go
+++ b/user.go
@@ -52,8 +52,8 @@ type User struct {
 	// The user's username.
 	Username string `json:"username"`
 
-	// The hash of the user's avatar. Use Session.UserAvatar
-	// to retrieve the avatar itself.
+	// The hash of the user's avatar. Use Session.UserAvatarDecode
+	// to retrieve the avatar itself (Session.UserAvatar is deprecated).
 	Avatar string `json:"avatar"`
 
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -52,8 +52,8 @@ type User struct {
 	// The user's username.
 	Username string `json:"username"`
 
-	// The hash of the user's avatar. Use Session.UserAvatar
-	// to retrieve the avatar itself.
+	// The hash of the user's avatar. Use Session.UserAvatarDecode
+	// to retrieve the avatar itself (Session.UserAvatar is deprecated).
```
FedorLap2006, line 56: I think it's better just to replace the function.
Also we should probably mention the `AvatarURL` here.
```suggestion
	// The hash of the user's avatar.
	// Use AvatarURL to retrieve the avatar URL or Session.UserAvatarDecode to retrieve the avatar itself.
```

FedorLap2006 (changes requested):

PR #1535, opened 2024-06-09
feat(interaction): Add GetOption method on Application Interaction Command

NotYusta:
Add a utility get option function, it will help.

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -326,6 +326,18 @@ type ApplicationCommandInteractionData struct {
 	TargetID string `json:"target_id"`
 }
 
+// Get the interaction data option by name
+func (d ApplicationCommandInteractionData) GetOption(name string) (option *ApplicationCommandInteractionDataOption) {
+	for _, opt := range d.Options {
+		if opt.Name == name {
+			option = opt
+			break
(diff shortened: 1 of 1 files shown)


interactions.go
```diff
@@ -326,6 +326,18 @@ type ApplicationCommandInteractionData struct {
 	TargetID string `json:"target_id"`
 }
 
+// Get the interaction data option by name
+func (d ApplicationCommandInteractionData) GetOption(name string) (option *ApplicationCommandInteractionDataOption) {
```
TheAdamSmith, line 330: Could we have these return an exists boolean as well?

TheAdamSmith (commented):
Not an approver for this repo. But would love to see this get implemented!

interactions.go
NotYusta: Well you can just null-check it, an exists boolean isn't needed i suppose.

FedorLap2006 (approved):

FedorLap2006 (changes requested):
Please note the check.

FedorLap2006:
Thanks for the contribution!

PR #1465, opened 2023-11-18
feat(UserFlags): add BotHTTPInteractions and ActiveDeveloper

Ju0x:
Added following UserFlags:
- BOT_HTTP_INTERACTIONS
- SPAMMER
- ACTIVE_DEVELOPER

The spammer flag is currently undocumented by discord, but seen in other discord libraries.

PR #1548, opened 2024-07-23
feat(User): Add Spammer flag found in User PublicFlags

brimstone:
This adds a UserFlagSpammer found in User PublicFlags. I've tested this against one guild and it matches output from discord.py's UserFlags https://github.com/Rapptz/discord.py/blob/ff638d393d0f5a83639ccc087bec9bf588b59a22/discord/enums.py#L516


brimstone:
I seem to have found https://github.com/bwmarrin/discordgo/pull/1465 Maybe enough time has passed to add this flag?

PR #1560, opened 2024-08-23
Expose s.request as s.RequestRaw so one can send JSON directly

ldemailly:
Otherwise you need to either artificially make a map[string]any from the json or duplicate
```go
	if bucketID == "" {
		bucketID = strings.SplitN(urlStr, "?", 2)[0]
	}
	return s.RequestWithLockedBucket(method, urlStr, contentType, b, s.Ratelimiter.LockBucket(bucketID), sequence, options...)
```
and I think the details of the rate limit/locking logic should stay hidden rather than duplicated in caller code

### DIFF
--- a/restapi.go
+++ b/restapi.go
@@ -178,13 +178,14 @@ func (s *Session) RequestWithBucketID(method, urlStr string, data interface{}, b
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


FedorLap2006 (approved):

ldemailly:
poke

PR #1578, opened 2024-10-03
Add message URL helpers

elliotwms:
This PR adds a helper func for generating canonical Discord message URLs


elliotwms:
Converting back to draft as it turns out Guild ID is no longer provided on the message body, so not guaranteed to exist

PR #1584, opened 2024-10-25
Add InteractionCallbackResponse

elliotwms:
The interaction callback endpoint returns an interaction callback response as described in the docs: https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-callback-interaction-callback-response-object

This PR adds the structs, and updates the `InteractionCallback` func to return them

### DIFF
--- a/examples/autocomplete/main.go
+++ b/examples/autocomplete/main.go
@@ -73,7 +73,7 @@ var (
 			switch i.Type {
 			case discordgo.InteractionApplicationCommand:
 				data := i.ApplicationCommandData()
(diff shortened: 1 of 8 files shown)


elliotwms:
Just checked the actual Discord API behaviour for both `InteractionResponseDeferredChannelMessageWithSource` and `InteractionResponseChannelMessageWithSource`, and it's a `204 No Content` in both cases.

This response body may be returned on other types but I've not been able to test it

elliotwms:
I don't currently have a use for this, and am not able to test it, so I'll leave the PR here in case someone would like to pick it up

PR #1586, opened 2024-11-23
restapi: retry on 500s as well as 502s

patrickxia:
This is a sequel to issue #227.

I run a matterbridge bridge and I'm getting 500s from the Discord API from time to time. We should treat them the same as a 503, instead of just immediately giving up.

Since we're adding entries to the switch statement, also clean up the naked 429 to be consistent.


patrickxia:
I looked over my logs and found a 503 as well so I've updated my branch. We might as well cluster 500, 503, 504 all the same because they have similar meanings from the client perspective.

Mxrk:
At the moment I'm running into a few 504 errors from time to time and this seems like a good fix. What's currently blocking this PR?

FedorLap2006 (approved):

PR #1589, opened 2025-01-04
Fixes to appease go static checker

Suhaibinator:
Resolve some complaints from the static go checker

### DIFF
--- a/event.go
+++ b/event.go
@@ -51,7 +51,6 @@ func registerInterfaceProvider(eh EventInterfaceProvider) {
 		// fmt.Errorf("event %s already registered", eh.Type())
 	}
 	registeredInterfaceProviders[eh.Type()] = eh
-	return
 }
 
 // eventHandlerInstance is a wrapper around an event handler, as functions
@@ -102,12 +101,14 @@ func (s *Session) addEventHandlerOnce(eventHandler EventHandler) func() {
 // to a struct corresponding to the event for which you want to listen.
 //
 // eg:
-//     Session.AddHandler(func(s *discordgo.Session, m *discordgo.MessageCreate) {
-//     })
+//
+//	Session.AddHandler(func(s *discordgo.Session, m *discordgo.MessageCreate) {
+//	})
 //
 // or:
-//     Session.AddHandler(func(s *discordgo.Session, m *discordgo.PresenceUpdate) {
-//     })
+//
+//	Session.AddHandler(func(s *discordgo.Session, m *discordgo.PresenceUpdate) {
+//	})
 //
 // List of events can be found at this page, with corresponding names in the
 // library for each event: https://discord.com/developers/docs/topics/gateway#event-names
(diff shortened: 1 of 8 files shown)


event.go
```diff
@@ -51,7 +51,6 @@ func registerInterfaceProvider(eh EventInterfaceProvider) {
 		// fmt.Errorf("event %s already registered", eh.Type())
 	}
 	registeredInterfaceProviders[eh.Type()] = eh
-	return
```
Suhaibinator, line 54: Unnecessary return removed

oauth2_test.go
```diff
@@ -52,6 +52,4 @@ func ExampleApplication() {
 	// Delete the application we created.
 	err = dg.ApplicationDelete(ap.ID)
 	log.Printf("Delete: err: %+v\n", err)
-
-	return
```
Suhaibinator, line 56: Unnecessary return removed

state.go
```diff
@@ -1061,7 +1062,7 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
 				oldCopy := *old
 				t.BeforeUpdate = &oldCopy
 			}
-			err = s.ChannelAdd(t.Channel)
```
Suhaibinator, line 1064: This error value is never used

state.go
```diff
@@ -1078,7 +1079,7 @@ func (s *State) OnInterface(se *Session, i interface{}) (err error) {
...
```
Suhaibinator, line 1081: This error value is never used

voice.go
```diff
@@ -501,8 +501,6 @@ func (v *VoiceConnection) onEvent(message []byte) {
 	default:
 		v.log(LogDebug, "unknown voice operation, %d, %s", e.Operation, string(e.RawData))
 	}
-
-	return
```
Suhaibinator, line 505: Unnecessary return removed

voice.go
```diff
@@ -787,7 +785,8 @@ func (v *VoiceConnection) opusSender(udpConn *net.UDPConn, close <-chan struct{}
 			sequence++
 		}
 
-		if (timestamp + uint32(size)) >= 0xFFFFFFFF {
```
Suhaibinator, line 790: The value given is greater than the max value of Uint32 which is impossible. Instead, we should check for integer ovrflow.

Suhaibinator:
@FedorLap2006 @bwmarrin

Suhaibinator:
@FedorLap2006 @bwmarrin bump

FedorLap2006 (approved):
LGTM

Suhaibinator:
@FedorLap2006 Can you merge please? I don't have merge perms

state.go
FedorLap2006: I think it's better to fix the shadowing rather than its consequences.

FedorLap2006 (changes requested):
After looking into it, I would change a couple of things, but other than that still LGTM.

PR #1593, opened 2025-01-18
Voice overhaul

ozraru:
I've rewrited entire code related to Voice.

This is very big pull request.
I'm sorry, I understand that it is difficult to review.
But I believe that this is valuable to check.

I'm not good at English. Wording fixes are welcome.
I'm not sure about the naming. Naming changes are also welcome.

I want to support E2EE in the future but it's not included in this pull request.

---

## Summary
- Use context to manage goroutines and connections
- Eliminate busy wait using sync.Cond
- Use Voice Gateway Version 8
- Change to new transport encryption mode
  - `aead_aes256_gcm_rtpsize` and `aead_xchacha20_poly1305_rtpsize`
- Resume Voice Conenction
  - https://discord.com/developers/docs/topics/voice-connections#resuming-voice-connection
- Delete reconnection (excludes resume)
  - It is very rare and it should be handled by an application
- Deliver unrecoverable error to the application

- Edit some examples to work with these changes

## Breaking changes
- misc
  - Some functions require context.Context in argument
    - It can be used to set timeout
- Session
  - Renamed ChannelVoiceJoinManual to VoiceStateUpdate
  - Remove ShouldReconnectVoiceOnSessionError
    - Because reconnection are deleted
- VoiceConenction
  - Removed Ready
    - Use Status instead
  - Removed Debug
    - already deprecated
  - Removed ChangeChannel
    - Use Session.VoiceStateUpdate
  - Removed Close
    - Kill is alternative, but why not use Disconnect?
- Packet
  - Removed Type
    - Separated to Flags and PayloadType

---

note: Currently using transport encryption mode(`xsalsa20_poly1305`) and voice gateway version(connecting without a version) is already discontinued according to the document. (but it works currently :thinking: )
> Versions below 4 as well as the default version behaviour will be discontinued as of November 18th, 2024. Connections without a version or with a version less than version 4 will be rejected as of this date.
> https://discord.com/developers/docs/topics/voice-connections#voice-gateway-versioning

> The deprecated encryption modes above will be discontinued as of November 18th, 2024. As of this date the voice gateway will not allow you to connect with one of the deprecated encryption modes.
> https://discord.com/developers/docs/topics/voice-connections#transport-encryption-and-sending-voice

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -10,63 +10,85 @@
 package discordgo
 
 import (
+	"context"
+	"crypto/aes"
+	"crypto/cipher"
 	"encoding/binary"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
-	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/gorilla/websocket"
-	"golang.org/x/crypto/nacl/secretbox"
+	"golang.org/x/crypto/chacha20poly1305"
 )
 
 // ------------------------------------------------------------------------------------------------
 // Code related to both VoiceConnection Websocket and UDP connections.
 // ------------------------------------------------------------------------------------------------
 
+type VoiceConnectionStatus int
+
+const (
+	VoiceConnectionStatusInvalid    VoiceConnectionStatus = iota // status not specified, bug?
+	VoiceConnectionStatusNew                                     // initiating connection
+	VoiceConnectionStatusConnecting                              // connecting websocket and udp
+	VoiceConnectionStatusReady                                   // ready to send/receive audio
+	VoiceConnectionStatusDead                                    // already dead(error or disconnected normally)
+)
+
 // A VoiceConnection struct holds all the data and functions related to a Discord Voice Connection.
 type VoiceConnection struct {
-	sync.RWMutex
+	Cond *sync.Cond
 
-	Debug        bool // If true, print extra logging -- DEPRECATED
-	LogLevel     int
-	Ready        bool // If true, voice is ready to send/receive audio
-	UserID       string
-	GuildID      string
-	ChannelID    string
-	deaf         bool
-	mute         bool
-	speaking     bool
-	reconnecting bool // If true, voice connection is trying to reconnect
+	// Status of this connection. Please don't change.
+	// New -> Connecting <-> Ready
+	// any -> Dead
+	Status VoiceConnectionStatus
 
-	OpusSend chan []byte  // Chan for sending opus audio
-	OpusRecv chan *Packet // Chan for receiving opus audio
+	// Closed if this VoiceConection status become Dead
+	Dead <-chan struct{}
+	dead chan struct{}
 
-	wsConn  *websocket.Conn
-	wsMutex sync.Mutex
+	// contains unrecoverable error
+	// if not nil, Status should be Dead
+	Err error
+
+	LogLevel int
+	GuildID  string
+
+	deaf     bool
+	mute     bool
+	speaking bool
+
+	OpusSend chan []byte  // Chan for sending opus audio, automatically closed after dead, DON'T CLOSE YOURSELF
+	OpusRecv chan *Packet // Chan for receiving opus audio, automatically closed after dead, DON'T CLOSE YOURSELF
+
+	// can be nil, use only for send message
+	// mostly this is available connection or nil, but rarely closed connection
+	wsConn *websocket.Conn
+
+	// calling this may close websocket and all related connection.
+	wsCancel context.CancelFunc
+
 	udpConn *net.UDPConn
+
 	session *Session
 
 	sessionID string
-	token     string
-	endpoint  string
 
-	// Used to send a close signal to goroutines
-	close chan struct{}
-
-	// Used to allow blocking until connected
-	connected chan bool
-
-	// Used to pass the sessionid from onVoiceStateUpdate
-	// sessionRecv chan string UNUSED ATM
-
-	op4 voiceOP4
 	op2 voiceOP2
+	op4 voiceOP4
+
+	cipher cipher.AEAD
 
 	voiceSpeakingUpdateHandlers []VoiceSpeakingUpdateHandler
+
+	seqAck atomic.Int32 // for heartbeat and resume
 }
 
 // VoiceSpeakingUpdateHandler type provides a function definition for the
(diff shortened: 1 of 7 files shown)


voice.go
```diff
@@ -231,314 +201,508 @@ type VoiceSpeakingUpdate struct {
...
 
-		// Start the opusReceiver
-		if !v.deaf {
-			if v.OpusRecv == nil {
-				v.OpusRecv = make(chan *Packet, 2)
+		case 2: // READY
+			op2 := voiceOP2{}
+
+			if err := json.Unmarshal(e.RawData, &op2); err != nil {
+				err := fmt.Errorf("OP2 unmarshal error, %s, %s", err, string(e.RawData))
```
fr-str, line 546: it's recommended to use `%w` with [fmt.Errorf](https://pkg.go.dev/fmt#Errorf) when you return it to client/user so they can match it with `errors.Is` and act on it. Otherwise you are hiding the underlying error type
```suggestion
				err := fmt.Errorf("OP2 unmarshal error, %w, %s", err, string(e.RawData))
```

voice.go
```diff
@@ -113,111 +138,56 @@ func (v *VoiceConnection) Speaking(b bool) (err error) {
...
-		v.session.wsMutex.Unlock()
-		v.sessionID = ""
-	}
-	v.Unlock()
-
-	// Close websocket and udp connections
-	v.Close()
-
-	v.log(LogInformational, "Deleting VoiceConnection %s", v.GuildID)
+	v.log(LogInformational, "called")
```
fr-str, line 158:
```suggestion
	v.log(LogInformational, "called 'Kill'")
```

voice.go
```diff
@@ -113,111 +138,56 @@ func (v *VoiceConnection) Speaking(b bool) (err error) {
...
-
-		v.log(LogInformational, "closing websocket")
-		err = v.wsConn.Close()
-		if err != nil {
-			v.log(LogError, "error closing websocket, %s", err)
-		}
-
-		v.wsConn = nil
-	}
+	v.log(LogInformational, "done")
```
fr-str, line 181:
```suggestion
	v.log(LogInformational, "voice connection killed")
```

voice.go
```diff
@@ -231,314 +201,508 @@ type VoiceSpeakingUpdate struct {
...
+	Mode      string `json:"mode"`
+}
+
+// A voiceOP8 stores the data for the voice operation 8 websocket event HELLO
+type voiceOP8 struct {
+	HeartbeatInterval int `json:"heartbeat_interval"`
 }
 
 // WaitUntilConnected waits for the Voice Connection to
 // become ready, if it does not become ready it returns an err
```
fr-str, line 250:
```suggestion
// waitUntilStatus waits for connection to be in given VoiceConnectionStatus
// returns error if context timeout or VoiceConnection.Err
```

voice.go
```diff
@@ -231,314 +201,508 @@ type VoiceSpeakingUpdate struct {
...
 		return
 	}
 
-	// TODO temp? loop to wait for the SessionID
-	i := 0
-	for {
-		if v.sessionID != "" {
-			break
-		}
+	go v.websocket(context.TODO(), *ev.Endpoint, ev.Token)
```
fr-str, line 299: what's the plan with this `context.TODO()`?
`VoiceConnection.websocket` is used only here so maybe remove context.Context argument and initialize it inside `VoiceConnection.websocket` function

ozraru: 1. Normally, top-level context should be generated in the top-level function (= main() of the application), but discordgo(.Session) doesn't accept it. So it is 'TODO'.
1. Functions related to Session (ex: Session.Open, Session.listen) may also have context to manage the lifecycle, like this patch.
It may deprecate Session.Close() and replace with cancel of context, and passing that context to here would resolve `TODO: Add support for Voice WS/UDP` (in the comment of Session.Close)
It is out of the scope of this patch, but websocket() supports being canceled from the caller.

voice.go
```diff
@@ -231,314 +201,508 @@ type VoiceSpeakingUpdate struct {
...
-			v.OpusSend = make(chan []byte, 2)
-		}
-		go v.opusSender(v.udpConn, v.close, v.OpusSend, 48000, 960)
+		switch e.Operation {
 
-		// Start the opusReceiver
-		if !v.deaf {
-			if v.OpusRecv == nil {
-				v.OpusRecv = make(chan *Packet, 2)
+		case 2: // READY
```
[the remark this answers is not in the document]
ozraru: It would be used only here, so why comment is not enough?

ozraru (commented):
Thanks for your review!
I wrote response or merged each comment or suggestion.

voice.go
fr-str: ok, sounds good

voice.go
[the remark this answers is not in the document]
fr-str: i'm referring to all cases in this switch statement. For me, it's a preference and will help avoid confusion in the future. `case 5:`  does not have a comment, for example. Additionally, people will change code but will forget to change comments which may cause confusion, if we have consts in place of 2, 3, 4, 5 code will be more self documenting and comments like `// READY`  won't be needed.

Of course these are the rules i like to follow, i'm just suggesting, not requiring.

sheo9000:
Could you make a proper example of voice receiver? I can't get it to work, and `examples/voice_receive/main.go` is not compileable
```
p.PayloadType undefined (type *discordgo.Packet has no field or method PayloadType)
```

galberti:
@ozraru Your branch works well, but after I while I get:

voice.go:207:failure() voice unrecoverable error, voice websocket closed, websocket: close 4022: Disconnected: Call terminated.

and the audio connections is not recovered.

RadCraftplay:
> voice.go:207:failure() voice unrecoverable error, voice websocket closed, websocket: close 4022: Disconnected: Call terminated.
> 
> and the audio connections is not recovered.

This has been fixed in #1629. These changes probably would have to be applied similarly in this branch. Vendoring discordgo and merging PRs to your fork manually is probably the best idea

Fl0GUI:
> I'd really like to have this working, so I'll keep debugging to try and find the cause.

I compared packet captures between this pr and discord.js when connecting to a channel. (I can't share the pcaps for security/privacy reasons, unless I spend a lot of effort anonymising it)

Some noteworthy differences/messages:

* when sending `op: 0 identify` discord.js uses `max_dave_protocol_version: 1` and discordgo uses `0`.
* discordgo sends an `op: 3 heartbeat` with a `seq_ack: -1` right after `op: 0 identify` while discord.js doesn't. It doesn't seem to break anything.
* After `op: 0 identify` the same packages are received by both implementation: operations 2, 11, 5, 18, and 20. 18 and 20 are undocumented.
* when sending `op: 1 select protocol` the discordgo address has trailing null bytes in the json string value (ie. `"10.10.10.10\u0000\u0000..."`).
* after the last `op: 4 session description` and `op: 15` (undocumented), discord.js receives encrypted messages while discordgo receives nothing.

Long story short, it seems like receiving `op 5: speaking` is only over the dave protocol. And since discordgo doesn't implement the dave protocol, no speaking updates are received 😔
But also, the happy flow channel connection of this pr is almost identical to discord.js.

I might give the dave protocol whitepaper a read, and maybe give an attempt at an implementation.

PR #1599, opened 2025-02-10
Add server-backup example

zazer0:
Per title; adds basic server backup bot example. Iterates all channels it has access to; so make sure to permission it suitably :))

### DIFF
--- a/examples/server_backup/main.go
+++ b/examples/server_backup/main.go
@@ -1,0 +1,387 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/bwmarrin/discordgo"
+)
+
+// Bot parameters
+var (
+	GuildID        = flag.String("guild", "", "Test guild ID. If not passed - bot registers commands globally")
+	BotToken       = flag.String("token", "", "Bot access token")
+	RemoveCommands = flag.Bool("rmcmd", true, "Remove all commands after shutdowning or not")
+	BackupDir      = flag.String("backupdir", "backups", "Directory to store backups")
+)
+
+var s *discordgo.Session
+
+func init() { flag.Parse() }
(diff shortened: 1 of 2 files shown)


examples/server_backup/main.go
```diff
@@ -0,0 +1,387 @@
...
+				Data: &discordgo.InteractionResponseData{
+					Content: backupList.String(),
+				},
+			})
+		},
+		"get-backup": func(s *discordgo.Session, i *discordgo.InteractionCreate) {
+			options := i.ApplicationCommandData().Options
+			filename := options[0].StringValue()
+
+			filepath := fmt.Sprintf("%s/%s", *BackupDir, filename)
```
ldemailly, line 267: that's not very safe... one could ask for `../../../etc/password` etc

also in general this seems quite involved (a lot of code) for an example?

examples/server_backup/main.go
```diff
@@ -0,0 +1,387 @@
...
+			defer file.Close()
+
+			s.InteractionRespond(i.Interaction, &discordgo.InteractionResponse{
+				Type: discordgo.InteractionResponseChannelMessageWithSource,
+				Data: &discordgo.InteractionResponseData{
+					Content: "Here's your backup file:",
+					Files: []*discordgo.File{
+						{
+							Name:   filename,
+							Reader: file,
```
ldemailly, line 297: what are the odds this fits in the attachment size limit?

examples/server_backup/main.go
```diff
@@ -0,0 +1,387 @@
...
+				sendFollowUpError(s, i, "Failed to marshal backup data", err)
+				return
+			}
+
+			if err := os.WriteFile(filename, data, 0644); err != nil {
+				sendFollowUpError(s, i, "Failed to write backup file", err)
+				return
+			}
+
+			s.InteractionResponseEdit(i.Interaction, &discordgo.WebhookEdit{
```
ldemailly, line 174: isn't this going to timeout (interaction failed) while collecting all that data?

PR #1601, opened 2025-02-14
feat: Function to get the user ID as a uint64

IkeVoodoo:
Added the function `IdLong()` on the User struct, allowing to get the user's ID as a uint64.

### DIFF
--- a/user.go
+++ b/user.go
@@ -102,6 +102,8 @@ type User struct {
 	Flags int `json:"flags"`
 }
 
+// Parses the `ID` field to a uint64 and returns it.
+// This function is not cached
 func (u *User) IdLong() uint64 {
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -102,6 +102,15 @@ type User struct {
 	Flags int `json:"flags"`
 }
 
+// Parses the `ID` field to a uint64 and returns it.
+// This function is not cached
+func (u *User) IdLong() uint64 {
```
FedorLap2006, line 107:
```suggestion
func (u *User) IDLong() uint64 {
```

FedorLap2006 (changes requested):
LGTM, but I'd rename the function

PR #1616, opened 2025-04-26
Components V2

FedorLap2006:
Implement components V2 (see [docs](https://discord.com/developers/docs/components/reference#section) for more info)

PR #1623, opened 2025-05-17
Add support for components v2

d2orbc:


glotchimo (approved):
Looks good, thanks for including an example! Really looking forward to using the new components

glotchimo:
Ah just realized this is already in the works in #1616

d2orbc:
I didn't see that one. I'm using my fork without issue. Leaving to the maintainers to do whatever they want.

FedorLap2006:
Merged in 5b2d241

PR #1644, opened 2025-08-03
Fix: update default Discord API version to v10 for Components v2 support

0x00-sys:
## Summary

This fixes a compatibility issue with Discord Components v2 that were added in v0.29.0. The new component types require Discord API v10, but the library defaults to v9, causing runtime errors.

## Problem

Discord Components v2 (TextDisplay, Container, Section) were introduced in v0.29.0, but they only work with Discord API v10. Since the library defaults to API v9, using these components results in errors like:

```
type must be one of (1,2,3,4,5)
```

The Go types exist and compile fine, but Discord's API v9 rejects the newer component type values.

## Solution

Update the default `APIVersion` from "9" to "10" in `endpoints.go`. This ensures Components v2 work out of the box.

## Changes

- Updated `APIVersion` variable from "9" to "10" 
- Applied gofmt formatting fixes across codebase
- All tests pass with race detection enabled
- go vet shows no issues

## Alternative Workaround

If you need to use Components v2 before this fix, you can manually set the API version in your code:

```go
func init() {
    discordgo.APIVersion = "10"
}
```

## Backward Compatibility

This is a breaking change for applications that specifically rely on API v9 behavior. Such applications can explicitly set `discordgo.APIVersion = "9"` to maintain the previous behavior.

The `APIVersion` variable remains exported, so applications can still control which API version to use.


anhgelus (approved):

0x00-sys:
Correct me if I'm wrong but I think this is fine.

You're right about the variable initialization issue, but changing the default actually fixes it completely.

The problem with the `init()` approach is that `EndpointAPI` gets computed at package initialization:

```go
var APIVersion = "9"
var EndpointAPI = EndpointDiscord + "api/v" + APIVersion + "/"  // Computed once as "api/v9/"
```

Setting `discordgo.APIVersion = "10"` in `init()` doesn't work because the endpoint strings are already computed.

This fix changes the default to "10", so `EndpointAPI` correctly computes to "api/v10/" from startup. Now both REST and WebSocket use v10 consistently:

- REST calls: Use `EndpointAPI` → v10 
- WebSocket: Uses `APIVersion` directly → v10 
- Interaction responses: Use `EndpointInteractionResponse` which depends on `EndpointAPI` → v10

Components v2 now work properly because all Discord API calls hit the v10 endpoints that support the new component types.

CarsonHoffman:
The formatting changes here don't seem to be ones prescribed by gofmt (with no flags); are you using another formatter? I think it'd be best to keep the changes localized to the goal of the PR, not having random comments scattered throughout the codebase touched by it.

PR #1647, opened 2025-08-07
Returning old state on update or delete events

GameBear64:
Hello,

Im opening this PR since I am making a logging bot for my server, audit logs are not enough for us.
Here I basically do more of the same, adding BeforeDelete and BeforeUpdate to channel, thread, guild member and guild roles

I might also do the same for emojis too later on.

### DIFF
--- a/events.go
+++ b/events.go
@@ -60,7 +60,7 @@ type ChannelUpdate struct {
 // ChannelDelete is the data for a ChannelDelete event.
 type ChannelDelete struct {
 	*Channel
-	BeforeDelete *Channel
+	BeforeDelete *Channel `json:"-"`
 }
 
 // ChannelPinsUpdate stores data for a ChannelPinsUpdate event.


events.go
```diff
@@ -60,6 +60,7 @@ type ChannelUpdate struct {
 // ChannelDelete is the data for a ChannelDelete event.
 type ChannelDelete struct {
 	*Channel
+	BeforeDelete *Channel `json:"-"`
```
ldemailly, line 63: Taking this one as an example, what is the value of the base *Channel vs the new BeforeDelete ?

If *Channel is hollow wouldn't it be better to just fill it up from cache rather than add a new field?

GameBear64: Yep, I also think it would be better but it would not be consistent with how the other events like Message update and delete are, those also return a blank message with only a couple of IDs filled in.

Should I revise how we handle cached data in the events I've edited only or should I also fix up the other events like Message, Thread and so on.

ldemailly: What is in *Channel part (in delete), just some id at the moment?

Maybe it can be an option to fill missing fields from cache though that should be fairly cheap in any case, but I would keep that only for delete (documented) as it's "disappearing" while for other events the caller can make the call themselves?

ps: this is just my thoughts, @FedorLap2006 should chime in

FedorLap2006: I think we shouldn't touch the payload Discord sends us. Since that would be against our goal - 1:1 wrapper.

PR #1651, opened 2025-08-25
(ForumTag): implement custom JSON unmarshalling for ID field

estr0gen:
Sometimes Discord returns a number like `1000` instead of a normal string snowflake.


estr0gen:
This seems to be part of mod queue channels and their IDs are hard coded: https://github.com/discord-userdoccers/discord-userdoccers/issues/419#issuecomment-3218837983

PR #1668, opened 2025-10-11
docs(CONTRIBUTING): Add section on how to interact with local changes

MagnusHJensen:
This PR adds a new section to the `CONTRIBUTING.md` docs, that explains what one needs to do to verify the local changes made, the most prominent step is the `replace` directive, which should help out new contributors.

_Some formatting changes, that my editor picked up on, which seems to just be correcting some earlier mistakes, the markdown looks the same._


DeathHound6 (approved):
Another approach I've used instead of modifying `go.mod`, is go workspaces
> $ go work init [./path/to/local/discordgo]
> $ go work use [./path/to/another/local/module]

When a go workspace is created, the path to the `go.work` file, by default, is assigned into the `GOWORK` env var. This is how go knows which modules to use instead of the published remote variants
If you instead wish to use the module from go.pkg.dev, then set `GOWORK=off`

PR #1669, opened 2025-10-11
feat(Soundboard): Add support for the soundboard API and events

MagnusHJensen:
This PR adds support for soundboard API's that was introduced [September, 20th 2024](https://discord.com/developers/docs/change-log#soundboard-api).

It also adds the events, so they can be handled.

Let me know if this should also have state tracking, then I can go ahead and add that.

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -764,6 +764,7 @@ const (
 // https://discord.com/developers/docs/events/gateway-events#voice-channel-effect-send-animation-types
 type AnimationType int
 
+// Constants for AnimationType
 const (
 	AnimationTypePremium AnimationType = 0
 	AnimationTypeBasic   AnimationType = 1


MagnusHJensen:
> I've actually been working on the same thing for a few days, didn't notice this PR Just thinking now if more audit log action types are needed for soundboard related events aswell?

Wdym about the `audit log action types`?

MagnusHJensen:
> Would it be appropriate to add support for the [`VOICE_CHANNEL_EFFECT_SEND`](https://discord.com/developers/docs/events/gateway-events#voice-channel-effect-send) event in this change?

I can add that, I can see how that makes sense

MagnusHJensen:
@CarsonHoffman Added the `VOICE_CHANNEL_EFFECT_SEND` event to this PR now, should be ready to go.

PR #1671, opened 2025-10-11
feat(restapi): Add support for voice state API endpoints

MagnusHJensen:
This PR adds support for the Voice State HTTP endpoints that was released as of [August 5th, 2024](https://discord.com/developers/docs/change-log#voice-state-endpoints).

Also adds a missing Error Code for Unknown Voice State.


FedorLap2006 (changes requested):
I think it's better to merge these two funcs into one - `UserVoiceState` and just document that `@me` fetches the current user voice state.

PR #1672, opened 2025-10-29
fix(SelectMenu, TextInput): Add `Required` field and `omitempty` to `TextInput.Label`

Tesohh:
Hi, (thanks for merging my `Label`s PR by the way), I was adding modals to my bot and run into a few minor issues.

1. I added the `Required` field `SelectMenu`. (as seen in https://discord.com/developers/docs/components/reference#user-select)
2. I made `TextInput.Label` be `omitempty` so that we don't get an error if we don't specify a Label on the TextInput (which is unsupported, and would have given an error in both cases that it is specified or not) (https://discord.com/developers/docs/components/reference#text-input)

### DIFF
--- a/examples/modals/main.go
+++ b/examples/modals/main.go
@@ -34,6 +34,8 @@ func init() {
 	}
 }
 
+var required = true
+
 var (
 	commands = []discordgo.ApplicationCommand{
 		{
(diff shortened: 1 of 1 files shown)


glotchimo (approved):
Great, ran into the same, glad to see this up, hope it lands soon. Great work on the Label stuff

components.go
```diff
@@ -261,6 +261,7 @@ type SelectMenu struct {
 
 	Options  []SelectMenuOption `json:"options,omitempty"`
 	Disabled bool               `json:"disabled"`
+	Required bool               `json:"required"`
```
FedorLap2006:
```suggestion
	Required *bool               `json:"required,omitempty"`
```

Since documentation says it's `true` by default

PR #1673, opened 2025-10-30
feat: implement FileUploadComponent (Modals)

DeathHound6:
Modal component docs https://discord.com/developers/docs/components/reference#file-upload

### DIFF
--- a/user.go
+++ b/user.go
@@ -40,6 +40,21 @@ const (
 	UserPremiumTypeNitroBasic   UserPremiumType = 3
 )
 
+type UserPrimaryGuild struct {
+	// The ID of the user's primary guild.
+	IdentityGuildID string `json:"identity_guild_id"`
+
+	// Whether the user is displaying the primary guild's server tag.
+	IdentityEnabled bool `json:"identity_enabled"`
+
+	// The server tag of the user's primary guild. Limited to 4 characters
+	Tag string `json:"tag"`
+
+	// The server tag badge hash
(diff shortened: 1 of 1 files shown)


glotchimo (approved):
LGTM would love to see this land, loving the modal/components v2 stuff

josephschorr:
> You are right though that if this is not specified, Discord will default the value to be true, making the file upload required. If the user/developer wishes to make the file upload not required, then they simply set the field here, otherwise they don't need to specify

I tried it locally (by setting it to `false`) and the JSON sent to Discord was indeed missing the `required` field entirely, which in turn meant it was being marked as required (even though I explicitly set it to `false`). Go considers `false` as an empty value, so the `,omitempty` will skip emitting it in JSON when the value is `false`, causing Discord to assume it is `true`

FedorLap2006 (changes requested):
The `Required` field can be made into a pointer and thus set to both `false` and `true`, while `nil` omits the field from request to Discord.

Otherwise, when field is not specified, `false` will be sent to Discord, and by default it should be `true`.

FedorLap2006:
Something broke, heh
image

DeathHound6:
@FedorLap2006 Yea, it seems to do that every time another commit is pushed to master. So I know for next time not to use master from my fork :/

At this point, I'm not too sure how to go about resolving this and getting the actual commit merged

DeathHound6:
New PR, same commit, from a different branch ^

PR #1680, opened 2025-11-27
feat(Session): pool decompression buffers and eliminate decoder alloc

glotchimo:
**TL;DR: Pooling compression buffers and switching to `Unmarshal` reduces allocated memory per event by 32-39%.**

Benchmarks: https://gist.github.com/glotchimo/ba3e91aac40c7ee0612d17b98542be24

I was drawn to `onEvent` when it jumped out in production heap profiling and these seemed like some free optimizations that could trim down allocations for apps that handle hot events like `MESSAGE_CREATE`.

### DIFF
--- a/components.go
+++ b/components.go
@@ -26,6 +26,7 @@ const (
 	SeparatorComponent             ComponentType = 14
 	ContainerComponent             ComponentType = 17
 	LabelComponent                 ComponentType = 18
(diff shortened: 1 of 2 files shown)


glotchimo:
Started doing more with fork than anticipated, will reopen if there's interest from a branch

PR #1696, opened 2026-02-14
New components

GameParrot:
Adds new components (https://docs.discord.com/developers/change-log#the-new-components)

### DIFF
--- a/components.go
+++ b/components.go
@@ -27,6 +27,9 @@ const (
 	ContainerComponent             ComponentType = 17
 	LabelComponent                 ComponentType = 18
 	FileUploadComponent            ComponentType = 19
+	RadioGroupComponent            ComponentType = 21
+	CheckboxGroupComponent         ComponentType = 22
+	CheckboxComponent              ComponentType = 23
 )
 
 // MessageComponent is a base interface for all message components.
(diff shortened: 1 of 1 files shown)


verzac:
Hey @FedorLap2006, I'd like to check: do we have any similar PRs/plans for adding the 3 new components?

image

Note: I haven't reviewed the changes here so I can't attest to its correctness (happy to help if needed)

Asking because one of my bot's users were asking if I could add a feature that utilises the new Checkbox components

components.go
```diff
@@ -668,6 +677,103 @@ func (f FileUpload) MarshalJSON() ([]byte, error) {
...
+		radioGroup: radioGroup(r),
+		Type:       r.Type(),
+	})
+}
+
+type RadioGroupOption struct {
+	Label       string `json:"label"`
+	Value       string `json:"value"`
+	Description string `json:"description,omitempty"`
+	Default     bool   `json:"default,omitempty"`
```
verzac, line 713: Default should be optional and should be a nil-pointer instead (since Default=false is a valid value)

(Encountered this problem when I wanted to set all my options to be false by default haha, not sure why Discord preserves the previous modal selections)

components.go
```diff
@@ -668,6 +677,103 @@ func (f FileUpload) MarshalJSON() ([]byte, error) {
...
+}
+
+// CheckboxGroup is an interactive component for selecting one or many options via checkboxes.
+// CheckboxGroups are available in modals and must be placed inside a Label.
+type CheckboxGroup struct {
+	// Unique identifier for the component; auto populated through increment if not provided.
+	ID        int                `json:"id,omitempty"`
+	CustomID  string             `json:"custom_id,omitempty"`
+	Options   []RadioGroupOption `json:"options"`
```
verzac, line 722: Shouldn't this be CheckboxGroupOption instead? They have the same struct, but they're intended for different purposes

verzac (changes requested):

PR #1715, opened 2026-04-21
fix(wsapi): nil-guard wsConn in Op1 heartbeat reply and ChannelVoiceJoinManual

### LINKED ISSUE #1702: Nil pointer dereference in onEvent() during reconnection - missing wsConn nil check
### Description

`onEvent()` in `wsapi.go` accesses `s.wsConn` without a nil check when responding to Op1 (heartbeat request). This causes a nil pointer dereference panic when `CloseWithCode()` sets `s.wsConn = nil` concurrently from another goroutine.

The panic is not recoverable because `listen()`, which calls `onEvent()`, has no `recover()`. So the panic propagates up and kills the entire process.

**`ChannelVoiceJoinManual()` has the same bug** -- it also writes to `s.wsConn` using only `s.wsMutex` without holding `s.RLock()` or checking for nil (line 783). The race condition and crash are identical.

### Stack Trace (onEvent)

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]

goroutine 13250824 [running]:
github.com/gorilla/websocket.(*Conn).beginMessage(0x0, 0x1)
github.com/gorilla/websocket.(*Conn).WriteJSON(0x0, {0x100e18de0, 0x14009b06ea0})
github.com/bwmarrin/discordgo.(*Session).onEvent(0x140099cc500, 0x1, {0x14009cf8000, ...})
        wsapi.go:606
github.com/bwmarrin/discordgo.(*Session).listen(0x140099cc500, 0x14009854240, 0x14009c58180)
        wsapi.go:259
```

The `0x0` receiver on `(*Conn).beginMessage` confirms `wsConn` is nil at the time of the call.

### Stack Trace (ChannelVoiceJoinManual -- same root cause)

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]

goroutine N [running]:
github.com/gorilla/websocket.(*Conn).beginMessage(0x0, 0x1)
github.com/gorilla/websocket.(*Conn).WriteJSON(0x0, {0x100e18de0, 0x14009b12340})
github.com/bwmarrin/discordgo.(*Session).ChannelVoiceJoinManual(0x140099cc500, ...)
        wsapi.go:783
```

Same nil receiver, same crash. Any caller joining/leaving a voice channel during a reconnection can trigger this.

### Root Cause

There is a race condition between these methods and `CloseWithCode()`. They protect `s.wsConn` with two different locks:

| Method | Lock used | What it does |
|--------|-----------|--------------|
| `onEvent()` (line 606) | `s.wsMutex` | Reads `s.wsConn` |
| `ChannelVoiceJoinManual()` (line 783) | `s.wsMutex` | Reads `s.wsConn` |
| `CloseWithCode()` (line 1007) | `s.Lock()` (session RWMutex) | Sets `s.wsConn = nil` |

Because these are different mutexes, they don't synchronize with each other. The race looks like this:

1. `onEvent()` or `ChannelVoiceJoinManual()` acquires `s.wsMutex.Lock()`, `s.wsConn` is non-nil at this point
2. `CloseWithCode()` acquires `s.Lock()`, sets `s.wsConn = nil` (line 1007), and releases `s.Unlock()`
3. The method calls `s.wsConn.WriteJSON(...)` and panics because `wsConn` is now nil

### Existing Nil Checks

Other methods in the same file already guard against this. The nil check exists in:

- **`heartbeat()`** - uses a local `wsConn` parameter instead of `s.wsConn` (safe by design)
- **`UpdateStatusComplex()`** (line 441-448) - checks `s.wsConn == nil` under `s.RLock()`, returns `ErrWSNotFound`
- **`GatewayWriteStruct()`** (line 534-543) - same pattern
- **`requestGuildMembers()`** (line 551-559) - same pattern

But `onEvent()` at line 606 and `ChannelVoiceJoinManual()` at line 783 don't have any of this:

```go
// wsapi.go:606 - current code, no nil check
s.wsMutex.Lock()
err = s.wsConn.WriteJSON(heartbeatOp{1, atomic.LoadInt64(s.sequence)})
s.wsMutex.Unlock()
```

```go
// wsapi.go:783 - current code, no nil check
s.wsMutex.Lock()
err = s.wsConn.WriteJSON(data)
s.wsMutex.Unlock()
```

### Reproduction

```go
func TestOnEventNilWsConn(t *testing.T) {
    seq := int64(0)
    s := &Session{sequence: &seq}
    // wsConn is nil, same state as after CloseWithCode() runs

    op1 := []byte(`{"op":1,"d":null}`)

    // Panics with: runtime error: invalid memory address or nil pointer dereference
    s.onEvent(websocket.TextMessage, op1)
}

func TestChannelVoiceJoinManualNilWsConn(t *testing.T) {
    seq := int64(0)
    s := &Session{sequence: &seq}
    // wsConn is nil, same state as after CloseWithCode() runs

    // Panics with: runtime error: invalid memory address or nil pointer dereference
    s.ChannelVoiceJoinManual("guild123", "channel456", false, false)
}
```

Running `go test -race` with a concurrent close also triggers the race detector.

### Suggested Fix

Use the same pattern that `UpdateStatusComplex()`, `GatewayWriteStruct()`, and `requestGuildMembers()` already use - check under the session read lock before writing:

**onEvent() fix:**
```go
if e.Operation == 1 {
    s.log(LogInformational, "sending heartbeat in response to Op1")

    s.RLock()
    if s.wsConn == nil {
        s.RUnlock()
        return e, ErrWSNotFound
    }

    s.wsMutex.Lock()
    err = s.wsConn.WriteJSON(heartbeatOp{1, atomic.LoadInt64(s.sequence)})
    s.wsMutex.Unlock()
    s.RUnlock()

    if err != nil {
        s.log(LogError, "error sending heartbeat in response to Op1")
        return e, err
    }

    return e, nil
}
```

**ChannelVoiceJoinManual() fix:**
```go
data := voiceChannelJoinOp{4, voiceChannelJoinData{&gID, channelID, mute, deaf}}
s.RLock()
defer s.RUnlock()
if s.wsConn == nil {
    return ErrWSNotFound
}
s.wsMutex.Lock()
err = s.wsConn.WriteJSON(data)
s.wsMutex.Unlock()
return
```

Both fixes keep the same lock ordering (`s.RLock()` then `s.wsMutex.Lock()`) and nil check pattern as the rest of the codebase. They properly synchronize with `CloseWithCode()` since `s.RLock()` and `s.Lock()` are the same RWMutex.

Also worth considering adding a `recover()` in `listen()` as a safety net, so that if something does panic in that goroutine it doesn't take down the whole process.

### Impact

This crash kills the entire Go process. For applications running multiple Discord sessions, one session hitting this race during reconnection takes down everything. Since the panic happens inside a goroutine spawned by discordgo (`listen`), the caller has no way to catch it.

### Version

Reproduced on `master` (current HEAD) and `v0.28.1`. The code path has been there since `onEvent()` was written.

SAY-5:
## What

Fixes #1702.

Two call sites in `wsapi.go` used only `s.wsMutex` to protect a `WriteJSON` on `s.wsConn`:

- Op1 heartbeat response inside `onEvent` (`wsapi.go:606`)
- Voice-channel join send in `ChannelVoiceJoinManual` (`wsapi.go:783`)

`CloseWithCode` clears `s.wsConn` while holding `s.Lock` (not `s.wsMutex`), so a reconnect that lands in the window between the heartbeat tick / voice-join call and the `WriteJSON` dispatches the method on a nil receiver. `gorilla/websocket` then panics inside `beginMessage` / `WriteJSON` and, because `listen()` has no `recover`, the panic propagates through the caller and kills the whole process.

## Fix

Follow the pattern that `sendEvent`, `heartbeat`, and `identify` already use in this file (wsapi.go:437, :531, :547): with `s.wsMutex` held, check if `s.wsConn` is nil and return `ErrWSNotFound` instead of dereferencing. No behavioural change on the happy path - the affected goroutine cleanly observes `no websocket connection exists` and the app stays alive through the reconnect.

## Verification

Locally on macOS, go 1.26.2:
- `gofmt -s -l wsapi.go`: clean
- `go vet ./...`: clean
- `go build ./...`: clean

Closes #1702

### DIFF
--- a/wsapi.go
+++ b/wsapi.go
@@ -603,6 +603,15 @@ func (s *Session) onEvent(messageType int, message []byte) (*Event, error) {
 	if e.Operation == 1 {
 		s.log(LogInformational, "sending heartbeat in response to Op1")
 		s.wsMutex.Lock()
+		// CloseWithCode can null out s.wsConn on a different goroutine
+		// during reconnection teardown. Hitting WriteJSON on the nil
+		// receiver panics inside gorilla/websocket and, because listen()
+		// has no recover, the entire host process dies (#1702).
+		if s.wsConn == nil {
+			s.wsMutex.Unlock()
+			s.log(LogInformational, "skipping Op1 heartbeat response: wsConn is nil (likely mid-reconnect)")
+			return e, ErrWSNotFound
+		}
 		err = s.wsConn.WriteJSON(heartbeatOp{1, atomic.LoadInt64(s.sequence)})
 		s.wsMutex.Unlock()
 		if err != nil {
@@ -714,10 +723,10 @@ type voiceChannelJoinOp struct {
 
 // ChannelVoiceJoin joins the session user to a voice channel.
 //
-//    gID     : Guild ID of the channel to join.
-//    cID     : Channel ID of the channel to join.
-//    mute    : If true, you will be set to muted upon joining.
-//    deaf    : If true, you will be set to deafened upon joining.
+//	gID     : Guild ID of the channel to join.
+//	cID     : Channel ID of the channel to join.
+//	mute    : If true, you will be set to muted upon joining.
+//	deaf    : If true, you will be set to deafened upon joining.
 func (s *Session) ChannelVoiceJoin(gID, cID string, mute, deaf bool) (voice *VoiceConnection, err error) {
 
 	s.log(LogInformational, "called")
@@ -761,10 +770,10 @@ func (s *Session) ChannelVoiceJoin(gID, cID string, mute, deaf bool) (voice *Voi
 //
 // This should only be used when the VoiceServerUpdate will be intercepted and used elsewhere.
 //
-//    gID     : Guild ID of the channel to join.
-//    cID     : Channel ID of the channel to join, leave empty to disconnect.
-//    mute    : If true, you will be set to muted upon joining.
-//    deaf    : If true, you will be set to deafened upon joining.
+//	gID     : Guild ID of the channel to join.
+//	cID     : Channel ID of the channel to join, leave empty to disconnect.
+//	mute    : If true, you will be set to muted upon joining.
+//	deaf    : If true, you will be set to deafened upon joining.
 func (s *Session) ChannelVoiceJoinManual(gID, cID string, mute, deaf bool) (err error) {
 
 	s.log(LogInformational, "called")
@@ -779,6 +788,14 @@ func (s *Session) ChannelVoiceJoinManual(gID, cID string, mute, deaf bool) (err 
 	// Send the request to Discord that we want to join the voice channel
 	data := voiceChannelJoinOp{4, voiceChannelJoinData{&gID, channelID, mute, deaf}}
 	s.wsMutex.Lock()
+	// Same race as onEvent's Op1 heartbeat path (#1702): a concurrent
+	// reconnect can clear s.wsConn while we hold only s.wsMutex, and
+	// a nil receiver into WriteJSON would panic in gorilla/websocket
+	// and crash the caller's process.
+	if s.wsConn == nil {
+		s.wsMutex.Unlock()
+		return ErrWSNotFound
+	}
 	err = s.wsConn.WriteJSON(data)
 	s.wsMutex.Unlock()
 	return


SAY-5:
Ping on this if you have a moment. Both call sites write to s.wsConn under s.wsMutex only, while CloseWithCode nils it under s.Lock, so a reconnect in that window panics inside WriteJSON and takes the process down since listen() has no recover (#1702).

0x00-sys:
There's no development for this project anymore it seems like

PR #1721, opened 2026-05-23
Handle unknown message component types

### LINKED ISSUE #1683: Unknown component type - panic on access
Hi, 

Discord "released" a new component type (20) for checkpoints without being listed on https://discord.com/developers/docs/components/reference#component-object-component-types

https://github.com/bwmarrin/discordgo/blob/22e5cd898d089e319c91f0941c42bf931cf0bf98/components.go#L77-L80

Because of the error handling here, it's currently simply returning an error which isn't stopping handlers to work with the event.
That means accessing the event in your handler does end in a panic:

```
wsapi.go:661:onEvent() error unmarshalling MESSAGE_CREATE event, unknown component type: 20"}
panic: runtime error: invalid memory address or nil pointer dereference
```

Guess simply removing the return err would fix the problem(?).

0x00-sys:
Summary:
- Preserve unknown message components as opaque components during unmarshal.
- Keep the original component payload available through MarshalJSON.
- Add a regression test for component type 20 in MESSAGE_CREATE data.

Why:
- Discord can send component types before discordgo has concrete structs for them. Returning an unmarshal error can leave message event data unusable and lead handlers to panic.

Severity:
- High

Upstream:
- Fixes bwmarrin/discordgo#1683
- Reviewed upstream PR #1696. It adds some known component types, but it does not fix the unknown-type fallback.

Tests:
- go test -run TestMessageCreateUnknownComponentType .
- go test -race -run TestMessageCreateUnknownComponentType .
- go test ./...
- git diff --check

Notes:
- Small compatibility fallback only.
- No new dependencies.

### DIFF
--- a/components.go
+++ b/components.go
@@ -39,6 +39,27 @@ type unmarshalableMessageComponent struct {
 	MessageComponent
 }
 
+type unknownMessageComponent struct {
+	ComponentType ComponentType
+	RawData       json.RawMessage
+}
+
+func (c unknownMessageComponent) Type() ComponentType {
+	return c.ComponentType
+}
+
+func (c unknownMessageComponent) MarshalJSON() ([]byte, error) {
+	if c.RawData != nil {
+		return c.RawData, nil
+	}
+
+	return Marshal(struct {
+		Type ComponentType `json:"type"`
+	}{
+		Type: c.Type(),
+	})
+}
+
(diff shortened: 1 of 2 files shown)


0x00-sys:
Moved this draft checkpoint to my fork for local integration: https://github.com/0x00-sys/discordgo/pull/3

PR #1722, opened 2026-05-23
Guard voice AEAD access

0x00-sys:
Summary:
- Copy the voice AEAD cipher under RLock before sender and receiver use it.
- Skip opus packets until the cipher is available.
- Add a regression test for nil AEAD in the opus sender.

Why:
- The OP4 handler sets the AEAD cipher under the voice lock, while opus sender and receiver goroutines read it separately. The sender can panic if audio arrives before the cipher is ready.

Severity:
- High

Upstream:
- Relates to bwmarrin/discordgo#1692
- Reviewed upstream PR #1692 and reused the small lock-copy pattern.

Tests:
- go test -run TestOpusSenderNilAEAD .
- go test -race -run TestOpusSenderNilAEAD .
- go test ./...
- go test -race ./...
- git diff --check

Notes:
- Small localized fix in voice.go.
- No public API changes.

### DIFF
--- a/voice.go
+++ b/voice.go
@@ -774,10 +774,17 @@ func (v *VoiceConnection) opusSender(udpConn *net.UDPConn, close <-chan struct{}
 
 		// encrypt the opus data
 		// add incrementing nonce counter as per discord's requirements
+		v.RLock()
+		aead := v.aead
+		v.RUnlock()
+		if aead == nil {
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


0x00-sys:
Moved this draft checkpoint to my fork for local integration: https://github.com/0x00-sys/discordgo/pull/4

